PR #1, opened 2017-05-08
Add Pg codec

EmadAlblueshi:


vietj:
first pass:

- in the decoder, factor the various cases of the switch in methods as it helps readability and more importantly it is more JIT friendly
- in `toString` avoid to use `StringBuffer` because it is not necessary, plain string concatenation actually generates bytecode that uses a `StringBuffer`

src/main/java/io/vertx/pgclient/codec/encoder/PgMessageEncoder.java
```diff
@@ -0,0 +1,70 @@
...
+  @Override
+  protected void encode(ChannelHandlerContext ctx, Message message, ByteBuf out) throws Exception {
+
+    if (message.getClass() == StartupMessage.class) {
+
+      StartupMessage startup = (StartupMessage) message;
+      out.writeInt(0);
+      out.writeShort(3);
+      out.writeShort(0);
+      for (String s : new String[]{
```
vietj: avoid `String[]` creation

EmadAlblueshi: @vietj arraylist or initialized String[] ?

src/main/java/io/vertx/pgclient/codec/decoder/PgMessageDecoder.java
```diff
@@ -0,0 +1,424 @@
...
+    final String ROLLBACK = "ROLLBACK";
+    final String DEALLOCATE = "DEALLOCATE";
+    final String TRUNCATE = "TRUNCATE";
+    final String LOCK = "LOCK";
+    final String GRANT = "GRANT";
+    final String REVOKE = "REVOKE";
+
+
+    String tag = readCString(data, UTF_8);
+    String[] parts = tag.split(" ");
```
vietj: avoid `split`, instead create a split util method somewhere that takes the input string and a char separator.

src/main/java/io/vertx/pgclient/codec/decoder/PgMessageDecoder.java
```diff
@@ -0,0 +1,424 @@
...
+        } else {
+          throw new DecoderException("error parsing command tag: " + command + " (" + Arrays.toString(parts) + ")");
+        }
+        break;
+      case CREATE:
+      case DROP:
+      case ALTER:
+      case DECLARE:
+      case CLOSE:
+        if (parts.length == 2) {
```
vietj: use switch/case ?

src/main/java/io/vertx/pgclient/codec/decoder/PgMessageDecoder.java
```diff
@@ -0,0 +1,424 @@
...
+    final String DEALLOCATE = "DEALLOCATE";
+    final String TRUNCATE = "TRUNCATE";
+    final String LOCK = "LOCK";
+    final String GRANT = "GRANT";
+    final String REVOKE = "REVOKE";
+
+
+    String tag = readCString(data, UTF_8);
+    String[] parts = tag.split(" ");
+    String command = parts[0];
```
vietj: use `StringBuilder` instead here

src/main/java/io/vertx/pgclient/codec/decoder/PgMessageDecoder.java
vietj: avoir array creation as it's used below with `command` that can be a `StringBuilder`, we can also avoid the `tag` creation` by just using an index.

Overall we can even use a `ByteBuf` instead a `StringBuilder`

src/main/java/io/vertx/pgclient/codec/utils/Utils.java
```diff
@@ -18,4 +18,15 @@ public static void writeCString(ByteBuf buffer, String input, Charset charset) {
     buffer.writeByte(0);
   }
 
+  public static int readSpaceCount(ByteBuf buffer) {
+    final byte SPACE = 32;
+    int spaceCount = 0;
+    for (int i = 0; i < buffer.readableBytes(); i++) {
```
vietj: use ByteBuf `forEachByte` : http://normanmaurer.me/presentations/2014-facebook-eng-netty/slides.html#18.0

EmadAlblueshi: @vietj I tried it before but I will try it again :)

EmadAlblueshi: The `readSpaceCount` is counting the no of spaces in order to parse the result that depends on the SQL command for example 

INSERT  -> INSERT oid rows
UPDATE -> UPDATE rows
DELETE -> DELETE rows
SELECT -> SELECT rows

rows should be parsed as a number to get `result.getUpdatedRows()` in the Result object

I have followed the [ PostgreSQL docs in the CommandComplete section](https://www.postgresql.org/docs/9.5/static/protocol-message-formats.html)

Regarding `ByteBufProcessor` I think it does not do that for spaces.

EmadAlblueshi:
@vietj Your feedback is valuable :)

src/main/java/io/vertx/pgclient/codec/encoder/MessageEncoder.java
```diff
@@ -39,6 +39,10 @@ protected void encode(ChannelHandlerContext ctx, Message message, ByteBuf out) t
       writeCString(out, "vertx-pg-client", UTF_8);
       writeCString(out, "client_encoding", UTF_8);
       writeCString(out, "utf8", UTF_8);
+      writeCString(out, "DateStyle", UTF_8);
+      writeCString(out, "ISO", UTF_8);
+      writeCString(out, "extra_float_digits", UTF_8);
+      writeCString(out, "2", UTF_8);
```
EmadAlblueshi: > **Java issues**
> If you happen to use Java as the frontend, there are some points that have to be
> taken into consideration. Java tends to pass some parameters to the server as part
> of the connection string. One of those parameters is `extra_float_digits`. This
> `postgresql.conf` parameter governs the floating-point behavior of PostgreSQL,
> and is set by Java to make things more deterministic.

[PostgreSQL Replication - Second Edition](https://www.packtpub.com/big-data-and-business-intelligence/postgresql-replication-second-edition)

EmadAlblueshi:
@vietj Kindly review :)

EmadAlblueshi:
👍

PR #12, opened 2017-05-30
Support Transaction Management 

EmadAlblueshi:


src/main/java/com/julienviet/pgclient/impl/PostgresSQLConnection.java
```diff
@@ -30,81 +38,108 @@ public PostgresSQLConnection(DbConnection conn) {
 
   @Override
   public SQLConnection setAutoCommit(boolean autoCommit, Handler<AsyncResult<Void>> handler) {
-    throw new UnsupportedOperationException();
+    inAutoCommit.set(autoCommit);
```
pmlopes: This should be only set if the rest of the code succeeded, in other words, right before any: `handler.handle(Future.succeededFuture())`

src/main/java/com/julienviet/pgclient/impl/PostgresSQLConnection.java
```diff
@@ -142,4 +198,64 @@ public SQLConnection getTransactionIsolation(Handler<AsyncResult<TransactionIsol
     conn.schedule(new PreparedTxQueryCommand(handler));
     return this;
   }
+
+  @Override
+  public void close(Handler<AsyncResult<Void>> handler) {
+    inAutoCommit.set(true);
```
pmlopes: This should only set if the commit succeeded

src/main/java/com/julienviet/pgclient/impl/PostgresSQLConnection.java
```diff
@@ -142,4 +198,64 @@ public SQLConnection getTransactionIsolation(Handler<AsyncResult<TransactionIsol
...
+
+  @Override
+  public void close() {
+    close(c -> {
+    });
+  }
+
+  private void beginTxIfNeeded(Handler<AsyncResult<Void>> handler) {
+    if (!inAutoCommit.get() && inTransaction.compareAndSet(false, true)) {
+      conn.schedule(new QueryCommand("BEGIN", new ResultSetBuilder(re -> {
```
pmlopes: This could be enhanced to support TX isolation levels, maybe a follow up pull request see: https://github.com/vert-x3/vertx-mysql-postgresql-client/pull/74/files#diff-7a3a21a56ba64de74ca084f12538d869

src/main/java/com/julienviet/pgclient/impl/PostgresSQLConnection.java
```diff
@@ -142,4 +198,64 @@ public SQLConnection getTransactionIsolation(Handler<AsyncResult<TransactionIsol
...
+    } else {
+      handler.handle(Future.succeededFuture());
+    }
+  }
+
+  private void endAndBeginTx(String command, Handler<AsyncResult<Void>> handler) {
+    if (inTransaction.compareAndSet(true, false)) {
+      conn.schedule(new QueryCommand(command, new ResultSetBuilder(re -> {
+        if(re.failed()) {
+          handler.handle(Future.failedFuture(re.cause()));
```
pmlopes: if it failed then the inTransaction is now in a invalid state since it has been modified to false before

src/main/java/com/julienviet/pgclient/impl/PostgresSQLConnection.java
```diff
@@ -142,4 +198,64 @@ public SQLConnection getTransactionIsolation(Handler<AsyncResult<TransactionIsol
...
+        } else {
+          conn.schedule(new QueryCommand("BEGIN", new ResultSetBuilder(rs -> {
```
pmlopes: See previous comment on TX isolation, maybe for later?!

EmadAlblueshi: IMHO, I prefer later because we need to implement other SQL connection methods with tests then we enhance

src/main/java/com/julienviet/pgclient/impl/PostgresSQLConnection.java
vietj: I think that we should rethink this part, because with pipelining it is different than with the other driver

vietj: so the problem is more than when `autocommit` is set to false and we are in a transaction then the `SQLConnection` either not be usable (i.e throw IllegalStateException) or the evaluation of `beginTxIfNeeded` should be delayed.

pmlopes: no, if we follow the jdbc semantics, turning autocommit to false during a tx should commit the tx if i'm not mistaken

pmlopes: The improvements I'v proposed to the async driver are doing this.

vietj:
autocommit set to true means there is never a TX around the statements, no ?

pmlopes:
when you do `autocommit=true` the expected behavior on java land and other places is that a transaction is started (internally it should issue a `begin transaction ...` command).

as of this moment you can execute any sql statement and the API will allow you to:

* `commit()` - executes `commit`
* `rollback()` - executes `rollback`

we don't have it but jdbc also has `savepoints` so lets stay away from this.

Now if you turn `autocommit=false` and the previous state was `true` then it is expected to be same as `commit()`, otherwise is a no-op.

Calling `commit()`/`rollback()` during a `autocommit=false` should be a warning, some engines throw error, but this is a non fatal action, usually the warning says something like `atempt to commit/rollback without a transaction`.

Calling `close()` during `autocommit=true` should be also handled as `commit()` and if the commit fails propagate the error to the `close()`

pmlopes:
Now in a `autocommit=true` you can have transactions within transaction, this is what some engines call savepoints. so you can rollback a to a savepoint so you can have a TX within autocommit.

src/main/java/com/julienviet/pgclient/impl/PostgresSQLConnection.java
```diff
@@ -30,81 +38,108 @@ public PostgresSQLConnection(DbConnection conn) {
 
   @Override
   public SQLConnection setAutoCommit(boolean autoCommit, Handler<AsyncResult<Void>> handler) {
-    throw new UnsupportedOperationException();
+    inAutoCommit.set(autoCommit);
+    if (inTransaction.compareAndSet(true, false) && autoCommit) {
```
vietj: the boolean autocommit should be evaluated before changing the state of the `inTransaction` atomic

vietj: overall I believe we should avoid to use atomic here and use plain synchronization because we are checking the state of two variables and there can be race issues.

vietj:
agreed with paulo to just have one flag, this opens race issues

vietj:
does pg has savepoints ? I haven't seen something like that

vietj:
is it Sync ?

EmadAlblueshi:
Yes it has naming save points

pmlopes:
this is a handy feature when you're working with long running transactions, on error you can rollback to the savepoint and retry from that point instead of rolling back the whole transaction, but is was never present on sql-common clients.

PR #17, opened 2017-09-13
Cleanup and PgUpdate support

EmadAlblueshi:


src/main/java/com/julienviet/pgclient/impl/PreparedUpdate.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package com.julienviet.pgclient.impl;
+
+import com.julienviet.pgclient.PgUpdate;
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Handler;
+import io.vertx.ext.sql.UpdateResult;
+
+public class PreparedUpdate implements PgUpdate {
```
vietj: rather `PgUpdateImpl`

vietj:
overall we should not duplicate the same class with a params and a param-less version. Param-less should be params command with empty params list.

EmadAlblueshi:
Done  https://github.com/vietj/vertx-pg-client/issues/14 https://github.com/vietj/vertx-pg-client/issues/15 and https://github.com/vietj/vertx-pg-client/issues/18

src/main/java/com/julienviet/pgclient/impl/PreparedBatchCommand.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import static com.julienviet.pgclient.codec.util.Util.*;
+
+/**
+ * @author <a href="mailto:emad.albloushi@gmail.com">Emad Alblueshi</a>
+ */
+
+class PreparedBatchCommand extends UpdateCommandBase {
```
vietj: please keep PreparedUpdateCommand and consider that an update is a batch of a single unit.

EmadAlblueshi: Yes now we have only PreparedUpdateCommand and can be used for batch and update

src/test/java/com/julienviet/pgclient/PgConnectionTestBase.java
```diff
@@ -1060,6 +920,44 @@ public void testSetAndGetSerializableTx(TestContext ctx) {
     });
   }
 
+  @Test
+  public void testPgUpdate(TestContext ctx) {
```
vietj: just testUpdate no need to say PgUpdate here

src/test/java/com/julienviet/pgclient/PgConnectionTestBase.java
```diff
@@ -1060,6 +920,44 @@ public void testSetAndGetSerializableTx(TestContext ctx) {
...
+          .execute(ctx.asyncAssertSuccess(r -> {
+            ctx.assertEquals("PgClient Rocks!", r.getRows().get(0).getValue("message"));
+            async.complete();
+          }));
+      }));
+    }));
+  }
+
+  @Test
+  public void testPgUpdateWithParams(TestContext ctx) {
```
vietj: testUpdateWithParams

vietj:
The documentation is not covering the update case, it should be updated (after query and before batch).

src/test/java/com/julienviet/pgclient/PgConnectionTestBase.java
EmadAlblueshi: The testUpdate will conflict with other test case so it's better to rename to testPreparedUpdate

src/test/java/com/julienviet/pgclient/PgConnectionTestBase.java
EmadAlblueshi: This will be renamed to testPreparedUpdateWithParams

PR #22, opened 2017-09-15
Decode SSLResponse message + Support Keys in UpdateResult

EmadAlblueshi:


vietj:
can you make two PRs instead of one ? this PR mixes two concerns, it should'nt be the case

EmadAlblueshi:
Ok

PR #29, opened 2017-11-28
Binary numeric data type decode

EmadAlblueshi:


src/main/java/com/julienviet/pgclient/codec/DataType.java
```diff
@@ -266,10 +287,24 @@ private static long decodeInt(int len, ByteBuf buff) {
   }
 
   private static Object decodeJson(String value) {
-    if(value.charAt(0)== '{') {
+    if(value.indexOf('{') != -1) {
```
vietj: you should keep using `charAt(0)`, `indexOf` will do a full scan of the string when it does not find the curly brace

src/test/java/com/julienviet/pgclient/PgDataTypeBase.java
```diff
@@ -0,0 +1,5 @@
+package com.julienviet.pgclient;
+
+public class PgDataTypeBase extends PgConnectionTest {
```
vietj: make it abstract and name it `DataTypeTestBase`

vietj: also it should not extend PgConnectionTest because it will run the PgConnectionTest and we only want to run the data type tests

src/main/java/com/julienviet/pgclient/codec/DataType.java
EmadAlblueshi: The reason I did this because JSON doesn't trim spaces but JSONB does.

EmadAlblueshi:
:+1: Thanks

PR #32, opened 2017-12-08
Minor Improvements

BillyYccc:
1. Add document for PgResult
2. Upgrade Vert.x Stack Version to 3.5.0 from 3.5.0 Beta1

After upgrading to 3.5.0, all unit tests have passed, and I made some fundamental tests on the main functions, it seems to work in a right way. But I didn't pay attention to the results of benchmark, maybe posting the comparison of  benchmark results in two stack versions is necessary?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
   </developers>
 
   <properties>
-    <stack.version>3.5.0.Beta1</stack.version>
+    <stack.version>3.5.0</stack.version>
     <jmh.version>1.19</jmh.version>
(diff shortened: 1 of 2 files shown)


vietj:
can you make 2 prs ?

at the moment I don't want to upgrade to 3.5.0 (soon for 0.4.0 release) because I'm still testing this client with 3.5.0.Beta1

BillyYccc:
OK, I will close this PR and open another PR for only adjusting docs.

PR #44, opened 2017-12-11
Api for easing transaction handling

vietj:
not meant to be merged - more for review


vietj:
@pmlopes can you have a look ?

vietj:
@pmlopes in particular, the callback when the transaction rollbacks on failure or not, e.g

```
conn.begin();
conn.query(stmt1, ar -> {});
conn.query(stmt2, ar -> {}); // This stmt produces an error and rollback the tx
conn.query(stmt3, ar -> {}); // Should the handler be called ?
conn.commit(ar -> {}); // Should the handler be called ?
```

pmlopes:
Both cases could be valid I can think of creating a temp table in the transaction and if already present the statement would fail but it would make sense to continue executing the remaining statements.

And a strict mode think of stopOnError(true) that would break on the first error however in this case you would require to also manage commit and rollback.

Because if an error is raised should this api execute the rollback?

As I write I’m more of the idea that the Boolean strict is probably good idea but not sure about automatically calling commit / rollback

vietj:
now that I think about it, this problem does not arise with blocking code, because one statement is executed after another, making it impossible to queue statements in a transaction.

the current implement mimics this behavior, i.e with:

```
conn.begin();
conn.query(q1);
conn.query(q2); // This is only enqueued when q1 has completed succesfully otherwise it's rollbacked
conn.commit();
```

mimics of real behavior would rather be to throw an exception if `conn.query(q2)` is executed before the first query has been executing, forcing the user to take a decision in the async result handler.

I'm leaning toward modifying to this behavior.

vietj:
if we think at the behavior of vertx jdbc client with this scheme, the queries would be queued on the execute blocking task queue and I believe the next statements would be executed and there is no automatic transaction rollback, right @pmlopes ?

pmlopes:
Right

vietj:
@pmlopes thanks for the info and details, I will think a bit about it

vietj:
@pmlopes does it make sense to rollback unfinished transaction when the connection is given back to the pool ?

vietj:
that's what Hikari CP does

pmlopes:
Yes as it mimics the bahavior of a close connection during a tx.

vietj:
I've reworked the current code to use the transaction status returned by Postgres after every command execution instead of relying on the query error itself, the tx status can be IDLE, ACTIVE or FAILED.

when PG reports the status to be failed then the current transaction is rollbacked, usually when this happens any subsequent execution fails with _current transaction is aborted, commands ignored until
 end of transaction block_

vietj:
I've merged this PR into master

vietj:
There is not support yet for save points, but it could be done later, I just wanted to have the minimum viable and correct API for tx.

PR #45, opened 2017-12-18
Pub-sub support - fixes #36

### LINKED ISSUE #36: Pub-sub support

vietj:
@pmlopes I remember you were interested by this feature, can you have a brief look ?

### DIFF
--- a/src/main/java/com/julienviet/pgclient/impl/PgPoolImpl.java
+++ b/src/main/java/com/julienviet/pgclient/impl/PgPoolImpl.java
@@ -93,6 +93,11 @@ public class PgPoolImpl extends PgClientBase<PgPoolImpl> implements PgPool {
     protected abstract void onFailure(Throwable cause);
 
     @Override
+    public void handleNotification(int processId, String channel, String payload) {
+      // What should we do ?
+    }
+
+    @Override
     public void handle(AsyncResult<Connection> ar) {
       if (ar.succeeded()) {
         Connection conn = ar.result();
--- a/src/main/java/com/julienviet/pgclient/impl/pubsub/PgSubscriberImpl.java
+++ b/src/main/java/com/julienviet/pgclient/impl/pubsub/PgSubscriberImpl.java
@@ -29,6 +29,7 @@ import java.util.stream.Collectors;
 
 public class PgSubscriberImpl implements PgSubscriber {
 
+  private static Logger log = LoggerFactory.getLogger(PgSubscriberImpl.class);
   private static final Function<Integer, Long> DEFAULT_RECONNECT_POLICY = count -> -1L;
 
   private final Vertx vertx;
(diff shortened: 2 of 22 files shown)


src/main/java/com/julienviet/pgclient/impl/PgPoolImpl.java
```diff
@@ -92,6 +92,11 @@ protected void onFailure(Throwable cause) {
 
     protected abstract void onFailure(Throwable cause);
 
+    @Override
+    public void handleNotification(int processId, String channel, String payload) {
+      // What should we do ?
```
pmlopes: Maybe route to active connections if they have a notification handler? This also means that returning to the pool it should clear the handlers

src/main/java/com/julienviet/pgclient/impl/pubsub/PgSubscriberImpl.java
```diff
@@ -0,0 +1,320 @@
...
+        }
+      }
+    }
+
+    void remove(ChannelImpl sub) {
+      subs.remove(sub);
+      if (subs.isEmpty()) {
+        channels.remove(name, this);
+        if (conn != null) {
+          conn.query("UNLISTEN " + name, ar -> {});
```
pmlopes: Are we sure we want to ignore the result? I guess that errors can be important

src/main/java/com/julienviet/pgclient/impl/pubsub/PgSubscriberImpl.java
```diff
@@ -0,0 +1,320 @@
...
+      if (!subscribed) {
+        if (conn != null) {
+          subscribed = true;
+          String sql = "LISTEN " + name;
+          conn.query(sql, ar -> {
+            if (ar.succeeded()) {
+              Handler<Void> handler = sub.subscribeHandler;
+              if (handler != null) {
+                handler.handle(null);
+              }
```
pmlopes: Should we just ignore errors? Maybe at least log it?

src/main/java/com/julienviet/pgclient/impl/PgPoolImpl.java
vietj: I think that in a pool, notification should not be used because it is asynchronous

src/main/java/com/julienviet/pgclient/impl/pubsub/PgSubscriberImpl.java
vietj: I think logging is ok; the thing is that I don't see a valid reason why it would fail (at least for now)

src/main/java/com/julienviet/pgclient/impl/pubsub/PgSubscriberImpl.java
vietj: I think logging is ok; the thing is that I don't see a valid reason why it would fail (at least for now)

PR #54, opened 2018-02-12
Supports UUID data type

EmadAlblueshi:
Fixes https://github.com/vietj/reactive-pg-client/issues/53

### DIFF
--- a/src/main/java/com/julienviet/pgclient/impl/codec/DataType.java
+++ b/src/main/java/com/julienviet/pgclient/impl/codec/DataType.java
@@ -435,10 +435,20 @@ public class DataType<T> {
   public static DataType<Object> MACADDR8 = new DataType<>(Object.class,774);
 
   // UUID
-  public static DataType<String> UUID = new DataType<String>(String.class, 2950) {
+  public static DataType<java.util.UUID> UUID = new DataType<java.util.UUID>(java.util.UUID.class, 2950) {
     @Override
-    public String decodeText(int len, ByteBuf buff) {
-      return buff.readCharSequence(len, StandardCharsets.UTF_8).toString();
+    public java.util.UUID decodeText(int len, ByteBuf buff) {
+      return java.util.UUID.fromString(buff.readCharSequence(len, StandardCharsets.UTF_8).toString());
+    }
+    @Override
+    public java.util.UUID decodeBinary(int len, ByteBuf buff) {
+      return new java.util.UUID(buff.readLong(), buff.readLong());
+    }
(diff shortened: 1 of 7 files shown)


src/main/java/com/julienviet/pgclient/impl/codec/DataType.java
```diff
@@ -435,10 +435,20 @@ public void encodeText(Buffer value, ByteBuf buff) {
   public static DataType<Object> MACADDR8 = new DataType<>(Object.class,774);
 
   // UUID
-  public static DataType<String> UUID = new DataType<String>(String.class, 2950) {
+  public static DataType<java.util.UUID> UUID = new DataType<java.util.UUID>(java.util.UUID.class, 2950) {
     @Override
-    public String decodeText(int len, ByteBuf buff) {
-      return buff.readCharSequence(len, StandardCharsets.UTF_8).toString();
+    public java.util.UUID decodeText(int len, ByteBuf buff) {
+      return java.util.UUID.fromString(buff.readCharSequence(len, StandardCharsets.UTF_8).toString());
```
vietj: shouldn't we use instead byte[] ?

vietj:
it seems this PR reformats a lot of code

vietj:
in the tests, we need to avoid that

src/main/java/com/julienviet/pgclient/impl/codec/DataType.java
EmadAlblueshi: Here we are getting data as text

EmadAlblueshi:
@vietj which test Binary or Text tests ?

EmadAlblueshi:
@vietj ok I got it

src/main/java/com/julienviet/pgclient/impl/codec/DataType.java
vietj: ah ok makes sense :-) and the other is using 2 longs which makes 16 bytes

vietj: and anyway it's not really going to be a bottleneck

EmadAlblueshi: Yes :)

EmadAlblueshi:
@vietj is it :+1: now ?

vietj:
much cleaner, I believe we need to refactor tests to remove boiler plate code soon so it remains maintainable

EmadAlblueshi:
👍 Ok

PR #55, opened 2018-02-12
Adds preparedBatch tests

EmadAlblueshi:
Fixes https://github.com/vietj/reactive-pg-client/issues/47 except `Geometry` types because they are not supported yet.


vietj:
I think a single test class is enough for now instead of two

EmadAlblueshi:
@vietj :+1:

PR #67, opened 2018-03-06
Support BigDecimal for NUMERIC (Text) data type

EmadAlblueshi:
Fixes https://github.com/vietj/reactive-pg-client/issues/61

@vietj would you please review ?

### DIFF
--- a/src/main/java/com/julienviet/pgclient/impl/ArrayTuple.java
+++ b/src/main/java/com/julienviet/pgclient/impl/ArrayTuple.java
@@ -94,8 +94,12 @@ public class ArrayTuple extends ArrayList<Object> implements Tuple {
 
   @Override
   public BigDecimal getBigDecimal(int pos) {
-    if(get(pos) instanceof BigDecimal)
-      return (BigDecimal) get(pos);
+    Object val = get(pos);
+    if(val instanceof BigDecimal) {
+      return (BigDecimal) val;
+    } else if (val instanceof Number) {
+      return new BigDecimal(val.toString());
+    }
     return null;
   }
 
--- a/src/main/java/com/julienviet/pgclient/impl/ArrayTuple.java
+++ b/src/main/java/com/julienviet/pgclient/impl/ArrayTuple.java
@@ -22,6 +22,7 @@ import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 
+import java.math.BigDecimal;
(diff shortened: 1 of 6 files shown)


src/main/java/com/julienviet/pgclient/impl/ArrayTuple.java
```diff
@@ -91,6 +92,13 @@ public Double getDouble(int pos) {
     return null;
   }
 
+  @Override
+  public BigDecimal getBigDecimal(int pos) {
+    if(get(pos) instanceof BigDecimal)
```
vietj: I think we need also to handle the case where the object is an instanceof Number and convert it

src/main/java/com/julienviet/pgclient/impl/RowImpl.java
```diff
@@ -131,4 +132,9 @@ public OffsetDateTime getOffsetDateTime(String name) {
   public UUID getUUID(String name) {
     return getUUID(desc.columnIndex(name));
   }
+
+  @Override
+  public BigDecimal getBigDecimal(String name) {
+    return getBigDecimal(desc.columnIndex(name));
```
vietj: we need to handle the case where columnIndex returns -1

vietj:
I think what we are missing is a patch so that NUMERIC type is always a BigInteger and not sometimes to Double

src/main/java/com/julienviet/pgclient/impl/RowImpl.java
EmadAlblueshi: @vietj is this should be handled for `BigDecimal` only ?

EmadAlblueshi:
Now `NUMERIC` will always return `BigDecimal` which handles both cases `Integer/Decimal`

src/main/java/com/julienviet/pgclient/impl/RowImpl.java
vietj: no, it should always be handled, perhaps we should handle it in a separate patch

EmadAlblueshi: ok

vietj: it seems actually is should be fine, we should check at least these cases are tested.

vietj: https://github.com/vietj/reactive-pg-client/issues/68

EmadAlblueshi:
👍

PR #71, opened 2018-03-09
Handle NUMERIC type with a custom Number - fixes #70

### LINKED ISSUE #70: Handle NUMERIC type with a custom Number
motivation: currently NUMERIC type is mapped to `BigDecimal` which does not allow a full representability of the type (such as ∞).
changes: provide a specific subclass of `Number` (perhaps named `PgNumeric` ?) that wraps the value as a `BigDecimal` when possible otherwise indicates which specific value it is. This can then be converted by the `Row` class to other values or can be retrieved as is. This mapping shall be bidirectional.

vietj:


src/main/java/com/julienviet/pgclient/Numeric.java
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * The Postgres <i>NUMERIC</i> type.
+ */
+public final class Numeric extends Number {
+
+  /**
+   * Constant for the {@code NaN} value.
+   */
+  public static final Numeric NaN = new Numeric(null);
```
gunnarmorling: You also might consider to handle `Infinity` and `-Infinity`.

src/main/java/com/julienviet/pgclient/Numeric.java
```diff
@@ -0,0 +1,132 @@
...
+
+  private BigDecimal value;
```
gunnarmorling: Seems this could be final?

src/main/java/com/julienviet/pgclient/Numeric.java
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * Return a {@code Numeric} instance for the given {@code number}.
+   *
+   * @param number the number
+   * @return the {@code Numeric} value
+   */
+  public static Numeric create(Number number) {
```
gunnarmorling: How would this deal with `number` being null?

src/main/java/com/julienviet/pgclient/Numeric.java
```diff
@@ -0,0 +1,132 @@
...
+  public BigDecimal bigDecimalValue() {
+     return value;
+  }
+
+  /**
+   * @return  the numeric value represented by this object after conversion
+   *          to type {@code BigInteger}.
+   */
+  public BigInteger bigIntegerValue() {
+    return value.toBigInteger();
```
gunnarmorling: You check for value to not be null in other places. I'm not quite sure whether it can be nullable or not?

src/main/java/com/julienviet/pgclient/Numeric.java
vietj: does Postgres support them ? according to the documentation, only _Nan_ makes sense

src/main/java/com/julienviet/pgclient/Numeric.java
vietj: yes

src/main/java/com/julienviet/pgclient/Numeric.java
vietj: it should throw NullPointerException but it hasn't been bested

src/main/java/com/julienviet/pgclient/Numeric.java
vietj: good catch!

vietj: I will also add test for this

src/main/java/com/julienviet/pgclient/Numeric.java
EmadAlblueshi: NaN for Numeric only

but the floating-point types have several special values:

Infinity
-Infinity
NaN

vietj: are we talking of PG or Java types ?

EmadAlblueshi: PG types

EmadAlblueshi: https://www.postgresql.org/docs/9.6/static/datatype-numeric.html

vietj: ah cool, it was not clear from the docs I read, will update it

vietj: so actually infinite values are not supported for numeric (https://stackoverflow.com/questions/45603040/does-postgresql-numeric-type-support-infinity-and-infinity), so I don't think that we should handle them in this class except than rejecting the invalid values when creating a numeric from a Number

vietj:
I've addressed the feedback from the PR and simplified the internal of `Numeric` by just wrapping a non null `Number`

vietj:
please review again :-)

vietj:
ping @gunnarmorling @EmadAlblueshi

vietj:
@EmadAlblueshi is that fine for merging ?

src/main/java/com/julienviet/pgclient/Numeric.java
```diff
@@ -0,0 +1,162 @@
+/*
+ * Copyright (C) 2017 Julien Viet
```
gunnarmorling: Just seeing the stale date. FWIW, we got rid of the date in [headers in Hibernate](https://github.com/hibernate/hibernate-validator/blob/4f0c69d3c6edf1e3c99a1260237e69e1bbfc9ad1/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java#L1-L6) projects, as its just a hazzle to maintain.

src/main/java/com/julienviet/pgclient/Numeric.java
```diff
@@ -0,0 +1,162 @@
...
+  /**
+   * @return  the numeric value represented by this object after conversion
+   *          to type {@code BigDecimal}.
+   */
+  public BigDecimal bigDecimalValue() {
+    if (value instanceof BigDecimal) {
+      return (BigDecimal) value;
+    } else if (value instanceof BigInteger) {
+      return new BigDecimal((BigInteger)value);
+    } else if (isNaN()) {
```
gunnarmorling: Perhaps rather throw an exception in that case? Or document it at least, as getting null may be surprising otherwise.

src/main/java/com/julienviet/pgclient/Numeric.java
```diff
@@ -0,0 +1,162 @@
...
+  /**
+   * @return  the numeric value represented by this object after conversion
+   *          to type {@code BigInteger}.
+   */
+  public BigInteger bigIntegerValue() {
+    if (value instanceof BigInteger) {
+      return (BigInteger) value;
+    } else if (value instanceof BigDecimal) {
+      return ((BigDecimal)value).toBigInteger();
+    } else if (isNaN()) {
```
gunnarmorling: Same here.

src/main/java/com/julienviet/pgclient/Numeric.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+        return l.compareTo(r) == 0;
+      }
+    }
+    return false;
+  }
+
+  @Override
+  public int hashCode() {
+    return intValue();
```
gunnarmorling: Interesting to base it on the int value. It'll cause a poor distribution though if you got multiple numerics (1.1, 1.2 etc.) in a small range, perhaps rather base it on the `BigDecimal` value of this (it might also be calculated once and cached)?

vietj: I've done that because it provides a consistent value between Number

src/main/java/com/julienviet/pgclient/Numeric.java
vietj: I think it should be documented, that's a situation that the developer can also take in account using `isNaN()` method before.

src/main/java/com/julienviet/pgclient/Numeric.java
vietj: to have `Number.create(4).hashcode() == Number.create(new BigInteger(4)).hashcode()`

gunnarmorling: I see. Is there a need for that, though? These values aren't considered as equal as per `equals()`, so it seems advantageous to put them into different hash buckets actually? I was just wondering about a case where you'd end up with lots of closely related decimal values in the same hash bucket. Not sure whether it makes a difference in practice :)

vietj: I think we need it because equality implies hashcode are the same

gunnarmorling: Ah yes, indeed you _do_ consider the two Numerics as equal, so they must have the same hash. You'd have a better value distribution by basing the hash on the BigDecimal value, but the current impl should work.

PR #74, opened 2018-03-15
fix #73 NPE thrown in ExtendedQueryResultHandler

### LINKED ISSUE #73: ExtendedQueryResultHandler must check the async result is succeeded before unboxing its boolean value
I'm firing the same insert query two times and expect to get a database exception, but I get a NPE instead. It seems to be a autoboxing issue: when there is an exception, ExtendedQueryResultHandler.handle is called and res returns null which cannot be casted to primitive boolean.

Version: 0.5.2
Stacktrace:
```17:58:42.481 [vert.x-eventloop-thread-0] ERROR io.vertx.core.impl.ContextImpl - Unhandled exception java.lang.NullPointerException: null at com.julienviet.pgclient.impl.ExtendedQueryResultHandler.handle(ExtendedQueryResultHandler.java:45) ~[reactive-pg-client-0.5.2.jar:na] at com.julienviet.pgclient.impl.ExtendedQueryResultHandler.handle(ExtendedQueryResultHandler.java:24) ~[reactive-pg-client-0.5.2.jar:na] at com.julienviet.pgclient.impl.CommandBase.handleMessage(CommandBase.java:44) ~[reactive-pg-client-0.5.2.jar:na] at com.julienviet.pgclient.impl.QueryCommandBase.handleMessage(QueryCommandBase.java:51) ~[reactive-pg-client-0.5.2.jar:na] at com.julienviet.pgclient.impl.ExtendedQueryCommandBase.handleMessage(ExtendedQueryCommandBase.java:68) ~[reactive-pg-client-0.5.2.jar:na] at com.julienviet.pgclient.impl.ExtendedQueryCommand.handleMessage(ExtendedQueryCommand.java:31) ~[reactive-pg-client-0.5.2.jar:na] at com.julienviet.pgclient.impl.SocketConnection.handleMessage(SocketConnection.java:233) ~[reactive-pg-client-0.5.2.jar:na]```

### PR DESCRIPTION
**Pull request by jklingsporn**

### CODE REVIEW

**Comment by vietj** [comment]
res should not be null, can you provide a reproducer ?

**Comment by jklingsporn** [comment]
Just check `CommandBase` Line 44. If that doesn't convince you, I'll try to create a reproducer.

**Comment by vietj** [comment]
a reproducer would be great

**Comment by jklingsporn** [comment]
See the latest commit. I've added a test-case for this in `PgConnectionTestBase`. It fails without the fix.

**Comment by vietj** [comment]
awesome! I will review it

File: src/main/java/com/julienviet/pgclient/impl/ExtendedQueryResultHandler.java [hunk 1]
```diff
@@ -42,7 +42,7 @@ public void handleResult(PgResult<T> result) {
 
   @Override
   public void handle(AsyncResult<Boolean> res) {
-    suspended = res.result();
+    suspended = res.succeeded() && res.result();
```
vietj: so NPE came from unboxing actually

src/test/java/com/julienviet/pgclient/PgConnectionTestBase.java
```diff
@@ -230,7 +255,7 @@ public void testBatchUpdate(TestContext ctx) {
         batch.add(Tuple.of("val0", 1));
         batch.add(Tuple.of("val1", 2));
         ps.batch(batch, ctx.asyncAssertSuccess(result -> {
-          for (int i = 0;i < 2;i++) {
+          for (int i = 0; i < 2; i++) {
```
vietj: remove reformat of non related code

jklingsporn:
Weird, seems to be some IntelliJ magic happened in the background. Regarding unboxing, that's what I wrote in the [first place](https://github.com/vietj/reactive-pg-client/issues/73) xD

PR #78, opened 2018-03-17
JSON (not JSONB) JsonObject and JsonArray binary support - JSON scalars not implemented

willks:
Would we accept this? It's just JsonObject and JsonArray support. Scalar support for queries will come, I just need help with it as it's been difficult to find documentation for binding/parameterisation for JSON scalars. Tests are passing for JsonObject/JsonArray support.


vietj:
what is difficult with scalar ?

willks:
@vietj Need a bit of help regarding query syntax for bind parameters. All tests result in JSON not valid errors. I left code in comments for someone with more PostgreSQL JSON experience to adjust. I tried string values, quoted - but ran the same queries through dbVisualizer, the latter worked, but only when the bind parameters were double quoted - but the results were double quoted as well. dbVisualizer uses JDBC so I'm not taking that as a basis. I've tried to find documentation from the official docs and postgresql forums for bind examples - but did not manage to find any meaningful examples. What I need is  a parameterised example to complete. 

If the code is not sufficient, please adjust or suggest how I can get the query parsing by server. At this point, I've got it working for my project's scenario - as I never query JSON but I accept many are. It's just a small step.

vietj:
I will have a look and reuse as much from your PR as possible

EmadAlblueshi:
I think below are all cases for `JSON`

`SELECT 'null'::JSON, '1.5'::JSON ,'1'::JSON, 'true'::JSON, 'false'::JSON, '"String"'::JSON, '[]'::JSON, '{}'::JSON`

The Pg `JSON` is text based like `TEXT` so what we need is `DataType.decodeJson` but for encoding we should have `DataType.encodeJson`

vietj:
I've implemented it in master, can you check ?

PR #86, opened 2018-03-22
Added binary support for some array types

counter567:
I added the support for most types in binary mode. Please correct me if I'm wrong somewhere.


vietj:
tests don't pass

vietj:
https://travis-ci.org/counter567/reactive-pg-client/builds/356858147?utm_source=email&utm_medium=notification

vietj:
it seems that Travis reports an issue, did you test it ?

counter567:
Sure, I ran the test multiple times without errors. It is the generation of the examples that fails

vietj:
I will have a look locally then

counter567:
Yeah its just the upload of the artifact:
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:2.8.2:deploy Return code is: 401, ReasonPhrase: Unauthorized

vietj:
good catch, will try to tweak the travis CI for that

vietj:
can you update the readme file with the binary type that become supported ?

counter567:
Sure, no problem.

vietj:
do you intent to contribute support for missing array types ?

counter567:
Textual support is going to follow, but I think earliest possible time is next weekend

counter567:
Yeah, also support for missing array types

src/main/java/io/reactiverse/pgclient/Tuple.java
```diff
@@ -332,6 +328,129 @@ static Tuple of(Object... elements) {
   @GenIgnore
   BigDecimal getBigDecimal(int pos);
 
+  @GenIgnore
+  Integer[] getIntArray(int pos);
```
vietj: should be called `getIntegerArray`

vietj:
for performance reason, we need to use primitive arrays when possible, i.e `Integer[]` -> `int[]` which are consume less heap and have guaranteed locality

vietj:
you need to the tests check to use the `ColumnChecker` as they make the tests easier to write and read

counter567:
I had a deeper look into it and I will add the functionality to check if arrays are equal

vietj:
it looks really nice now, I need to check it out locally for more review

counter567:
Sure, thanks :)

vietj:
right now tests use a single element array, I think instead we have different sizes : 0, 1, 2 at least

counter567:
I actually had one test with 2 elements. 
The handling does not differ between the array types, so I just added more elements to some tests.

vietj:
can you also do `0` elements ?

counter567:
Done

vietj:
I think we can refactor code in DataType like:
```
  interface ArrayFactory<T> {
    T[] create(int size);
  }
```

then

```
    public static <T> T[] decodeBinary(ArrayFactory<T> supplier, DataType<T> type, int len, ByteBuf buff) {
      if (len == 12) {
        return supplier.create(0);
      }
      buff.readerIndex(buff.readerIndex() + 4);
      int offset = buff.readInt();
      buff.readerIndex(buff.readerIndex() + 4);
      int length = buff.readInt();
      T[] array = supplier.create(length);
      buff.readerIndex(buff.readerIndex() + offset + 4);
      for (int i = 0; i < array.length; i++) {
        array[i] = type.decodeBinary(buff.readInt(), buff);
      }
      return array;
    }
```

and then 

```
  public static final DataType<String[]> BPCHAR_ARRAY = new DataType<String[]>(String[].class, 1014) {

    @Override
    public String[] decodeBinary(int len, ByteBuf buff) {
      ArrayFactory<String> a = String[]::new;
      return decodeBinary(a, DataType.BPCHAR, len, buff);
    }
    ...
  };
```

vietj:
it would work for most types except primitive types

vietj:
@counter567 any news about the PR ? we are close to finish it and I would like to release it soon

counter567:
@vietj Sorry, did not receive the emails until now, not a great connection Here.
I am currrently away for the weekend. I will return tomorrow evening. I am going to implement your suggestion and push them.

counter567:
Implemented your suggestions and also did some refactoring for the encoding.

vietj:
can you remove the special case for empty arrays ? I believe the generic case shall handle it.

counter567:
Done

src/main/java/io/reactiverse/pgclient/impl/codec/DataType.java
```diff
@@ -79,8 +86,42 @@ public void encodeBinaryInternal(Boolean value, ByteBuf buff) {
...
     }
   };
+  //Commented version cause other datatypes are always the same procedure
+  public static final DataType<boolean[]> BOOL_ARRAY = new DataType<boolean[]>(boolean[].class, 1000) {
 
-  public static DataType<boolean[]> BOOL_ARRAY = new DataType<>(boolean[].class, 1000);
+    @Override
+    public boolean[] decodeBinary(int len, ByteBuf buff) {
+      //If length equals 12 then it is an empty array
+      if (len == 12) {
```
vietj: remove this special case

src/main/java/io/reactiverse/pgclient/impl/codec/DataType.java
```diff
@@ -99,7 +140,40 @@ public void encodeBinaryInternal(Short value, ByteBuf buff) {
     }
   };
 
-  public static DataType<short[]> INT2_ARRAY = new DataType<>(short[].class, 1005);
+  public static final DataType<short[]> INT2_ARRAY = new DataType<short[]>(short[].class, 1005) {
+
+    @Override
+    public short[] decodeBinary(int len, ByteBuf buff) {
+      if (len == 12) {
```
vietj: remove this special case

src/main/java/io/reactiverse/pgclient/impl/codec/DataType.java
```diff
@@ -118,7 +192,40 @@ public void encodeBinaryInternal(Integer value, ByteBuf buff) {
     }
   };
 
-  public static DataType<int[]> INT4_ARRAY = new DataType<>(int[].class, 1007);
+  public static final DataType<int[]> INT4_ARRAY = new DataType<int[]>(int[].class, 1007) {
+
+    @Override
+    public int[] decodeBinary(int len, ByteBuf buff) {
+      if (len == 12) {
```
vietj: remove this special case

src/main/java/io/reactiverse/pgclient/impl/codec/DataType.java
```diff
@@ -137,7 +244,40 @@ public void encodeBinaryInternal(Long value, ByteBuf buff) {
     }
   };
 
-  public static DataType<int[]> INT8_ARRAY = new DataType<>(int[].class, 1016);
+  public static final DataType<long[]> INT8_ARRAY = new DataType<long[]>(long[].class, 1016) {
+
+    @Override
+    public long[] decodeBinary(int len, ByteBuf buff) {
+      if (len == 12) {
```
vietj: remove this special case

src/main/java/io/reactiverse/pgclient/impl/codec/DataType.java
```diff
@@ -158,7 +298,40 @@ public void encodeBinaryInternal(Float value, ByteBuf buff) {
     }
   };
 
-  public static DataType<float[]> FLOAT4_ARRAY = new DataType<>(float[].class, 1021);
+  public static final DataType<float[]> FLOAT4_ARRAY = new DataType<float[]>(float[].class, 1021) {
+
+    @Override
+    public float[] decodeBinary(int len, ByteBuf buff) {
+      if (len == 12) {
```
vietj: remove this special case

src/main/java/io/reactiverse/pgclient/impl/codec/DataType.java
```diff
@@ -179,7 +352,40 @@ public void encodeBinaryInternal(Double value, ByteBuf buff) {
     }
   };
 
-  public static DataType<double[]> FLOAT8_ARRAY = new DataType<>(double[].class, 1022);
+  public static final DataType<double[]> FLOAT8_ARRAY = new DataType<double[]>(double[].class, 1022) {
+
+    @Override
+    public double[] decodeBinary(int len, ByteBuf buff) {
+      if (len == 12) {
```
vietj: remove this special case

src/main/java/io/reactiverse/pgclient/impl/codec/DataType.java
```diff
@@ -215,9 +421,55 @@ public Character decodeText(int len, ByteBuf buff) {
...
+      buff.setInt(index, buff.writerIndex() - 4 - index);
+    }
   };
+  // Single length character
+  public static final DataType<char[]> CHAR_ARRAY = new DataType<char[]>(char[].class, 1002) {
 
-  public static DataType<Object> CHAR_ARRAY = new DataType<>(Object.class,1002);
+    @Override
+    public char[] decodeBinary(int len, ByteBuf buff) {
+      if (len == 12) {
```
vietj: remove this special case

src/main/java/io/reactiverse/pgclient/impl/codec/DataType.java
```diff
@@ -79,8 +86,38 @@ public void encodeBinaryInternal(Boolean value, ByteBuf buff) {
...
   };
+  //Commented version cause other datatypes are always the same procedure
+  public static final DataType<boolean[]> BOOL_ARRAY = new DataType<boolean[]>(boolean[].class, 1000) {
 
-  public static DataType<boolean[]> BOOL_ARRAY = new DataType<>(boolean[].class, 1000);
+    @Override
+    public boolean[] decodeBinary(int len, ByteBuf buff) {
+      buff.readerIndex(buff.readerIndex() + 4);//Skip dimensions cause its always 1
+      int offset = buff.readInt();//Read the offset, used to skip the bitmap
+      buff.readerIndex(buff.readerIndex() + 4);//Skip the oid cause we know what type it is
```
vietj: instead use `buff.skipBytes(4)`

src/main/java/io/reactiverse/pgclient/impl/codec/DataType.java
```diff
@@ -79,8 +86,38 @@ public void encodeBinaryInternal(Boolean value, ByteBuf buff) {
...
+      int length = buff.readInt(); //Read the length to create a fixed length array
+      boolean[] array = new boolean[length];//Create the array
+      buff.readerIndex(buff.readerIndex() + offset + 4);//Skip, if exists, the bitmap and the lower boundry
```
vietj: instead use `buff.skipBytes(4)`

vietj:
Commented to use `ByteBuf#skipBytes(n)` instead of moving forward the reader index in two ops

src/main/java/io/reactiverse/pgclient/impl/codec/DataType.java
counter567: That wont work, the default would expect the total number of elements in the array. But if the array is empty the header just contains the `0` dimensions, the `0` offset and the `OID`. The check at the start is in my opinion the most effective way to handle 0 size arrays

vietj: ok the check makes sense then

note: you commented on the `skipBytes` comment which is valid.

counter567: Oh, sorry did not mean to do that, yeah that part is valid.

src/main/java/io/reactiverse/pgclient/impl/codec/DataType.java
```diff
@@ -79,8 +86,41 @@ public void encodeBinaryInternal(Boolean value, ByteBuf buff) {
...
+    public boolean[] decodeBinary(int len, ByteBuf buff) {
+      if (len == 12) {
+        return new boolean[0];
+      }
+      buff.skipBytes(4);//Skip dimensions cause its always 1
+      int offset = buff.readInt();//Read the offset, used to skip the bitmap
+      buff.skipBytes(4);//Skip the oid cause we know what type it is
+      int length = buff.readInt(); //Read the length to create a fixed length array
+      boolean[] array = new boolean[length];//Create the array
+      buff.readerIndex(buff.readerIndex() + offset + 4);//Skip, if exists, the bitmap and the lower boundry
```
vietj: use `skipBytes(n)` as well

src/main/java/io/reactiverse/pgclient/impl/codec/DataType.java
```diff
@@ -99,7 +139,40 @@ public void encodeBinaryInternal(Short value, ByteBuf buff) {
     }
   };
 
-  public static DataType<short[]> INT2_ARRAY = new DataType<>(short[].class, 1005);
+  public static final DataType<short[]> INT2_ARRAY = new DataType<short[]>(short[].class, 1005) {
+
+    @Override
+    public short[] decodeBinary(int len, ByteBuf buff) {
+      if (len == 12) {
+        return new short[0];
```
vietj: use a single instance of zero length array since they are immutable

counter567: That works for the primitive types but not for the Objects which use the ArrayFactory. Or am I missing something?

vietj:
that's still possible, the ArrayFactory can provide it as an extra method or as a field (meaning it becomes an AbstractClass)

vietj:
looks great, thanks for the contrib

PR #87, opened 2018-03-26
Provide the ability to configure from env variable

BillyYccc:
See the issue https://github.com/reactiverse/reactive-pg-client/issues/39.

Supported Environment Variables:
* PGHOST
* PGHOSTADDR
* PGPORT
* PGDATABASE
* PGUSER
* PGPASSWORD

### DIFF
--- a/src/main/docs/index.md
+++ b/src/main/docs/index.md
@@ -51,7 +51,7 @@ Here is the simplest way to connect, query and disconnect
 ```
 
 You can also use environment variables to set default connection setting values, this is useful
-when you want to avoid hard-coding database connection information. 
+when you want to avoid hard-coding database connection information. You can refer to the [official documentation](https://www.postgresql.org/docs/9.6/static/libpq-connect.html#LIBPQ-CONNSTRING) for more details. 
 The following parameters are supported:
 
 * `PGHOST`
--- a/src/main/docs/index.md
+++ b/src/main/docs/index.md
@@ -54,12 +54,12 @@ You can also use environment variables to set default connection setting values,
 when you want to avoid hard-coding database connection information. 
 The following parameters are supported:
 
-* PGHOST
-* PGHOSTADDR
-* PGPORT
-* PGDATABASE
-* PGUSER
-* PGPASSWORD
+* `PGHOST`
+* `PGHOSTADDR`
+* `PGPORT`
+* `PGDATABASE`
+* `PGUSER`
+* `PGPASSWORD`
 
 ## Connecting to Postgres
 
--- a/src/main/docs/index.md
+++ b/src/main/docs/index.md
@@ -50,7 +50,7 @@ Here is the simplest way to connect, query and disconnect
 {@link examples.Examples#gettingStarted()}
 ```
 
-Environment variables can also be used to set default connection setting values, this is useful
+You can also use environment variables to set default connection setting values, this is useful
 when you want to avoid hard-coding database connection information. 
 The following parameters are supported:
 
--- a/src/main/docs/index.md
+++ b/src/main/docs/index.md
@@ -50,6 +50,17 @@ Here is the simplest way to connect, query and disconnect
 {@link examples.Examples#gettingStarted()}
 ```
 
+Environment variables can also be used to set default connection setting values, this is useful
+when you want to avoid hard-coding database connection information. 
+The following parameters are supported:
+
+* PGHOST
+* PGHOSTADDR
+* PGPORT
+* PGDATABASE
+* PGUSER
+* PGPASSWORD
+
 ## Connecting to Postgres
 
 Most of the time you will use a pool to connect to Postgres:
--- a/src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
+++ b/src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
@@ -56,7 +56,7 @@ public class PgConnectOptions extends NetClientOptions {
     } else {
       host = getenv("PGHOSTADDR");
     }
-    if (getenv("PGPORT") != null) port = parseInt(getenv("PGPORT");
+    if (getenv("PGPORT") != null) port = parseInt(getenv("PGPORT"));
     if (getenv("PGDATABASE") != null) database = getenv("PGDATABASE");
     if (getenv("PGUSER") != null) username = getenv("PGUSER");
     if (getenv("PGPASSWORD") != null) password = getenv("PGPASSWORD");
--- a/src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
+++ b/src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
@@ -24,6 +24,9 @@ import io.vertx.core.net.*;
 
 import java.util.Set;
 
+import static java.lang.Integer.*;
+import static java.lang.System.*;
+
 /**
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
  */
@@ -38,11 +41,11 @@ public class PgConnectOptions extends NetClientOptions {
   public static final boolean DEFAULT_CACHE_PREPARED_STATEMENTS = false;
   public static final int DEFAULT_PIPELINING_LIMIT = 256;
 
-  private String host = DEFAULT_HOST;
-  private int port = DEFAULT_PORT;
-  private String database = DEFAULT_DATABASE;
-  private String username = DEFAULT_USERNAME;
-  private String password = DEFAULT_PASSWORD;
+  private String host = getenv("PGHOSTADDR") == null ? ((getenv("PGHOST") == null) ? DEFAULT_HOST : getenv("PGHOST")) : getenv("PGHOSTADDR");
+  private int port = getenv("PGPORT") == null ? DEFAULT_PORT : parseInt(getenv("PGPORT"));
+  private String database = getenv("PGDATABASE") == null ? DEFAULT_DATABASE : getenv("PGDATABASE");
+  private String username = getenv("PGUSER") == null ? DEFAULT_USERNAME : getenv("PGUSER");
+  private String password = getenv("PGPASSWORD") == null ? DEFAULT_PASSWORD : getenv("PGPASSWORD");
   private boolean cachePreparedStatements = DEFAULT_CACHE_PREPARED_STATEMENTS;
   private int pipeliningLimit = DEFAULT_PIPELINING_LIMIT;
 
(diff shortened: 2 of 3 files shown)


vietj:
actually it should rather be `DEFAULT_HOST ` that is retrieved from `ENV` otherwise it is assigned the default value.

BillyYccc:
I took a look at the document in the https://www.postgresql.org/docs/9.6/static/libpq-connect.html#LIBPQ-PARAMKEYWORDS, it states:
> If host is specified without hostaddr, a host name lookup occurs.

> If hostaddr is specified without host, the value for hostaddr gives the server network address. The connection attempt will fail if the authentication method requires a host name.

> If both host and hostaddr are specified, the value for hostaddr gives the server network address. The value for host is ignored unless the authentication method requires it, in which case it will be used as the host name.

So I think the priority should be : 1. hostaddr env --> 2. host env 3. default_host
WDYT?

vietj:
yes, but you don't need to resolve host because vertx will do it.

BillyYccc:
I know this and I don't do the resolving myself, I'm just giving the client a possibility to set an IP address through the env. What if some user wants to connect the pg with an ip address.
What I mean here is something like this:
```
if(env_hostaddr == null){
  if(env_host == null){
    host = DEFAULT_HOST;
  }else{
    host = env_host;
  }
}else{
  host = env_hostaddr;
}
```

Then let vertx connect the host(or hostaddr), resolve it
`netclient.connect(port, host, etc...)`

I just give an alternative of setting the hostaddr, hope I make this clear this time...

vietj:
it looks fine, then

perhaps it should be in the default constructor so

BillyYccc:
I did an adjustment, is this what's expected? @vietj

vietj:
did you test it ?

can you use curly brace ?

BillyYccc:
I'm sorry... Just miss this

BillyYccc:
I did a simple test, it should take effect now.

vietj:
you need to document it properly in the `package-info`

BillyYccc:
I have added the document and tested it with my local machine through the Jekyll rendering, it looks good now.

src/main/docs/index.md
```diff
@@ -50,6 +50,17 @@ Here is the simplest way to connect, query and disconnect
 {@link examples.Examples#gettingStarted()}
 ```
 
+Environment variables can also be used to set default connection setting values, this is useful
```
vietj: rather word it : "You can also use environment variables to set ..."

BillyYccc: Fixed in https://github.com/reactiverse/reactive-pg-client/pull/87/commits/1d086c389ecc5f521a9482fbe72ebede6625f81b.

src/main/docs/index.md
```diff
@@ -50,6 +50,17 @@ Here is the simplest way to connect, query and disconnect
 {@link examples.Examples#gettingStarted()}
 ```
 
+You can also use environment variables to set default connection setting values, this is useful
+when you want to avoid hard-coding database connection information. 
+The following parameters are supported:
+
+* PGHOST
```
vietj: use backticks for the values, i.e `PGHOST`

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
```diff
@@ -48,6 +51,15 @@
 
   public PgConnectOptions() {
     super();
+    if (getenv("PGHOSTADDR") == null) {
+      host = getenv("PGHOST") == null ? DEFAULT_HOST : getenv("PGHOST");
+    } else {
+      host = getenv("PGHOSTADDR");
+    }
+    if (getenv("PGPORT") != null) port = parseInt(getenv("PGPORT"));
```
vietj: curly braces

vietj: should catch NumberFormatException to use the default value in case

src/main/docs/index.md
BillyYccc: Fixed in https://github.com/reactiverse/reactive-pg-client/pull/87/commits/bfbb7c37eadd99ebd74769a9121341b68ff36d32.

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
BillyYccc: Fixed in https://github.com/reactiverse/reactive-pg-client/pull/87/commits/2d25295cca19ce3f8217f1d7995a0d8f80cdd69a.

BillyYccc:
I have updated this, I also make the `PgConnectOptions(JsonObject json)` and `PgConnectOptions(PgConnectOptions other)` two constructors able to configure through the env variables.

In addition, I'm sorry I have revised for so many times, maybe need to squash commits to one?

vietj:
@BillyYccc  the copy constructor should not configure via env variables as it provides a copy of the current state.

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
```diff
@@ -48,15 +51,18 @@
 
   public PgConnectOptions() {
     super();
+    configureWithEnvVars();
```
vietj: call it `init()` instead

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
```diff
@@ -48,15 +51,18 @@
...
 
   public PgConnectOptions(JsonObject json) {
     super(json);
+    configureWithEnvVars();
     PgConnectOptionsConverter.fromJson(json, this);
   }
 
   public PgConnectOptions(PgConnectOptions other) {
     super(other);
+    configureWithEnvVars();
```
vietj: don't call for this

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
```diff
@@ -340,7 +339,7 @@ public PgConnectOptions setEnabledSecureTransportProtocols(Set<String> enabledSe
   /**
    * Configure the connection with environment variables.
```
vietj: Actually it configures from non-env variable as well when they are not set and the current doc does not reflect that

BillyYccc: https://github.com/reactiverse/reactive-pg-client/pull/87/commits/72303c6ec0eea12ee68b2fe1b1ac1da2c0e6fad6 is this appropriate?

vietj: rather : initialize with the default options, if env variables are specified they take precedence over these options.

BillyYccc: Fixed as you said.

src/main/docs/index.md
```diff
@@ -50,6 +50,17 @@ Here is the simplest way to connect, query and disconnect
 {@link examples.Examples#gettingStarted()}
 ```
 
+You can also use environment variables to set default connection setting values, this is useful
+when you want to avoid hard-coding database connection information.
```
vietj: I think we are missing a link to the libpg page that defines these env variables

BillyYccc: so maybe I need to add something like:
 > You can refer to the [official documentation](https://www.postgresql.org/docs/9.6/static/libpq-connect.html#LIBPQ-CONNSTRING) for more details.

this is ok?

vietj: sounds good

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
```diff
@@ -330,4 +335,31 @@ public PgConnectOptions setTcpQuickAck(boolean tcpQuickAck) {
...
+    if (getenv("PGHOSTADDR") == null) {
+      host = getenv("PGHOST") == null ? DEFAULT_HOST : getenv("PGHOST");
+    } else {
+      host = getenv("PGHOSTADDR");
+    }
+    if (getenv("PGPORT") != null) {
+      try {
+        port = parseInt(getenv("PGPORT"));
+      } catch (NumberFormatException e) {
+        port = DEFAULT_PORT;
```
vietj: no need for that, port is already set

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
```diff
@@ -330,4 +335,31 @@ public PgConnectOptions setTcpQuickAck(boolean tcpQuickAck) {
   public PgConnectOptions setEnabledSecureTransportProtocols(Set<String> enabledSecureTransportProtocols) {
     return (PgConnectOptions) super.setEnabledSecureTransportProtocols(enabledSecureTransportProtocols);
   }
+
+  /**
+   * Initialize with the default options, if env variables are specified they take precedence over these options.
+   */
+  private void init() {
+    if (getenv("PGHOSTADDR") == null) {
+      host = getenv("PGHOST") == null ? DEFAULT_HOST : getenv("PGHOST");
```
vietj: no need to set default_host, it's already set

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
BillyYccc: I use a comment to explain what the port will be.

src/main/docs/index.md
```diff
@@ -50,6 +50,17 @@ Here is the simplest way to connect, query and disconnect
...
+when you want to avoid hard-coding database connection information. You can refer to the [official documentation](https://www.postgresql.org/docs/9.6/static/libpq-connect.html#LIBPQ-CONNSTRING) for more details. 
+The following parameters are supported:
+
+* `PGHOST`
+* `PGHOSTADDR`
+* `PGPORT`
+* `PGDATABASE`
+* `PGUSER`
+* `PGPASSWORD`
+
```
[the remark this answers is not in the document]
vietj: well those are env variable and are defined by libpq, not us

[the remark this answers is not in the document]
EmadAlblueshi: Ok 👍🏻

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
```diff
@@ -330,4 +335,33 @@ public PgConnectOptions setTcpQuickAck(boolean tcpQuickAck) {
   public PgConnectOptions setEnabledSecureTransportProtocols(Set<String> enabledSecureTransportProtocols) {
     return (PgConnectOptions) super.setEnabledSecureTransportProtocols(enabledSecureTransportProtocols);
   }
+
+  /**
+   * Initialize with the default options, if env variables are specified they take precedence over these options.
+   */
+  private void init() {
```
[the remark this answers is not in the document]
BillyYccc: see this in https://github.com/reactiverse/reactive-pg-client/pull/87/commits/3360c309dacaa6ccbb40e90e0b7f566029274c12.

BillyYccc:
Thank you for your reviewing and patience! @vietj

vietj:
thanks for your contribution @BillyYccc

PR #92, opened 2018-03-29
Enable the client to configure from a connection URI

BillyYccc:
This PR is for https://github.com/reactiverse/reactive-pg-client/issues/40.

I did some investigation and provide an ability to configure from URI, something like this:
```
// Connection URI
String connectionUri = "postgresql://dbuser:secretpassword@database.server.com:3211/mydb";

// Create the pooled client
PgPool client = PgClient.pool(connectionUri);
```

The details handling the invalid URI is complicated, I think we need further discussions about this.

### DIFF
--- a/src/main/docs/index.md
+++ b/src/main/docs/index.md
@@ -50,6 +50,17 @@ Here is the simplest way to connect, query and disconnect
 {@link examples.Examples#gettingStarted()}
 ```
 
+You can also use environment variables to set default connection setting values, this is useful
+when you want to avoid hard-coding database connection information. You can refer to the [official documentation](https://www.postgresql.org/docs/9.6/static/libpq-connect.html#LIBPQ-CONNSTRING) for more details. 
+The following parameters are supported:
+
+* `PGHOST`
+* `PGHOSTADDR`
+* `PGPORT`
+* `PGDATABASE`
+* `PGUSER`
+* `PGPASSWORD`
+
 ## Connecting to Postgres
 
 Most of the time you will use a pool to connect to Postgres:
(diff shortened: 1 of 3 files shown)


vietj:
out of curiosity : why is equals/hashcode needed on options ?

BillyYccc:
This is for unit tests.

BillyYccc:
And maybe looks cleaner like the Father class `NetClientOptions` , etc...

vietj:
thats right. implementing these is always painful and I wanted to ahve the generation of these in the *Converter in vertx

vietj:
please see : https://github.com/iceddev/pg-connection-string for the details

vietj:
what do you want to discuss about invalid URIs ?

BillyYccc:
I have just tried with the library above, but I find it can not handle the reserve character well.
For instance:
If you give a string `'postgresql://$%^{}[]/qwer'`, then output is
```
{ port: null,
  host: '',
  database: '$%^{}[]/qwer',
  user: '',
  password: '' }
```

My implementation is based on `java.net.URI`, so far most of the cases are fine, except for this one https://github.com/BillyYccc/vertx-pg-client/blob/ed62aa57ecb6b51411067997ad8e647f01b5a2ef/src/test/java/io/reactiverse/pgclient/PgConnectionUriParserTest.java#L196
This URI is valid in the [libpq definition](https://www.postgresql.org/docs/9.6/static/libpq-connect.html#LIBPQ-CONNSTRING), but invalid in the java uri library, so extra effort are needed, this is an annoying case...

BillyYccc:
In addition, the generating function in Intellij is powerful, it can save a lot of work for me to generate equals(), maybe what we need is an IDE plugin for Vert.x...

vietj:
`java.net.URI` is really a pita

vietj:
I think we should simply try to do it with a regular expression

PR #93, opened 2018-03-29
Enable the client to configure from a connection URI - fixes #40

### LINKED ISSUE #40: Configure from URI
e.g : `postgresql://dbuser:secretpassword@database.server.com:3211/mydb`

BillyYccc:
Add feature for connecting with connection URI. Fixes https://github.com/reactiverse/reactive-pg-client/issues/40.

### DIFF
--- a/src/main/docs/index.md
+++ b/src/main/docs/index.md
@@ -61,6 +61,12 @@ The following parameters are supported:
 * `PGUSER`
 * `PGPASSWORD`
 
+We also provide you an alternative way to connect when you want to configure with a connection URI:
+
+```$lang
+{@link examples.Examples#connectingWithUri}
+```
+
 ## Connecting to Postgres
 
 Most of the time you will use a pool to connect to Postgres:
--- a/src/main/docs/index.md
+++ b/src/main/docs/index.md
@@ -67,6 +67,8 @@ We also provide you an alternative way to connect when you want to configure wit
 {@link examples.Examples#connectingWithUri}
 ```
 
+More information about connection string formats can be found in the [PostgreSQL Manuals](https://www.postgresql.org/docs/9.6/static/libpq-connect.html#LIBPQ-CONNSTRING).
+
 ## Connecting to Postgres
 
 Most of the time you will use a pool to connect to Postgres:
--- a/src/main/docs/index.md
+++ b/src/main/docs/index.md
@@ -64,7 +64,7 @@ The following parameters are supported:
 We also provide you an alternative way to connect when you want to configure with a connection URI:
 
 ```$lang
-{@link examples.Examples#connectingWithUri}
+{@link examples.Examples#configureFromUri}
 ```
 
 More information about connection string formats can be found in the [PostgreSQL Manuals](https://www.postgresql.org/docs/9.6/static/libpq-connect.html#LIBPQ-CONNSTRING).
--- a/src/main/java/io/reactiverse/pgclient/PgClient.java
+++ b/src/main/java/io/reactiverse/pgclient/PgClient.java
@@ -27,8 +27,6 @@ import io.vertx.core.*;
 
 import java.util.List;
 
-import static io.reactiverse.pgclient.PgConnectOptionsProvider.*;
-
 /**
  * Defines the client operations with a Postgres Database.
  *
@@ -86,13 +84,6 @@ public interface PgClient {
         connect(vertx, options, handler);
       });
     }
-  }
-
-  /**
-   * Like {@link #connect(Vertx, PgConnectOptions, Handler)} but with a specific connection URI to configure the connection.
-   */
-  static void connect(Vertx vertx, String connectionUri, Handler<AsyncResult<PgConnection>> handler) {
-    connect(vertx, fromUri(connectionUri), handler);
   }
 
   /**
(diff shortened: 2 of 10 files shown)


BillyYccc:
Hi! I have reworked to implement based on regular expressions, I tidy the commit records and this PR is ready to be reviewed.

src/main/java/io/reactiverse/pgclient/impl/PgConnectionUriParser.java
```diff
@@ -0,0 +1,221 @@
...
+  private static int NET_LOCATION_GROUP = 2;
+  private static int PORT_GROUP = 3;
+  private static int DATABASE_GROUP = 4;
+  private static int PARAMETER_GROUP = 5;
+
+  private final String connectionUri;
+
+  private JsonObject configuration = new JsonObject();
+
+  private PgConnectionUriParser(String connectionUri) {
```
vietj: make everything static  and pass the object along the way in methods

BillyYccc:
I doubt we should not provide configuration by 
`PgConnectOptions options = PgConnectionOptionsProvider.fromEnv();`
because the https://www.postgresql.org/docs/9.6/static/libpq-envars.html said that 
> The following environment variables can be used to select default connection parameter values, which will be used by PQconnectdb, PQsetdbLogin and PQsetdb **if no value is directly specified by the calling code.**


I think the priority is When options is specified, we use the specified one; if not specified, we use the env vars if the env vars exist, only when the option is not given and env vars do not exist, we use the default values finally.

BillyYccc:
`PgConnectOptions options = PgConnectionOptionsProvider.fromUri(uri);`
This is a very appealing idea and I find this is also used in [libbq](https://www.postgresql.org/docs/9.6/static/libpq-connect.html).
> `PQconninfoOption *PQconninfoParse(const char *conninfo, char **errmsg);`

The connection string can either a connection uri like
 `postgresql://[user[:password]@][netloc][:port][/dbname][?param1=value1&...]` 
or a Keyword/Value Connection Strings e.g. 
` host=localhost port=5432 dbname=mydb connect_timeout=10`

So we can achieve one 
```
PgConnectOptions options = PgConnectionOptionsProvider.fromUri(uri);
PgConnectOptions options = PgConnectionOptionsProvider.fromKeyValues(kvString);
```

What do you think of this?

vietj:
@BillyYccc I mean to set other settings like connection pool size, proxy, etc...

vietj:
for now let's go static for the URI / keypairs

vietj:
and also uses `PgConnectOptionsProvider` static methods

src/main/java/io/reactiverse/pgclient/impl/PgConnectionUriParser.java
BillyYccc: changes are in https://github.com/reactiverse/reactive-pg-client/pull/93/commits/92afd499d8a45971da47ebfb5ddf23fc505c176f.

BillyYccc:
I also think we need a part `Configuration` on purpose in the documentation between the part `Getting started` and `Connecting to Postgres` to clarify the 3 ways of configuring:
1. setter 
2. uri 
3. env

vietj:
yes but we can see that in another PR

BillyYccc:
Do I need to add a same license like other files in the new file header ?

vietj:
yes and you need to agree on the copyright terms

src/main/docs/index.md
```diff
@@ -61,6 +61,12 @@ The following parameters are supported:
 * `PGUSER`
 * `PGPASSWORD`
 
+We also provide you an alternative way to connect when you want to configure with a connection URI:
```
vietj: add a reference to the documentation that explains the connection string format

src/main/java/io/reactiverse/pgclient/PgClient.java
```diff
@@ -55,6 +57,23 @@ static PgPool pool(Vertx vertx, PgPoolOptions options) {
     return new PgPoolImpl(vertx, false, options);
   }
 
+  /**
+   * Create a connection pool to the database configured with the given {@code connectionUri}.
+   *
+   * @param connectionUri the connection URI
+   * @return the connection pool
+   */
+  static PgPool pool(String connectionUri) {
```
vietj: we should not need both for the moment as user should do it itself using the provider

src/main/java/io/reactiverse/pgclient/PgClient.java
```diff
@@ -86,6 +88,13 @@ static void connect(Vertx vertx, PgConnectOptions options, Handler<AsyncResult<P
     }
   }
 
+  /**
+   * Like {@link #connect(Vertx, PgConnectOptions, Handler)} but with a specific connection URI to configure the connection.
+   */
+  static void connect(Vertx vertx, String connectionUri, Handler<AsyncResult<PgConnection>> handler) {
```
vietj: we need to remove this for now

vietj:
the invalid test should test catch exception, i.e invalid config should not return half configured options

BillyYccc:
Now I feel let the parser returns a null when parsing invalid uris seems like a bad idea because the user doesn't know what happened(we need to at least give some warning), maybe throwing an IllegalArgumentException is a better choice? @vietj

vietj:
looks nice, I'm going to merge it, thanks for the contribution!

BillyYccc:
Feeling grateful for your guidance, Thank you!
I'm going to make another PR to improve the document later.

PR #101, opened 2018-04-22
Support for Unix domain socket - fixes #100

### LINKED ISSUE #100: Can it use unix domain sockets?
App server and pg server often placed in one unix machine. Communication via unix domain sockets significantly improve performance. Does it use it internally?

BillyYccc:
Fixes #100.

I did some investigation and follow the socket address format, something like `${domainSocketAddress}/.s.PGSQL.${port}`, see the `unix_socket_directories (string)` part in [official documentation](https://www.postgresql.org/docs/9.6/static/runtime-config-connection.html)  for more details.

I have done some integration tests in my local machine because the embedded Postgres database does not support Unix Domain Socket, so are there any choices on how to test the connection for domain socket in the project tests?

### DIFF
--- a/src/main/java/examples/Examples.java
+++ b/src/main/java/examples/Examples.java
@@ -205,8 +205,8 @@ public class Examples {
   public void connecting06(Vertx vertx) {
 
     // Pool Options
+    // Socket file name will be /var/run/postgresql/.s.PGSQL.5432
     PgPoolOptions options = new PgPoolOptions()
-      .setDomainSocket(true)
       .setHost("/var/run/postgresql")
       .setPort(5432)
       .setDatabase("the-db");
@@ -215,7 +215,7 @@ public class Examples {
     PgPool client = PgClient.pool(options);
 
     // Create the pooled client with a vertx instance
-    // Make sure the vertx instance enables native transports
+    // Make sure the vertx instance has enabled native transports
     PgPool client2 = PgClient.pool(vertx, options);
   }
 
--- a/src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
+++ b/src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
@@ -56,11 +56,7 @@ public class PgConnectOptions extends NetClientOptions {
 
     if (getenv("PGHOSTADDR") == null) {
       if (getenv("PGHOST") != null) {
-        String host = getenv("PGHOST");
-        if (host.startsWith("/")) {
-          pgConnectOptions.setDomainSocket(true);
-        }
-        pgConnectOptions.setHost(host);
+        pgConnectOptions.setHost(getenv("PGHOST"));
       }
     } else {
       pgConnectOptions.setHost(getenv("PGHOSTADDR"));
@@ -93,7 +89,6 @@ public class PgConnectOptions extends NetClientOptions {
   public static final String DEFAULT_PASSWORD = "pass";
   public static final boolean DEFAULT_CACHE_PREPARED_STATEMENTS = false;
   public static final int DEFAULT_PIPELINING_LIMIT = 256;
-  public static final boolean DEFAULT_DOMAIN_SOCKET_OPTION = false;
 
   private String host;
   private int port;
@@ -102,7 +97,6 @@ public class PgConnectOptions extends NetClientOptions {
   private String password;
   private boolean cachePreparedStatements;
   private int pipeliningLimit;
-  private boolean domainSocket;
 
   public PgConnectOptions() {
     super();
@@ -123,7 +117,6 @@ public class PgConnectOptions extends NetClientOptions {
     user = other.user;
     password = other.password;
     pipeliningLimit = other.pipeliningLimit;
-    domainSocket = other.domainSocket;
   }
 
   public String getHost() {
@@ -189,15 +182,6 @@ public class PgConnectOptions extends NetClientOptions {
 
   public PgConnectOptions setCachePreparedStatements(boolean cachePreparedStatements) {
     this.cachePreparedStatements = cachePreparedStatements;
-    return this;
-  }
-
-  public boolean isDomainSocket() {
-    return domainSocket;
-  }
-
-  public PgConnectOptions setDomainSocket(boolean domainSocket) {
-    this.domainSocket = domainSocket;
     return this;
   }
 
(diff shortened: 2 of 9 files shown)


BillyYccc:
I'm now considering https://github.com/fabric8io/docker-maven-plugin to see if it helps with our integration tests for domain sockets.

vietj:
that would be great, I think if you do that then we could have it in CI with TravisCI that supports docker

BillyYccc:
Updated with the integration tests and docker support.

pom.xml
```diff
@@ -306,6 +313,68 @@
...
+                <PGDATABASE>postgres</PGDATABASE>
+                <PGUSER>postgres</PGUSER>
+                <PGPASSWORD>postgres</PGPASSWORD>
+              </environmentVariables>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>io.fabric8</groupId>
```
vietj: can we try to use TestContainer instead ? https://www.testcontainers.org

BillyYccc: I have migrated it to `TestContainer`. By the way, this is a really nice tool!

vietj: I haven't tried it yet but it looks promisinig for testing with docker

vietj:
I'm wondering if we should not simply drop the useDomainSocket boolean and parse the `host` so that if it looks like a path then we chose to configure domain sockets instead

src/main/java/examples/Examples.java
```diff
@@ -202,6 +202,23 @@ public void connecting05(Vertx vertx) {
     });
   }
 
+  public void connecting06(Vertx vertx) {
+
+    // Pool Options
+    PgPoolOptions options = new PgPoolOptions()
+      .setDomainSocket(true)
+      .setHost("/var/run/postgresql")
+      .setPort(5432)
```
vietj: port is irrelevant

BillyYccc: Please take a look at this: https://www.postgresql.org/docs/9.6/static/runtime-config-connection.html.
 It's said in `unix_socket_directories` part that 
socket file will be named as `${PATH}/.s.PGSQL.${SERVER_PORT}` , so the port is useful here.

BillyYccc:
Sorry I don't get the point above,  do you mean we don't need `useDomainSocket` boolean and decide how to connect by the value of  `host`? What do you expect to be achieved?

vietj:
we parse the `host` string and if it looks like an IP we use normal sockets and if it looks like a path we use domain sockets

vietj:
and give ability to use this in URLs

vietj:
https://insomnia.rest/blog/unix-domain-sockets/

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
```diff
@@ -452,7 +434,10 @@ public int hashCode() {
     result = 31 * result + password.hashCode();
     result = 31 * result + (cachePreparedStatements ? 1 : 0);
     result = 31 * result + pipeliningLimit;
-    result = 31 * result + (domainSocket ? 1 : 0);
     return result;
   }
+
+  public boolean isUsingDomainSocket() {
```
BillyYccc: Using this public method for letting https://github.com/reactiverse/reactive-pg-client/blob/6920ec7489b63dfe359e65eb57d226ecba01ac25/src/main/java/io/reactiverse/pgclient/PgClient.java#L72 to know whether a vertx instance should `setPreferNativeTransport()`

BillyYccc:
URI is supported as https://www.postgresql.org/docs/9.6/static/libpq-connect.html stated.

>  So, to specify a non-standard Unix-domain socket directory, either omit the host specification in the URI and specify the host as a parameter, or percent-encode the path in the host component of the URI:
```
postgresql:///dbname?host=/var/lib/postgresql
postgresql://%2Fvar%2Flib%2Fpostgresql/dbname
```

> Port number to connect to at the server host, or `socket file name extension for Unix-domain connections`.

vietj:
when I run the integration test on OSX I get : 

```
Caused by: org.rnorth.ducttape.RetryCountExceededException: Retry limit hit with exception
Caused by: org.testcontainers.containers.ContainerLaunchException: Could not create/start container
Caused by: java.lang.reflect.UndeclaredThrowableException
Caused by: java.lang.reflect.InvocationTargetException
Caused by: com.github.dockerjava.api.exception.DockerException: 
Mounts denied: 
The path /var/run/postgresql
is not shared from OS X and is not known to Docker.
You can configure shared paths from Docker -> Preferences... -> File Sharing.
See https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.
```

is it expected ?

BillyYccc:
@vietj I guess this may be relevant to the volume mount in OS X, this might help:  https://stackoverflow.com/questions/45122459/docker-mounts-denied-the-paths-are-not-shared-from-os-x-and-are-not-known, but I don't have a Mac right now to test it, so I'm not very certain about this.

vietj:
I'm on it.

vietj:
I can't make it pass on OSX because TestContainer actually use a shaded Netty and there is a conflict with JNI:

```
java.lang.UnsatisfiedLinkError: io.netty.channel.unix.ErrorsStaticallyReferencedJniMethods.errnoENOENT()I
	at io.netty.channel.unix.ErrorsStaticallyReferencedJniMethods.errnoENOENT(Native Method)
	at io.netty.channel.unix.Errors.<clinit>(Errors.java:37)
	at io.netty.channel.unix.FileDescriptor.<clinit>(FileDescriptor.java:49)
	at io.netty.channel.kqueue.Native.<clinit>(Native.java:62)
	at io.netty.channel.kqueue.KQueue.<clinit>(KQueue.java:33)
	at io.vertx.core.net.impl.transport.KQueueTransport.isAvailable(KQueueTransport.java:58)
	at io.vertx.core.net.impl.transport.Transport.nativeTransport(Transport.java:70)
```

vietj:
so I think we can try running without test containers and see if the actual integration can support domain sockets

BillyYccc:
So what is needed to be done now? I have tested on my local machine(Ubuntu), set up a postgres server directly and the client can communicate with database through the domain docket.

vietj:
I'm taking care of it to see if that's possible since I do have OSX

vietj:
actually I'm able to use the current database setup to run with domain sockets

vietj:
adding

--unix_socket_directories=/foo/bar.sock
should work

vietj:
update has been pushed in master

vietj:
I think it's time for a release

PR #132, opened 2018-07-19
Rethrow exceptions to the end user

EmadAlblueshi:
Fixes https://github.com/reactiverse/reactive-pg-client/issues/123

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -210,6 +210,12 @@ The *Reactive Postgres Client* currently supports the following data types
 |`j.t.OffsetDateTime[]`
 |&#10004;
 
+|`INTERVAL`
+|`i.r.p.d.Interval`
+|&#10004;
+|`i.r.p.d.Interval[]`
+|&#10004;
+
 |`BYTEA`
 |`i.v.c.b.Buffer`
 |&#10004;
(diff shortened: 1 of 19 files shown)


EmadAlblueshi:
@vietj 

```java
PgPool client = PgClient.pool(options);
client.query("SELECT 1,2", rs -> {
  if (rs.succeeded()) {
    Row r = rs.result().iterator().next();
    // here the IndexOutOfBoundsException will be thrown silently and should be caught by the user
    Integer value = r.getInteger(100);
  } else {
    rs.cause().printStackTrace();
  }
});
```
In this PR I only `Throwable#printStackTrace()` to inform the user what exactly is happening and I think the user should take care of closing the connection `client.close()` in case of any exception occurred.

I'm not sure about the current behavior in case of a thrown exception like `IndexOutOfBoundsException` in NON Vert.x applications who use the driver.

I have added tests for `Row#getXXX(int pos)` too.

Thanks!

EmadAlblueshi:
@vietj 

Now, user exceptions will be re-thrown in `PgConnection`  and `ConnectionPool` in case of `null` Holder with tests.

Issue #135, opened 2018-07-21
Get column name from Row


pratikpparikh:
Dear Authors,

Can we make it so that we can get the columnName from resultset rather than depending client to keep it context?  

Regards,
Pratik Parikh

PR #136, opened 2018-07-21
Supports getting column name from Row

pratikpparikh:
This is an enhancement for #135

### DIFF
--- a/src/main/java/io/reactiverse/pgclient/impl/RowImpl.java
+++ b/src/main/java/io/reactiverse/pgclient/impl/RowImpl.java
@@ -49,7 +49,7 @@ public class RowImpl extends ArrayTuple implements Row {
   @Override
   public String getColumnName(int pos) {
     final ColumnDesc[] columnDescs = desc.columns();
-    return pos == -1 || columnDescs.length < pos ? null : columnDescs[pos].getName();
+    return pos < 0 || columnDescs.length - 1 < pos ? null : columnDescs[pos].getName();
   }
 
   @Override
--- a/src/main/java/io/reactiverse/pgclient/impl/RowImpl.java
+++ b/src/main/java/io/reactiverse/pgclient/impl/RowImpl.java
@@ -20,6 +20,7 @@ package io.reactiverse.pgclient.impl;
 import io.reactiverse.pgclient.Json;
 import io.reactiverse.pgclient.Numeric;
 import io.reactiverse.pgclient.Row;
+import io.reactiverse.pgclient.impl.codec.ColumnDesc;
 import io.reactiverse.pgclient.impl.codec.decoder.RowDescription;
 import io.reactiverse.pgclient.data.Point;
 import io.vertx.core.buffer.Buffer;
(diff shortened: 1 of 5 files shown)


EmadAlblueshi:
Hi @pratikpparikh, 

I think we need test for `Row` in [RowTest](https://github.com/reactiverse/reactive-pg-client/blob/master/src/test/java/io/reactiverse/pgclient/RowTest.java)

src/main/java/io/reactiverse/pgclient/impl/RowImpl.java
```diff
@@ -45,6 +46,12 @@ public RowImpl(RowImpl row) {
     this.desc = row.desc;
   }
 
+  @Override
+  public String getColumnName(int pos) {
+    final ColumnDesc[] columnDescs = desc.columns();
+    return pos == -1 || columnDescs.length < pos ? null : columnDescs[pos].getName();
```
EmadAlblueshi: I think if pos < 0 would be better to make sure of all negative numbers

pratikpparikh: i am adding a negative test as well.

pratikpparikh:
@EmadAlblueshi we should be covered not with test cases and appropriate code check-in for this request.

src/test/java/io/reactiverse/pgclient/RowTest.java
```diff
@@ -97,6 +97,44 @@ public void testGetNonExistingRows(TestContext ctx) {
...
+        ctx.asyncAssertSuccess(result -> {
+          Row row = result.iterator().next();
+          ctx.assertEquals("foo",row.getColumnName(0));
+          async.complete();
+        }));
+    }));
+  }
+
+  @Test
+  public void testNegativeNotEqualGetColumnNameRows(TestContext ctx) {
```
EmadAlblueshi: Should rename the test method without **negative**

src/test/java/io/reactiverse/pgclient/RowTest.java
```diff
@@ -97,6 +97,44 @@ public void testGetNonExistingRows(TestContext ctx) {
...
+  }
+
+  @Test
+  public void testNegativeGetColumnNameRows(TestContext ctx) {
+    Async async = ctx.async();
+    PgClient.connect(vertx, options, ctx.asyncAssertSuccess(conn -> {
+      conn.query("SELECT 2 \"foo\"",
+        ctx.asyncAssertSuccess(result -> {
+          Row row = result.iterator().next();
+          ctx.assertEquals(null,row.getColumnName(1));
```
EmadAlblueshi: Should use `ctx.assertNull(row.getColumnName(-1))` instead

pratikpparikh: @EmadAlblueshi this is still a valid test case (edge length test case), maybe we need to add an additional test case.

src/test/java/io/reactiverse/pgclient/RowTest.java
pratikpparikh: @EmadAlblueshi okay.

src/test/java/io/reactiverse/pgclient/RowTest.java
EmadAlblueshi: @pratikpparikh Yes because this test should reflects its name :)

pratikpparikh: @EmadAlblueshi agreed.  The test cases are now named appropriately.

pratikpparikh: @EmadAlblueshi we should be good with test case naming now as well.

vietj:
I'll merge it

vietj:
thanks for the contribution!

PR #145, opened 2018-08-06
Use quoted identifiers in LISTEN / UNLISTEN

brosenfi:
Bug Fix: Using quoted identifiers for channel names would result in notifications not coming back to the channel since the PgNotification would contain a channel name without quotes.  Solved this by making all channel statements use quoted identifiers - this also matches up with use of a standard HashMap (case-sensitive) to match back to channels.

### DIFF
--- a/src/main/docs/index.md
+++ b/src/main/docs/index.md
@@ -400,6 +400,13 @@ provides per channel subscription:
 {@link examples.Examples#pubsub02(io.vertx.core.Vertx)}
 ```
 
+The channel name that is given to the channel method will be the exact name of the channel as held by Postgres for sending
+notifications.  Note this is different than the representation of the channel name in SQL, and
+internally {@link io.reactiverse.pgclient.pubsub.PgSubscriber} will prepare the submitted channel name as a quoted identifier:
+
+```$lang
+{@link examples.Examples#pubsub03(io.vertx.core.Vertx)}
+```
 You can provide a reconnect policy as a function that takes the number of `retries` as argument and returns an `amountOfTime`
 value:
 
@@ -408,7 +415,7 @@ value:
 * when `amountOfTime > 0`: the subscriber retries after `amountOfTime` milliseconds
 
 ```$lang
-{@link examples.Examples#pubsub03(io.vertx.core.Vertx)}
+{@link examples.Examples#pubsub04(io.vertx.core.Vertx)}
 ```
 
 The default policy is to not reconnect.
--- a/src/main/java/examples/Examples.java
+++ b/src/main/java/examples/Examples.java
@@ -569,8 +569,54 @@ public class Examples {
       }
     });
   }
-
+  
   public void pubsub03(Vertx vertx) {
+
+    PgSubscriber subscriber = PgSubscriber.subscriber(vertx, new PgConnectOptions()
+      .setPort(5432)
+      .setHost("the-host")
+      .setDatabase("the-db")
+      .setUser("user")
+      .setPassword("secret")
+    );
+    
+    subscriber.connect(ar -> {
+        if (ar.succeeded()) {
+          // Complex channel name - name in PostgreSQL requires a quoted ID
+          subscriber.channel("Complex.Channel.Name").handler(payload -> {
+            System.out.println("Received " + payload);
+          });
+          subscriber.channel("Complex.Channel.Name").subscribeHandler(subscribed -> {
+        	  subscriber.actualConnection().query(
+        			  "NOTIFY \"Complex.Channel.Name\", 'msg'", notified -> {
+        		  System.out.println("Notified \"Complex.Channel.Name\"");
+        	  });
+          });
+
+          // PostgreSQL simple ID's are forced lower-case 
+          subscriber.channel("simple_channel").handler(payload -> {
+              System.out.println("Received " + payload);
+          });
+          subscriber.channel("simple_channel").subscribeHandler(subscribed -> {
+        	  // The following simple channel identifier is forced to lower case
+              subscriber.actualConnection().query(
+            		"NOTIFY Simple_CHANNEL, 'msg'", notified -> {
+          		  System.out.println("Notified simple_channel");
+          	  });
+          });
+          
+          // The following channel name is longer than the current
+          // (NAMEDATALEN = 64) - 1 == 63 character limit and will be truncated
+          subscriber.channel(
+        		  "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaabbbbb"
+        		  ).handler(payload -> {
+              System.out.println("Received " + payload);
+          });          
+        }
+      });
+  }
+
+  public void pubsub04(Vertx vertx) {
 
     PgSubscriber subscriber = PgSubscriber.subscriber(vertx, new PgConnectOptions()
       .setPort(5432)
--- a/src/main/java/io/reactiverse/pgclient/impl/pubsub/PgSubscriberImpl.java
+++ b/src/main/java/io/reactiverse/pgclient/impl/pubsub/PgSubscriberImpl.java
@@ -198,7 +198,7 @@ public class PgSubscriberImpl implements PgSubscriber {
 
     ChannelList(String name) {
       this.name = name;
-	  quotedName = "\"" + name + "\"";
+	  quotedName = "\"" + name.replace("\"", "\"\"") + "\"";
     }
 
     void add(ChannelImpl sub) {
--- a/src/main/java/io/reactiverse/pgclient/impl/pubsub/PgSubscriberImpl.java
+++ b/src/main/java/io/reactiverse/pgclient/impl/pubsub/PgSubscriberImpl.java
@@ -47,6 +47,15 @@ public class PgSubscriberImpl implements PgSubscriber {
     this.options = new PgConnectOptions(options);
   }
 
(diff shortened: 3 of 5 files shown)


vietj:
can you add a test for this fix ?

brosenfi:
@vietj Tests added - just piggy-backed on the existing pub-sub tests, but with a channel name that requires being a quoted identifier.  There is one scenario for use of standard ID's that this will not support - that's where the caller expects the channel name to be supported in a case-insensitive manner, however that's already not supported through the use of a case-sensitive HashMap to map notifications back to ChannelList instances.  In order to keep things simple, I think this PgSubscriber should continue to just support channel names in a case-sensitive manner (whether they require quoting or not).

vietj:
I think we need to to a better job for the quoted name, we need to check the name validity and possibly escape " if that's allowed.

from the postgres grammar:

```
/* Column identifier --- names that can be column, table, etc names.
 */
ColId:		IDENT									{ $$ = $1; }
			| unreserved_keyword					{ $$ = pstrdup($1); }
			| col_name_keyword						{ $$ = pstrdup($1); }
		;
```

brosenfi:
Right so this update, like I was saying, isn't supporting the full domain of possible channel names.  I wasn't looking straight at the grammar just knowing that channel names are identifiers from the PG docs, and looking at https://www.postgresql.org/docs/10/static/sql-syntax-lexical.html .

The core problem here in its current state is users can pass any valid PG SQL identifier as the channel name - they can use a quoted ID, unicode escape sequences, etc - but it won't be the exact same string coming back on the PgNotification, and so it won't match the channel name key in the hash map.

What I can do is leave the interface the way it was - where the user just needs to supply a valid PG SQL identifier (they supply the quotes or escaping, etc.), and inside PgSubscriber we figure out exactly what that channel name is going to be when it comes back on a PgNotification, and then make use of that as the key on the channel HashMap - how does that sound?  At a short glance that's going to be a bigger change...

vietj:
when a channel is quoted, the quote is part of the name ?

vietj:
i.e if there is a notification sent to `"abc"` then listen works on `abc` or `"abc"` ?

that's not clear to me.

brosenfi:
This is how I ran into this bug - first off if I run the statement:
`LISTEN somedomain.test.1;`
It will fail because having a period in an unquoted identifier just isn't valid - so if I quote it:
`LISTEN "somedomain.test.1";`
That succeeds - however the channel name actually comes back as `somedomain.test.1` on notifications from Postgres not as `"somedomain.test.1"` .  With a standard unquoted identifier, Postgres also forces everything to lower case - so I could run `LISTEN The_Channel;` and then `NOTIFY THE_CHANNEL;` and the message would be received by the listener with the channel name `the_channel` .  Each of these scenarios means the string the user supplied for the channel name is not what Postgres delivers back as the channel name on the notification.  Let me roll back the changes I made to PgSubscriber then you'll see on the CI build the new tests will fail.

brosenfi:
There - as you can see with the last build where I used your existing implementation of PgSubscriber with the new tests it's a bunch of failures for the reasons I had stated (note the time-outs are just the fact that the channel name from the notification that came back was not found on the channels HashMap).

vietj:
@brosenfi I see, how about we use prepared query instead of simple query ? (it would also be good to use because of injection in the channel name).

brosenfi:
@vietj After updating the tests to use prepared statements and having them fail with a statement syntax error I realized prepared statements don't work on channel names - which are identifiers like table and column names (makes sense).

I think now that the quoting of the channel names works best for what you want - that is the user specifying the exact channel name to PgSubscriber (not necessarily what would go in a NOTIFY or LISTEN query).  Also, quoted ID's support almost every character (only NULL isn't supported - and no-one should be using that inside a channel name).  I have put back the quoted ID code into PgSubscriberImpl and I have also updated the code to properly escape double-quotes as well (shouldn't be any injection threat).

vietj:
@brosenfi ok thanks for checking that, do you have the clear grammar rules for channel names so I can have a look ?

brosenfi:
@vietj Took what I just talked about straight from this documentation - https://www.postgresql.org/docs/9.6/static/sql-syntax-lexical.html:

```
There is a second kind of identifier: the delimited identifier or quoted identifier. It is formed by enclosing an arbitrary sequence of characters in double-quotes ("). A delimited identifier is always an identifier, never a key word. So "select" could be used to refer to a column or table named "select", whereas an unquoted select would be taken as a key word and would therefore provoke a parse error when used where a table or column name is expected. The example can be written with quoted identifiers like this:

UPDATE "my_table" SET "a" = 5;

Quoted identifiers can contain any character, except the character with code zero. (To include a double quote, write two double quotes.) This allows constructing table or column names that would otherwise not be possible, such as ones containing spaces or ampersands. The length limitation still applies.
```
You can see the source of the scanner as well, although there's several states to look through to find exactly what they have described in the documentation - https://github.com/postgres/postgres/blob/master/src/backend/parser/scan.l :

```
xdstart			{dquote}
...
{xdstart}		{
					SET_YYLLOC();
					BEGIN(xd);
					startlit();
				}
...
<xd>{xdstop}	{
					char	   *ident;

					BEGIN(INITIAL);
					if (yyextra->literallen == 0)
						yyerror("zero-length delimited identifier");
					ident = litbufdup(yyscanner);
					if (yyextra->literallen >= NAMEDATALEN)
						truncate_identifier(ident, yyextra->literallen, true);
					yylval->str = ident;
					return IDENT;
				}

```

vietj:
thanks that's useful, I'll have a look tomorrow, I think I need to spend some time understanding the actual issue

vietj:
@brosenfi when a channel (that does not require quotation) was subscribed with a quotation, does the notifier requires to use quotation ?

i.e can the subscription `LISTEN "the_channel"` be notified with `NOTIFY the_channel` ?

brosenfi:
@vietj Yes - the subscribers will be notified in that case.  This goes along with that other issue I said a while back in the conversation too, for non-quoted identifiers Postgres forces them to be lower case, so they would be notified as well with `NOTIFY THE_channEL` .  When you quote an identifier, Postgres uses the exact string within the quotes - so `LISTEN "The_Channel"` would not be notified by `NOTIFY the_channel` - and thanks to the forced lower case of standard identifiers it would also not be notified by `NOTIFY The_Channel`.

The confusion here I think is really over what is the precise meaning of the `String channelName` argument handed to the PgSubscriber instance - is it the exact channel name Postgres will hold and send notifications on, or is it a valid PostgeSQL identifier that could appear in the LISTEN / NOTIFY statements?  If it's the exact channel name as I thought you were describing recently then I think this quoted identifier solution inside PgSubscriberImpl works perfectly - the user just has to understand it's the exact name they supply.  If instead it's a valid Postgres identifier that should be supplied, then the code would need to be changed to parse that identifier to extract the exact resulting channel name to use as the channel map key.

brosenfi:
@vietj I added a test in PubSubTests - testSubscribeNotifyWithUnquotedId - that shows the exact situation you just asked about and I described (works as I expected).

vietj:
@brosenfi right, I was about to ask for it

vietj:
@brosenfi I believe then we need to check that the channel name does not contain the `"` char and either refuse it or escape it

brosenfi:
@vietj That code is actually already in there - it escapes double quotes by doubling them up.  I will add a test that shows that code works.

brosenfi:
@vietj Test with a channel name that contains quotes added - works as expected.

vietj:
@brosenfi great, so that should protect against injection right ?

brosenfi:
@vietj Yes, it protects against injection, and the channel name ends up as the exact string the user submitted, up to a length limit.  Actually, I was just thinking it's a good idea to have that character length limit in as well (accepts identifiers up to 63 characters long - anything over that it truncates).

brosenfi:
@vietj I will put that length limit check in now.

brosenfi:
@vietj The length limit check and corresponding test are in and both working as expected.  On the test we even get a nice warning notice for the NOTIFY statement where the channel name exceeds the length limit:

```
SimpleQueryCommand should handle message NoticeResponse{severity='NOTICE', code='42622',
message='identifier "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaabbbbb"
will be truncated to "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"',
detail='null', hint='null', position='null', internalPosition='null', internalQuery='null', where='null',
file='scansup.c', line='205', routine='truncate_identifier', schema='null', table='null', column='null',
dataType='null', constraint='null'}
```

vietj:
great

I think we should handle then the notice response and support it, probably with a noticeHandler that would by default log it to the console.

vietj:
see https://github.com/reactiverse/reactive-pg-client/issues/147

vietj:
@brosenfi I think now the last missing part is to document the channel name in the documentation and tell in the apidoc that the value will be quoted and what the user should expect to use as identifier in his application with examples

brosenfi:
@vietj I added some documentation - both on the javadoc and the api documentation - take a look and see if it's enough.

vietj:
thanks for the contribution @brosenfi

brosenfi:
Sorry I had to travel for work - didn't have my github creds - sure, no problem @vietj .  We'll actually start making use of this change immediately in our Vert.x services - thanks for merging.

PR #161, opened 2018-08-27
Treat unknown datatypes as text instead of binary

davidzzzzz:
The current implementation treats unknown as binary, which deviates from other postgres drivers. This issue manifests in the current support for reading enum and enum array values.

This PR:
- Removes the existing enum implementation (the oid was for the text fixture enum, not for enums in general)
- Changes `DataType.UNKNOWN` to be a String instead of a binary type
- Updates encoders and decoders to handle both Strings and array values
- Adds a few tests around array parsing (empties, 1+, etc)

For a point of reference, both the [postres driver](https://github.com/pgjdbc/pgjdbc/blob/9534e9ca0e1840445ad5f4eee75bc1e2ac102dde/pgjdbc/src/main/java/org/postgresql/jdbc/PgResultSet.java#L220) and[ Postgres Async](https://github.com/mauricio/postgresql-async/blob/5d0a7e017e90f137a9cb0b0e20947c6bc8817b0a/postgresql-async/src/main/scala/com/github/mauricio/async/db/postgresql/column/PostgreSQLColumnDecoderRegistry.scala#L121) default to this behavior.

### DIFF
--- a/src/main/java/io/reactiverse/pgclient/impl/codec/DataTypeCodec.java
+++ b/src/main/java/io/reactiverse/pgclient/impl/codec/DataTypeCodec.java
@@ -36,9 +36,11 @@ import java.time.*;
 import java.time.temporal.ChronoField;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.UUID;
 import java.util.function.IntFunction;
+import java.util.stream.Collectors;
 
 import static java.util.concurrent.TimeUnit.*;
 
@@ -102,6 +104,10 @@ public class DataTypeCodec {
         break;
       case NUMERIC_ARRAY:
         textEncodeNUMERIC_ARRAY((Numeric[]) value, buff);
+        break;
+      case UNKNOWN:
+        //default to treating unknown as a string
+        buff.writeCharSequence(String.valueOf(value), StandardCharsets.UTF_8);
         break;
       default:
         System.out.println("Data type " + id + " does not support text encoding");
@@ -244,12 +250,6 @@ public class DataTypeCodec {
       case INTERVAL_ARRAY:
         binaryEncodeArray((Interval[]) value, DataType.INTERVAL, buff);
         break;
-      case ENUM:
-        binaryEncodeTEXT((String) value, buff);
-        break;
-      case ENUM_ARRAY:
-        binaryEncodeArray((String[]) value, DataType.ENUM, buff);
-        break;
       default:
         System.out.println("Data type " + id + " does not support binary encoding");
         defaultEncodeBinary(value, buff);
@@ -347,10 +347,6 @@ public class DataTypeCodec {
         return binaryDecodeINTERVAL(len, buff);
       case INTERVAL_ARRAY:
         return binaryDecodeArray(INTERVAL_ARRAY_FACTORY, DataType.INTERVAL, len, buff);
-      case ENUM:
-        return binaryDecodeTEXT(len, buff);
-      case ENUM_ARRAY:
-        return binaryDecodeArray(STRING_ARRAY_FACTORY, DataType.ENUM, len, buff);
       default:
         System.out.println("Data type " + id + " does not support binary decoding");
         return defaultDecodeBinary(len, buff);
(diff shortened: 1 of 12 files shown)


davidzzzzz:
The reason for the removal of enum is because any custom types, such as enums, will not have deterministic [oids](https://www.postgresql.org/docs/9.6/static/datatype-oid.html). An oid for a custom type is going to be an auto-incremented value on the `pg_type` table. The built-in types do have deterministic oids however, which is how `varchar` is always going to be `1043`. 

Unfortunately, the wire protocol doesn't have enough information to determine whether we're an enum, which is why we fall back to text, as do most other implementations (examples linked above). Considering pg already has support for blob binary data, it seems like a safe assumption. 

The current enum test cases were passing because the `mood` enum was always being created in the same order when the database was seeded, always giving the oid of `16385`. This PR adds another custom type before `mood`  which causes the tests to break.

I'll add back the enum types to this PR, and restructure it a bit so it doesn't break the existing testing since I reordered the enum type creation.

vietj:
ok that makes sense and hence when we parse a text value we need to check whether it's an array or not

src/main/java/io/reactiverse/pgclient/impl/codec/DataTypeCodec.java
```diff
@@ -472,16 +463,29 @@ public static Object prepare(DataType type, Object value) {
...
         return javaType.isInstance(value) ? javaType.cast(value) : REFUSED_SENTINEL;
     }
   }
 
   private static Object defaultDecodeText(int len, ByteBuf buff) {
-    // Default to null
-    buff.skipBytes(len);
-    return null;
+    // decode unknown text values as text or as an array if it begins with `{`
+    if (buff.getByte(buff.readerIndex()) == '{') {
```
vietj: can this fail because of an empty text value ?

vietj:
can non reserved oids be assigned to something else than an enum ?

src/main/java/io/reactiverse/pgclient/impl/codec/DataTypeCodec.java
davidzzzzz: Hmm, not entirely sure but I'll err on the side of caution here and add a check

vietj: there should be a check that the readable bytes size is > 0

davidzzzzz: 👍

davidzzzzz:
Yes, non-reserved oids that are classified as `unknown` can be resolved as something else, but the data coming back will be text. I'll add tests explicitly for custom types that are not enums for both single and array cases

vietj:
@davidzzzzz ok

davidzzzzz:
@vietj I've addressed your comments and added more tests around custom types. Would you still like the removal of the enum constant values to be a seperate pr?

src/main/java/io/reactiverse/pgclient/impl/codec/DataTypeCodec.java
```diff
@@ -472,16 +463,30 @@ public static Object prepare(DataType type, Object value) {
...
         Class<?> javaType = type.type;
         return javaType.isInstance(value) ? javaType.cast(value) : REFUSED_SENTINEL;
     }
   }
 
   private static Object defaultDecodeText(int len, ByteBuf buff) {
-    // Default to null
-    buff.skipBytes(len);
-    return null;
+    // decode unknown text values as text or as an array if it begins with `{`
```
vietj: rather

```
    int idx = buff.readerIndex();
    if (idx < buff.writerIndex() && buff.getByte(idx) == '{') {
       ...
    }
```

this might be better since `readerIndex` is accessed once (`readableBytes()` returns `writerIndex - readerIndex`).

davidzzzzz:
I've definetly learned more about Postgres from this PR than I cared to know :) And yeah, I'm pretty amazed (or horrified) that I can send plain text for a custom type and Postgres will happily parse it and insert it. At least Postgres does parse/validate the values before persisting data.

As far as I've been able to determine (from looking at specs and other implementations), `705` really does mean `Unknown` to Postgres. Perhaps could add a `CustomType` enum value, but it would be handled exactly the same was as `Unknown`, and in either case the original oid is lost.

All of this behavior seems like the curse of backwards compatibility. Postgres has changed their wire protocol in years, and as they add more features, they don't update the protocol. Which leaves situations like this one, where a text fallback for newer features works.

sumitsrv:
@vietj By when can this pull request be expected to merge and be released? I was looking for exactly the same use case, eventually added similar changes locally and use it for the time being. Then saw the PR and it seems to be of great help!

vietj:
we need this to be documented, once it's done it can be released soon after.

vietj:
@davidzzzzz can you document this improvement ?

davidzzzzz:
@vietj totally! Things at the day job got a bit busy :) I'll get too it soon

davidzzzzz:
@vietj added docs, sorry about the delay. lmk if the content's ok

vietj:
thanks for the great contribution, I'll make a release today

PR #177, opened 2018-09-17
[173] Move to Vertx logger for warnings instead of Sysout

sumitsrv:


vietj:
Don't use interpolation because it is only supported by SLF4J, use concatenation instead.

sumitsrv:
Updated to concatenation.

vietj:
looks good thanks

Issue #179, opened 2018-09-20
Type handling not lenient enough


marcust:
So I was trying to integrate jOOQ with reactive-pg-client, because I'm playing around with possible project setups. 

A problem that I encountered is that jOOQ strongly believes that the `OFFSET` and `LIMIT` parameters are `Integer`, whereas this client sees an `INT8` and expects a `Long`.

I did some magic now (https://github.com/marcust/jzenith/blob/master/jzenith-postgresql/src/main/java/org/jzenith/postgresql/PostgresqlClient.java#L96-L135) in order to avoid that, but I think if there is a type that is easily widened like an `Integer` to a `Long` that should be handled internally in the pg-client. 

I would have created a PR but I'm not sure what the best approach is, because the whole type mapping is rather 1 on 1 and I don't know what a good approach would be to make that more dynamic.

vietj:
@marcust can you check with the latest snapshot ?

marcust:
@vietj works with my test suite, so seems to be fine.

vietj:
do you have integration with jooq ? perhaps you could contribute this integration in this project ?

marcust:
No, not really an integration. I don't think that jOOQ does real async yet, from what I've seen they postpone it from version to version. I have some glue code between jOOQ, the JDBC Postgres driver and the reactive client that translates the queries to pure Postgres SQL and then just executes them (https://github.com/marcust/jzenith/blob/master/jzenith-postgresql/src/main/java/org/jzenith/postgresql/PostgresqlClient.java)

PR #183, opened 2018-09-21
A simple test for numeric typing

marcust:
That gives the error message `Values [5] cannot be coerced to [Long]` which is
at least missleading.

Demonstrating #179


vietj:
thanks, will have a look asap, I suspect it's rather an improvement than a bug fix where we need to do type coercion.

PR #204, opened 2018-11-12
Added unit-test to verify querying of -ve integer

arrowwood49:
Added unit-test to verify querying of column with -ve integer value.

See issue https://github.com/reactiverse/reactive-pg-client/issues/203


vietj:
thanks for reporting, this will be fixed

PR #205, opened 2018-11-14
Copy From Support

davidzzzzz:
This PR adds support for bulk insertion of data into Postgres either over binary, text or csv.

The `copyFrom` API accepts a `ReadStream` of `CopyTuple` or `Buffer` data, with various options to control basic copy from options, as well as the target table and columns.


vietj:
I will have a look as son as possible

vietj:
it looks great, however I believe that instead of accepting a `ReadStream` we should expose the `WriteStream` instead, this will give more control over operations to the user.

davidzzzzz:
@vietj that was my initial design actually, can easily go back to it. And apologies for the delayed reply, I was on holiday

vietj:
yes I would prefer that, at some point we might add a `pipe` method in ReadStream that takes a `WriteStream` as argument to have the pump done automatically

vietj:
n/p for holidays, I hope you had good times :-)

davidzzzzz:
hi @vietj, made your suggestions and moved to a `WriteSteam` Please provide feedback when you get a chance, and I'll add docs once everything looks ok. Thanks!

vietj:
sure I'm not forgetting about it :-)

vietj:
I'm reviewing the PR and it seems that there is code that is not really necessary in the API : CopyData/CopyTuple . Can you elaborate why those are needed to be in the API ?

davidzzzzz:
Oh course! Postgres supports two modes of copying data into a table: text and binary. For text, it’s very straight-forward: csv or plain text, and Postgres will marshall the text data into the relevant type.

For binary copying, things get a bit more complicated. Postgres needs to know specifically which type it’s dealing with. We can infer from the Java type what the intent may be, but there’s no real way to tell the different between JSON and JSONB or an int2 and an int4 without some kind of hint. And Postgres will fail the operation if the wrong type is provided.

So that’s where `CopyTuple` and `CopyData` come in: they give the end-user a way to send actual Java types with the correct Postgres type as part of the binary copy. The `CopyTuple` provides overloads that explicitly set the correct Postgres type if the setter is used, while also allowing for the explit add of a `CopyData` object directly.

Does that provide enough context?

vietj:
I see.

As the guessed data type is only used in the implementation (in CopyTupleWriteStreamImpl#writeCopyData), I think actually these new interfaces a not necessary for the user and we should remove them

davidzzzzz:
But the guessing is going to be wrong often.

It's possible to remove the `CopyData` class itself, but the `CopyTuple` needs to exist (or something similar) in order to make binary copy useful. Without it, the binary copy support most likely will not be useful to many people.

vietj:
I mean that data can be translated on the fly without the existence of this as a class itself :-)

davidzzzzz:
took another pass and removed CopyData, lmk what you think

vietj:
@davidzzzzz thanks will have a look soon :-)

vietj:
it looks like a great contribution and shaping up nicely :-)

I'm still trying to understand the usefulness of having a specific `CopyTuple` extension, I can see it declares new data types like varchar, why is it specifically needed and shouldn't be it in the `Tuple` interface after all ?

vietj:
@davidzzzzz ping :-)

vietj:
I'll do another review after, we are making good progress :-)

PR #210, opened 2018-11-28
Reorganize data type codec tests

BillyYccc:
Motivation:
Currently all data type codec tests are organized in two single files, tests may throw code too large exception as the number of supported data types grows.

This PR is to break down the monolith and reorganize the data type codec tests, following new data type tests will benefit from this.


vietj:
before doing such kind of refactoring, I think we should first rewrite these tests so that they don't need a table to exist in database, some tests already are using this using postgres casting.

BillyYccc:
If I understand correctly, take Numeric type for an example, do you mean rewriting sql like this?
`SELECT 123 :: INT4 "tableNotExist"` in simple query,
`SELECT c FROM (VALUES ($1 :: NUMERIC)) as t (c)` with values() constant table in extended query.

but how to achieve this in an easy way in encoding tests (i.e UPDATE command) ?

vietj:
that's a good question...

BillyYccc:
Maybe we could do refactoring the decoding tests first and set aside the encoding tests temporarily @vietj ?

vietj:
yes @BillyYccc that seems to be the most appropriate for now

BillyYccc:
I have rewritten binary decoding tests and text decoding tests for all data types(except for Enum & Custom) without using an existed table now.

vietj:
looks great, just a couple renaming changes for the test packages:

- use `io.reactiverse.pgclient.data` instead
- don't use simple/extended sub package instead use Simple / Extended in the test name so all are in the same package without children pkg

BillyYccc:
> looks great, just a couple renaming changes for the test packages:
> 
> * use `io.reactiverse.pgclient.data` instead
> * don't use simple/extended sub package instead use Simple / Extended in the test name so all are in the same package without children pkg

updated with https://github.com/reactiverse/reactive-pg-client/pull/210/commits/c45f801b24045d3a526377f566ea91e160eea1ec.

BillyYccc:
Can this be merged into master now @vietj?

EmadAlblueshi:
@BillyYccc Thanks!

PR #215, opened 2018-12-05
Better SSL failure handling when connecting to non SSL Postgres server - fixes #206

### LINKED ISSUE #206: SSL upgradefuture failure is not properly handled
#### Version
0.10.7
#### How to reproduce:
1. Set up a postgres server without SSL on.
2. Connect with the server using client with ssl option on.
3. AsyncResult failure is never handled.

#### Simple Code:
```
import io.reactiverse.pgclient.PgClient;
import io.reactiverse.pgclient.PgConnectOptions;
import io.reactiverse.pgclient.PgConnection;
import io.vertx.core.Vertx;

public class Reproducer {
  public static void main(String[] args) {
    Vertx vertx = Vertx.vertx();

    PgConnectOptions options = new PgConnectOptions()
      .setSsl(true)
      .setHost("localhost")
      .setPort(5432)
      .setUser("postgres")
      .setPassword("password")
      .setDatabase("postgres");

    PgClient.connect(vertx, options,
      ar -> {
        if (ar.succeeded()) {
          System.out.println("success");
          PgConnection pgConnection = ar.result();
          System.out.println("is SSL?" + pgConnection.isSSL());
        } else {
          //FIXME this is never called
          System.out.println("fail");
        }
      });

    // blocking the main method to avoid exit
    try {
      Thread.sleep(5000);
      e.printStackTrace();
    }
  }
}
```

BillyYccc:
This PR is for #206 but with some addtional changes:
1. Add support for unit testing with non SSL embedded Postgres server and configurable for this.
2. Make all tests by default use non SSL Postgres server if necessary.

3.Expose connection failure to users directly when connecting to a non-SSL Postgres server with SSL option on.

Failure handling is more straightforward when connecting to Non SSL Postgres server like this:
```
    PgConnectOptions options = new PgConnectOptions()
      .setSsl(true)
      .setTrustAll(true)
      .setHost("localhost")
      .setPort(5432)
      .setUser("postgres")
      .setPassword("password")
      .setDatabase("postgres");

    PgClient.connect(vertx, options,
      ar -> {
        if (ar.succeeded()) {
          // not called here
        } else {
          // Expected Behavior
          // directly report to users "Postgres Server does not handle SSL connection"
          System.out.println(ar.cause().getMessage());
        }
      });
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -325,10 +325,11 @@ You can run tests with an external database:
 You need to add some properties for testing:
 
 ```
-> mvn test -Dconnection.uri=postgres://$username:$password@$host:$port/$database -Dunix.socket.directory=$path
+> mvn test -Dconnection.uri=postgres://$username:$password@$host:$port/$database -Dtls.connection.uri=postgres://$username:$password@$host:$port/$database -Dunix.socket.directory=$path
 ```
 
 - connection.uri(mandatory): configure the client to connect the specified database
+- tls.connection.uri(mandatory): configure the client to run `TLSTest` with the specified Postgres with SSL enabled
 - unix.socket.directory(optional): the single unix socket directory(multiple socket directories are not supported) to test Unix domain socket with a specified database, domain socket tests will be skipped if this property is not specified
 (Note: Make sure you can access the unix domain socket with this directory under your host machine)
 - unix.socket.port(optional): unix socket file is named `.s.PGSQL.nnnn` and `nnnn` is the server's port number,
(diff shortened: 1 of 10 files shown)


vietj:
looks good, however I think that we cannot build both SSL and domain socket config at the same time, so the combo ssl=true + domain socket = true should be rejected.

vietj:
have you checked how this works with https://github.com/reactiverse/reactive-pg-client/tree/master/docker/postgres ?

vietj:
because they use SSL by default, so I believe that we cannot run the SSL upgrade failure test with the current docker image (like domain sockets are not available either).

vietj:
yes that's something possible on two different ports too.

BillyYccc:
Hi, I have managed to make this work with docker-compose now. I'm also getting the unix domain socket tests working with docker now, but I only test with my own Linux machines(OS X not tested), can you have a look @vietj ?

vietj:
did you update the README that I can follow ?

BillyYccc:
Yes, please check https://github.com/reactiverse/reactive-pg-client/blob/d50e382d2df1dfea8fa7ffaf3135b969470b7ca2/README.adoc and https://github.com/reactiverse/reactive-pg-client/blob/d50e382d2df1dfea8fa7ffaf3135b969470b7ca2/docker/postgres/README.md

vietj:
great contribution!

BillyYccc:
Thanks!😃

PR #217, opened 2018-12-13
Add getDetail to PgException

counter567:


counter567:
And you also need to review the semaphoreci build so it does not try to upload the artifact :D

counter567:
Implemented your suggested changes.

PR #224, opened 2018-12-20
Add support for sslmode - fixes #207

### LINKED ISSUE #207: SSL mode support
https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLMODE
https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-PROTECTION

BillyYccc:
This PR is to fix #207.

Changes:
- add support for **sslmode**
- **ssl** parameter is now deprecated and **sslmode** should be used instead.
 Client is in `DISABLE` mode by default, `setSsl(true)` is equivalent to `setSslMode(REQUIRE)` and `setSsl(false)` is equivalent to `setSslMode(DISABLE)` so that we can avoid break changes.
- refactor PgConnectionFactory and SocketConnection class

### DIFF
--- a/src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
+++ b/src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
@@ -258,7 +258,7 @@ public class PgConnectOptions extends NetClientOptions {
   @Override
   public PgConnectOptions setSsl(boolean ssl) {
     if (ssl) {
-      setSslMode(SslMode.REQUIRE);
+      setSslMode(SslMode.VERIFY_CA);
     } else {
       setSslMode(SslMode.DISABLE);
     }
--- a/src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
+++ b/src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
@@ -79,6 +79,9 @@ public class PgConnectOptions extends NetClientOptions {
     if (getenv("PGPASSWORD") != null) {
       pgConnectOptions.setPassword(getenv("PGPASSWORD"));
     }
+    if (getenv("PGSSLMODE") != null) {
+      pgConnectOptions.setSslMode(SslMode.of(getenv("PGSSLMODE")));
+    }
     return pgConnectOptions;
   }
 
@@ -89,6 +92,7 @@ public class PgConnectOptions extends NetClientOptions {
   public static final String DEFAULT_PASSWORD = "pass";
   public static final boolean DEFAULT_CACHE_PREPARED_STATEMENTS = false;
   public static final int DEFAULT_PIPELINING_LIMIT = 256;
+  public static final SslMode DEFAULT_SSLMODE = SslMode.DISABLE;
 
   private String host;
   private int port;
@@ -97,6 +101,7 @@ public class PgConnectOptions extends NetClientOptions {
   private String password;
   private boolean cachePreparedStatements;
   private int pipeliningLimit;
+  private SslMode sslMode;
 
   public PgConnectOptions() {
     super();
@@ -118,6 +123,7 @@ public class PgConnectOptions extends NetClientOptions {
     password = other.password;
     pipeliningLimit = other.pipeliningLimit;
     cachePreparedStatements = other.cachePreparedStatements;
+    sslMode = other.sslMode;
   }
 
   public String getHost() {
@@ -186,6 +192,15 @@ public class PgConnectOptions extends NetClientOptions {
     return this;
   }
 
+  public SslMode getSslMode() {
+    return sslMode;
+  }
+
+  public PgConnectOptions setSslMode(SslMode sslmode) {
+    this.sslMode = sslmode;
+    return this;
+  }
+
   @Override
   public PgConnectOptions setSendBufferSize(int sendBufferSize) {
     return (PgConnectOptions)super.setSendBufferSize(sendBufferSize);
@@ -231,8 +246,15 @@ public class PgConnectOptions extends NetClientOptions {
     return (PgConnectOptions)super.setIdleTimeout(idleTimeout);
   }
 
+  @Deprecated
   @Override
   public PgConnectOptions setSsl(boolean ssl) {
+    // keep consistence with Vert.x NetClient behavior to avoid breaking changes, try setSslMode() instead.
+    if (ssl) {
+      setSslMode(SslMode.REQUIRE);
+    } else {
+      setSslMode(SslMode.DISABLE);
+    }
     return (PgConnectOptions)super.setSsl(ssl);
   }
 
@@ -397,6 +419,7 @@ public class PgConnectOptions extends NetClientOptions {
     password = DEFAULT_PASSWORD;
     cachePreparedStatements = DEFAULT_CACHE_PREPARED_STATEMENTS;
     pipeliningLimit = DEFAULT_PIPELINING_LIMIT;
+    sslMode = DEFAULT_SSLMODE;
   }
 
   @Override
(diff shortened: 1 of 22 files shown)


src/main/java/io/reactiverse/pgclient/PgPoolOptions.java
```diff
@@ -200,6 +205,7 @@ public PgPoolOptions setIdleTimeout(int idleTimeout) {
     return (PgPoolOptions) super.setIdleTimeout(idleTimeout);
   }
 
+  @Deprecated
```
vietj: why this needs to be deprecated ?

vietj: I believe we can let it as is and instead of having a boolean ssl, have it set the SSL mode to REQUIRE/DISABLE instead

BillyYccc: Updated with https://github.com/reactiverse/reactive-pg-client/pull/224/commits/905ee6008c187f760ce56dc0951d9c7da8e1cae2

src/main/java/io/reactiverse/pgclient/impl/PgConnectionFactory.java
```diff
@@ -86,7 +90,88 @@ public void close() {
...
+        initiateConnection(true, ar -> {
+          if (ar.succeeded()) {
+            handler.handle(Future.succeededFuture(ar.result()));
+          } else {
+            initiateConnection(false, handler);
+          }
+        });
+        break;
+      case VERIFY_FULL:
+        if (hostnameVerificationAlgorithm == null || hostnameVerificationAlgorithm.isEmpty()) {
```
vietj: can't we verify this before connecting ?

src/test/java/io/reactiverse/pgclient/UnixDomainSocketTest.java
```diff
@@ -95,4 +96,14 @@ public void connectWithVertxInstance(TestContext context) {
       vertx.close();
     }
   }
+
+  @Test
+  public void testIgnoreSslMode(TestContext context) {
```
vietj: why is SSL mode is ignored with domain sockets ?

BillyYccc: See the documentation in https://www.postgresql.org/docs/current/libpq-connect.html.
> sslmode is ignored for Unix domain socket communication.

I'm not sure whether Postgres supports SSL over domain socket, maybe we can make domain socket behave the same like TCP connection and let the server handle this?

src/main/java/io/reactiverse/pgclient/impl/PgConnectionFactory.java
BillyYccc: The purpose here is to force the check to avoid unneccesary connection if SSL options(via NetClientOptions) are not configured correctly and break the restrictions of sslmode.

src/test/java/io/reactiverse/pgclient/UnixDomainSocketTest.java
vietj: I was actually curious to understand that. I think it's fine this way then as it's not supported at all.

src/main/java/io/reactiverse/pgclient/impl/codec/decoder/InitiateSslHandler.java
```diff
@@ -81,7 +79,9 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
       DecoderException err = (DecoderException) cause;
       cause = err.getCause();
     }
-    upgradeFuture.fail(cause);
+    if (!upgradeFuture.isComplete()) {
```
vietj: rather use `upgradeFuture.tryFail(cause)`

vietj:
I can see some parts of the `SocketConnection` have been moved to the `PgConnectionFactory` for SSL. Can you put them back in `SocketConnection` as I think it's where it belongs to.

src/main/java/io/reactiverse/pgclient/impl/codec/decoder/InitiateSslHandler.java
BillyYccc: Updated in https://github.com/reactiverse/reactive-pg-client/pull/224/commits/d874dcb80c57141e5cdb7321dee348257753f29a.

BillyYccc:
> I can see some parts of the `SocketConnection` have been moved to the `PgConnectionFactory` for SSL. Can you put them back in `SocketConnection` as I think it's where it belongs to.

I have moved them back in https://github.com/reactiverse/reactive-pg-client/pull/224/commits/41e004c23eb549e060ad6bd4310f4bffd1fb1bbf.

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
```diff
@@ -233,7 +248,12 @@ public PgConnectOptions setIdleTimeout(int idleTimeout) {
 
   @Override
   public PgConnectOptions setSsl(boolean ssl) {
-    return (PgConnectOptions)super.setSsl(ssl);
+    if (ssl) {
+      setSslMode(SslMode.REQUIRE);
```
vietj: does REQUIRE validates the certificates ?

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
```diff
@@ -186,6 +192,15 @@ public PgConnectOptions setCachePreparedStatements(boolean cachePreparedStatemen
     return this;
   }
 
+  public SslMode getSslMode() {
+    return sslMode;
+  }
+
+  public PgConnectOptions setSslMode(SslMode sslmode) {
```
vietj: need Javadoc

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
```diff
@@ -186,6 +192,15 @@ public PgConnectOptions setCachePreparedStatements(boolean cachePreparedStatemen
     return this;
   }
 
+  public SslMode getSslMode() {
```
vietj: missingjavadoc

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
vietj: I believe it is rather equivalent to verify_ca

BillyYccc: It's stated in https://www.postgresql.org/docs/current/libpq-connect.html

> only try an SSL connection. If a root CA file is present, verify the certificate in the same way as if verify-ca was specified

BillyYccc:
Javadoc updated in https://github.com/reactiverse/reactive-pg-client/pull/224/commits/3383d9f6c5eb940ddc314e6cf51063cd555f24c9.

src/main/java/io/reactiverse/pgclient/PgConnectOptions.java
BillyYccc: Why setSsl(true) is equivalent to verify_ca?

BillyYccc:
Fine, I have make changes in https://github.com/reactiverse/reactive-pg-client/pull/224/commits/9d11e8d8471c1336c39ca1aa8c8d00774b94bc11.

src/test/java/io/reactiverse/pgclient/PgClientTestBase.java
```diff
@@ -101,7 +101,7 @@ public void testConnectInvalidUsername(TestContext ctx) {
   @Test
   public void testConnectNonSSLServer(TestContext ctx) {
     Async async = ctx.async();
-    options.setSsl(true).setTrustAll(true);
+    options.setSslMode(SslMode.REQUIRE).setTrustAll(true);
```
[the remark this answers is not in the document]
BillyYccc: I think using `REQUIRE` is fine because verifying the certificate is not related to the testing target here?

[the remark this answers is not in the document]
vietj: right

BillyYccc:
@vietj Thanks for reviewing!

PR #236, opened 2019-01-14
Request cancellation support

BillyYccc:
Add support for request cancellation - fixes https://github.com/reactiverse/reactive-pg-client/issues/37.

### DIFF
--- a/src/main/docs/index.md
+++ b/src/main/docs/index.md
@@ -473,7 +473,7 @@ The default policy is to not reconnect.
 
 ## Cancelling Request
 
-Postgres supports cancelling requests in progress. In order to cancel query in processing of a connection, each {@link io.reactiverse.pgclient.PgConnection} is assigned a processId and a secretKey after connecting to the server, the client will open a new connection to the server and send a CancelRequest message with a processId and a secretKey to tell the server on which connection the query in processing should be cancelled.
+Postgres supports cancellation of requests in progress. You can cancel inflight requests using {@link io.reactiverse.pgclient.PgConnection#cancelRequest}. Cancelling a request opens a new connection to the server and cancels the request and then close the connection.
 
 ```$lang
 {@link examples.Examples#cancelRequest(io.reactiverse.pgclient.PgConnection)}
--- a/src/main/docs/index.md
+++ b/src/main/docs/index.md
@@ -471,6 +471,18 @@ value:
 
 The default policy is to not reconnect.
 
+## Cancelling Request
+
+Postgres supports cancelling requests in progress. In order to cancel query in processing of a connection, each {@link io.reactiverse.pgclient.PgConnection} is assigned a processId and a secretKey after connecting to the server, the client will open a new connection to the server and send a CancelRequest message with a processId and a secretKey to tell the server on which connection the query in processing should be cancelled.
+
+```$lang
+{@link examples.Examples#cancelRequest(io.reactiverse.pgclient.PgConnection)}
+```
+
+> The cancellation signal might or might not have any effect — for example, if it arrives after the backend has finished processing the query, then it will have no effect. If the cancellation is effective, it results in the current command being terminated early with an error message.
+
+More information can be found in the [official documentation](https://www.postgresql.org/docs/11/protocol-flow.html#id-1.10.5.7.9).
+
 ## Using SSL/TLS
 
 To configure the client to use SSL connection, you can configure the {@link io.reactiverse.pgclient.PgConnectOptions}
--- a/src/main/java/examples/Examples.java
+++ b/src/main/java/examples/Examples.java
@@ -814,9 +814,7 @@ public class Examples {
         System.out.println("Failed to query due to " + ar.cause().getMessage());
       }
     });
-    int pid = connection.getProcessId();
-    int secretKey = connection.getSecretKey();
-    connection.cancelRequest(pid, secretKey, ar -> {
+    connection.cancelRequest(ar -> {
       if (ar.succeeded()) {
         System.out.println("Cancelling request has been sent");
       } else {
--- a/src/main/java/examples/Examples.java
+++ b/src/main/java/examples/Examples.java
@@ -803,4 +803,25 @@ public class Examples {
         }
       });
   }
+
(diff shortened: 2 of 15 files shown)


BillyYccc:
I think letting user open connection himself with our current API is not possible because the connection used for cancelling request is different from a `PgConnection` class, the connection does not send a startup message and do authentication but just sends a cancelRequest message, so it can't be that users create a `PgConnection` themself and call the cancel request method.

BillyYccc:
I wonder maybe this API is better? 
```
PgConnection connToCancelRequest = ...;
connToCancelRequest.cancelRequest(ar-> {});
```

vietj:
you mean that a request cancellation won't send a startup message ?

vietj:
ok I see now it makes sense :-)

BillyYccc:
so our last concern for this is making a choice between `connToCancel.cancelRequest()`, `conn.cancelRequest(pid, key)` and `conn.cancelRequest(connToCancel)`...

vietj:
I think you demonstrated that the last one is not possible

vietj:
it seems fine as it is and the documentation clearly says that this will open a connection

src/main/docs/index.md
```diff
@@ -471,6 +471,18 @@ value:
 
 The default policy is to not reconnect.
 
+## Cancelling Request
+
+Postgres supports cancelling requests in progress. In order to cancel query in processing of a connection, each {@link io.reactiverse.pgclient.PgConnection} is assigned a processId and a secretKey after connecting to the server, the client will open a new connection to the server and send a CancelRequest message with a processId and a secretKey to tell the server on which connection the query in processing should be cancelled.
```
vietj: Postgres supports cancellation of requests in progress. You can cancel inflight requests  , using {@link io.reactiverse.pgclient.PgConnection#cancelRequest}. 

Cancelling a request opens a new connection to the server and cancels the request and then close the connection.

src/main/java/examples/Examples.java
```diff
@@ -803,4 +803,25 @@ public void collector02Example(PgClient client) {
         }
       });
   }
+
+  public void cancelRequest(PgConnection connection) {
+    connection.query("SELECT pg_sleep(20)", ar -> {
```
vietj: 20 seconds or millis ?

src/main/java/examples/Examples.java
```diff
@@ -803,4 +803,25 @@ public void collector02Example(PgClient client) {
...
+      if (ar.succeeded()) {
+        // imagine this is a long query and is still running
+        System.out.println("Query success");
```
vietj: no println please

vietj: I think however that the test should exit with a response from this part, using an Async from vertx-unit

src/main/java/examples/Examples.java
BillyYccc: 20 seconds here, maybe it could be shorter

src/main/java/examples/Examples.java
vietj: ah sorry it is example :-)

src/main/java/examples/Examples.java
vietj: it's fine it is example :-)

src/main/java/io/reactiverse/pgclient/PgConnection.java
```diff
@@ -83,6 +83,29 @@
   @Fluent
   PgConnection notificationHandler(Handler<PgNotification> handler);
 
+  /**
+   * @return The process ID of the target backend
+   */
+  int getProcessId();
```
vietj: rather `processId()`

src/main/java/io/reactiverse/pgclient/PgConnection.java
```diff
@@ -83,6 +83,29 @@
...
+
+  /**
+   * @return The secret key for the target backend
+   */
+  int getSecretKey();
```
vietj: rather `secretKey()`

src/main/java/io/reactiverse/pgclient/PgConnection.java
```diff
@@ -83,6 +83,29 @@
...
+
+  /**
+   * Open a new connection to the server and send a CancelRequest message
```
vietj: send a request cancellation message. No need to ref the actual protocol details here.

BillyYccc:
That may be a little complexed to achieve, having a static method means we would create another connectionFactory... 
I'm leaning to change it to `PgConnection cancelRequest(Handler<AsyncResult<Void>> handler);` so users can just call `connToCancel.cancelRequest(handler)`

src/main/java/io/reactiverse/pgclient/impl/PgConnectionImpl.java
```diff
@@ -144,4 +146,32 @@ public void close() {
...
+  @Override
+  public int secretKey() {
+    return conn.getSecretKey();
+  }
+
+  @Override
+  public PgConnection cancelRequest(Handler<AsyncResult<Void>> handler) {
+    Context current = Vertx.currentContext();
+    if (current == context) {
+      factory.connect(ar -> {
```
vietj: if I understand correctly, this will initiate the the Postgres protocol before the CancelRequest message is sent, right ?

BillyYccc: No, as reworked in commit https://github.com/reactiverse/reactive-pg-client/commit/07c8b82ebc4edacbe6543ede266cda7023fc85f8, `connect` method in `PgConnectionFactory` will only initiate a raw tcp connection to server and do SSL information exchange(if enabled), `create` method will actually send a startup message(i.e init Postgres protocol), the code is clear in
https://github.com/reactiverse/reactive-pg-client/blob/07c8b82ebc4edacbe6543ede266cda7023fc85f8/src/main/java/io/reactiverse/pgclient/impl/PgConnectionFactory.java#L92-L102

vietj: right, I missed that

BillyYccc:
Thanks for the review! 👍

Issue #237, opened 2019-01-15
Can not insert data with tx.preparedQuery


foxgem:
PgTransaction.preparedQuery could not work. Here is an example with PG 9.4:

~~~
pgPool.begin(res -> {
    if (res.succeeded()) {
        PgTransaction tx = res.result();

        // table: create table test (id bigserial not null, topic varchar(10) not null);
        tx.query("insert into test (topic) values ('topic')", this::voidHandler);
        tx.preparedQuery("insert into test (topic) values ($1)", Tuple.of("topic"), this::voidHandler);

        tx.commit(ar -> {
            if (ar.succeeded()) {
                // ...
            } else {
                // ...
            }
        });
    }
})

// log error only
private void voidHandler(AsyncResult<PgRowSet> result) {
    if (result.failed()) {
        logger.error("Sql execution failed, cause: {}", result.cause());
    }
}
~~~

In the example above, the first insert stmt with "tx.query" can insert data successfully. However, for the second insert stmt with "tx.preparedQuery", you can get an exception as below:

~~~
2019-01-15 11:53:52,850 [vert.x-eventloop-thread-0] ERROR t.d.e.clock.job.AbstractJobHandler - Sql execution failed, cause: {}
io.vertx.core.VertxException: Transaction already completed
	at io.reactiverse.pgclient.impl.Transaction.checkPending(Transaction.java:90)
	at io.reactiverse.pgclient.impl.Transaction.lambda$wrap$3(Transaction.java:138)
	at io.reactiverse.pgclient.impl.Transaction$$Lambda$102/316707258.handle(Unknown Source)
	at io.reactiverse.pgclient.impl.SocketConnection$CachedPreparedStatement.handle(SocketConnection.java:138)
	at io.reactiverse.pgclient.impl.SocketConnection$CachedPreparedStatement.handle(SocketConnection.java:120)
	at io.reactiverse.pgclient.impl.SocketConnection.handleMessage(SocketConnection.java:227)
	at io.reactiverse.pgclient.impl.SocketConnection.lambda$initiateProtocol$1(SocketConnection.java:110)
	at io.reactiverse.pgclient.impl.SocketConnection$$Lambda$43/1734258255.handle(Unknown Source)
	at io.vertx.core.streams.impl.InboundBuffer.handleEvent(InboundBuffer.java:225)
	at io.vertx.core.streams.impl.InboundBuffer.write(InboundBuffer.java:123)
[...]

PR #238, opened 2019-01-25
add testcases for #237.

foxgem:
Add two test cases for #237. And I tried these two cases could pass with current code. This issue might have been fixed somewhere.


vietj:
looks good but is there a point of using chars for the test which are not in the ASCII range ?

foxgem:
Not necessarily, it is just a test data in Chinese.

vietj:
@foxgem can you write them in plain ascii ? I don't dislike chinese, but I'm more worried about text editors opening such files and saving them incorrectly.

foxgem:
@vietj Done, please check it, thanks.

PR #258, opened 2019-03-31
Log notice response on the console (#147)

### LINKED ISSUE #147: Handle notice response
currently notice responses are ignored.

we should provide the ability to handle them and at least log them, e.g:

```
client.noticeHandler(msg -> {
  // handle notice
});
```

default handler would simply log it

danielamorais:
Fix #147


danielamorais:
This is my first PR and Eclipse Contributor Agreement is already signed

vietj:
this is actually not the feature asked.

I will rework the issue to more precisely describe it.

PR #293, opened 2019-05-23
MySQL datatype improvements

### LINKED ISSUE #286: Complete support for MySQL TIME datatype
Currently we haven't implemented binary encoding for MySQL `TIME` datatype, and we need to add support for this.

In addition, MySQL `TIME` datatype is diffrent and its value ranges from '-838:59:59' to '838:59:59', java type `java.time.LocalTime` will not cover this so we need to find a better type mapping, `java.time.Duration` will be a viable option for this.

BillyYccc:
This PR brings following improvements for MySQL datatype handling.

* add support and tests for several MySQL specific data types including `BINARY`, `VARBINARY`, `BLOB`, `TEXT`, `TIME`, `TIMESTAMP`, `YEAR` - fixes #286 
* improve the binary encoding logic and now the encoding data type is inferred by the java class type of the param value


BillyYccc:
Thanks for the review!

PR #295, opened 2019-05-24
Mysql codec improvements

BillyYccc:
This PR contains several optimizations for MySQL internal codecs, most notable changes are listed below

* abstract the process of encoding a packet
* moving capabilities flags into connection
* small improvements for codec logic


vietj:
we need to avoid using a lambda with `encodePacket`, basically when you write a capturing lambda (a lambda that reads variables in the scope in which it is declared) it will instantiate an object and that's the kind of things we want to avoid.

BillyYccc:
> we need to avoid using a lambda with `encodePacket`, basically when you write a capturing lambda (a lambda that reads variables in the scope in which it is declared) it will instantiate an object and that's the kind of things we want to avoid.

Updated with https://github.com/eclipse-vertx/vertx-sql-client/pull/295/commits/97d2f0a64274edb26c0806ab06b468d1c71f6e15.

PR #298, opened 2019-05-25
Add support for MySQL 16MBytes+ size packet codecs

### LINKED ISSUE #283: MySQL 16Mb+ size packet support
Currently we can't handle MySQL packet whose payload size is larger than 16Mb.

protocol reference: https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_packets.html#sect_protocol_basic_packets_sending_mt_16mb

BillyYccc:
Fix #283.

### DIFF
--- a/src/main/java/io/vertx/mysqlclient/impl/MySQLRowImpl.java
+++ b/src/main/java/io/vertx/mysqlclient/impl/MySQLRowImpl.java
@@ -100,7 +100,8 @@ public class MySQLRowImpl extends ArrayTuple implements RowInternal {
 
   @Override
   public Buffer getBuffer(String name) {
-    throw new UnsupportedOperationException();
+    int pos = rowDesc.columnIndex(name);
+    return pos == -1 ? null : getBuffer(pos);
   }
 
   @Override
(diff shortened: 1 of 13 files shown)


vietj:
we need to have the tests use random buffers and not zero values buffers

BillyYccc:
Updated with https://github.com/eclipse-vertx/vertx-sql-client/pull/298/commits/5442d4f908e840672590ba114028b7b9ca29e56c and I actually find a small issue and get it fixed in https://github.com/eclipse-vertx/vertx-sql-client/pull/298/commits/003a00151ea5325b7295e6808c449c1048f49a5a.

vietj:
I'm not sure how we check in the tests that the data we inserted is the same than the data in the table after the insertion

BillyYccc:
> I'm not sure how we check in the tests that the data we inserted is the same than the data in the table after the insertion

I had tried to use JDBC MySQL to check this before migrating the repo in https://github.com/BillyYccc/vertx-sql-client/commit/d14ed505cf79be981a4173d1b81670e2e00f9bd0#diff-d40bbfde64de783e502741b373ae1ffa, I'm not sure how to check to this now.

BillyYccc:
I think we will fall into the dilemma if we want to check it from the tables without a verified client(such as JDBC).

vietj:
can we verify using the MD5 function ? i.e we can insert the data, then do a select that retrivies the MD5 value and we compare against the same MD5 value locally

BillyYccc:
> can we verify using the MD5 function ? i.e we can insert the data, then do a select that retrivies the MD5 value and we compare against the same MD5 value locally

But how does that differ from directly comparing between the data?

BillyYccc:
I think MySQL `repeat()` function might be a solution to verify our decoding logic, then encoding logic can also be verified, let me have a try on this.

BillyYccc:
@vietj I have updated the tests and now it should be fine with that. I also find another issue and get it fixed, MySQL documentation is really bad :( and get users confused.

vietj:
good job!

PR #305, opened 2019-05-31
Separate pool options and connect options

### LINKED ISSUE #303: Separate Connect Options and Pooling Options
Currently `PgPoolOptions` is subclass of `PgConnectOptions`, we should rename `PgPoolOptions` to `PoolOptions` and move it to `sqlclient` package so that it can be shared by different database clients(PostgreSQL, MySQL...), we also need to change some APIs to adapt to these changes.

BillyYccc:
fix #303.


BillyYccc:
There are some essential changes for the examples, let's revisit the documentation and examples later.

PR #311, opened 2019-06-03
Introduce MySQLException class and remove ErrPacket in MySQL

### LINKED ISSUE #310: Introduce MySQLException class
We need a `MySQLException` class to indicate an error occured, the exception is originated from MySQL ERR_PACKET and contains essential error elements(error code, sql state and error message) sent from MySQL server.

BillyYccc:
fix #310.


BillyYccc:
Thanks for the review!

PR #316, opened 2019-06-08
MySQL row improvements

BillyYccc:
* add support for getting value by class type in MySQL row
* MySQL datatype tests improvements
* TCK tests improvements

### DIFF
--- a/vertx-mysql-client/README.adoc
+++ b/vertx-mysql-client/README.adoc
@@ -29,6 +29,7 @@
 
 MySQL data types are mapped as below.
 
+- BOOLEAN(j.l.Byte)
 - INT1(j.l.Byte)
 - INT2(j.l.Short)
 - INT3(j.l.Integer)
@@ -51,7 +52,6 @@ MySQL data types are mapped as below.
 
 these data types are not fully supported yet:
 
-* [ ] BOOLEAN(j.l.Byte)=INT1(j.l.Byte)
 * [ ] Numeric: BIT
 * [ ] String: ENUM, SET
 * [ ] JSON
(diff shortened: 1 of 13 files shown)


vertx-mysql-client/README.adoc
```diff
@@ -29,6 +29,7 @@
 
 MySQL data types are mapped as below.
 
+- BOOLEAN(j.l.Byte)
```
vietj: why not `java.lang.Boolean` ?

BillyYccc: Because there is no `BOOLEAN` data type in MySQL, a `BOOLEAN` in MySQL is synonyms for `TINYINT(1)`, so we can only get a byte value for a boolean type, but we don't know whether it's `BOOLEAN` or `TINYINT`. It's more clear stated in the https://dev.mysql.com/doc/refman/8.0/en/numeric-type-overview.html.

BillyYccc: What we can do the best is store the `TINYINT` value as a byte value, when users directly call `Row#getValue` they get a byte value, when they call `Row#getBoolean` we have an intelligent auto mapping to boolean for them.

vietj: so we don't support boolean in Tuple for MySQL ? or is it that the natural type mapping in MySQL for TINYINT in `j.l.Byte` but we can still use boolean to read/write TINYINT ?

vietj: it seems that it is still possible according to code that we can read/write booleans but that the natural type mapping for TINYINT is j.l.Byte, is that right ?

vietj: @BillyYccc yes indeed and we should document that .

BillyYccc: yes, when we encode a `boolean`, it's auto transformed to byte 1 or 0 in `TINYINT` format.

BillyYccc: +1 for documentation, I will add this in documentation later, let's proceed with this first.

Issue #317, opened 2019-06-09
MySQL documentation


BillyYccc:
Add user guide documentation for MySQL.

PR #318, opened 2019-06-09
MySQL documentation

BillyYccc:
Initial work for MySQL documentation, related to #317.

### DIFF
--- a/vertx-mysql-client/src/main/asciidoc/index.adoc
+++ b/vertx-mysql-client/src/main/asciidoc/index.adoc
@@ -1,3 +1,342 @@
 = Reactive MySQL Client
 
-Todo.
+The Reactive MySQL Client is a client for MySQL with a straightforward API focusing on
+scalability and low overhead.
+
+*Features*
+
+* Event driven
+* Lightweight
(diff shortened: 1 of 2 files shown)


BillyYccc:
@vietj I think we may not need a dedicated section for `SqlClient`, we can just build paralled documentation for only MySQL and PostgreSQL because I find there is not too much things which can be shared between them.

vietj:
I believe we should have a common section like there is with vertx-sql-common

vietj:
then we can include this section (and repeat it) in all implementations using asciidoc include

vietj:
I believe that an example should take a pool or connection as input and then any example that does not have to use specific Pg / MySQL prefixed implementation can be shared:

```
public void example(SqlConnection conn) {
   ...
}
```

BillyYccc:
ok, shall we first get this into merged and then see what can be extracted to the common section of documentation?

vietj:
I have no strong opinion about this @BillyYccc , up to you

BillyYccc:
I have created a ticket for this, let's first merge this and then solve #319 later.

Issue #314, opened 2019-06-07
preparedQuery method will exhaust MySQL statements


BillyYccc:
A simple reproducer is 
```
  @Test
  public void testPreparedQuery(TestContext ctx) {
    MySQLConnection.connect(vertx, options, ctx.asyncAssertSuccess(conn -> {
      for (int i = 0; i < 20000; i++) {
        conn.preparedQuery("SELECT 'hello'", ctx.asyncAssertSuccess(res ->{
          ctx.assertEquals("hello", res.iterator().next().getValue(0));
        }));
      }
    }));
  }
```

error is 
`io.vertx.mysqlclient.MySQLException: Can't create more than max_prepared_stmt_count statements (current value: 16382)`

because executing `SqlConnection#preparedQuery`creates a prepared statement but does not close it after executing.

BillyYccc:
A quick fix for this would mark a executingCommand with a closeAfterExectuing flag so that we can close the one-shot preparedQuery after executing, but we would not benefit much from the one-shot prepared query because we will have to execute like prepare->execute->close and that increases the round-trips.

Another way to fix can be move the `Prepared queries caching` feature to sqlclient, we can close the preparedStatement when it's evicted from the cache, in this way we may need more features for the cache as the cache in pgclient is actually a hashmap, we can also provide configurations for this.
In addition we can make the cache pluggable but that may be another issue we should think about.

vietj:
is it possible to determine the max number of prepared queries in MYSQL ?

BillyYccc:
yes, we can execute a query like `SET GLOBAL max_prepared_stmt_count = 50000;`, but I don't think this will always work in any use cases.

vietj:
I mean retrieve the value, overall I think we should have a max size value that is ocnfigurable and when the max size is reached then close old statements

vietj:
and we make this possible on `MySQLConnection` with a get/set of max statement cache

vietj:
and we provide an initial value in options.

BillyYccc:
I wonder whether we can also provide options for SQL placeholder interpolation like JDBC does by default, in this way users can eliminate this kind of error totally.

BillyYccc:
Closing as the best practice for now is added in https://github.com/eclipse-vertx/vertx-sql-client/pull/540.

PR #325, opened 2019-06-13
Prepared statement cache retrofit

BillyYccc:
1. move prepared statement cache from `pgclient` to `sqlclient` so that we can use it in `mysqlclient`
2. introduce a `LruCache` for the cache implementation
3. add support for configuring with cache size and cache sql length limit
4. close the statement when it's evicted from the cache


BillyYccc:
Using cache will somehow alleviate producing `Can't create more than max_prepared_stmt_count ` error stated in #314 when preparing a lot of statements with the same sql string.

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlConnectOptions.java
```diff
@@ -9,11 +9,18 @@
...
+  public static final int DEFAULT_PREPARED_STATEMENT_CACHE_SQL_LIMIT = 2048;
+
   private String host;
   private int port;
   private String user;
   private String password;
   private String database;
+  private boolean cachePreparedStatements = DEFAULT_CACHE_PREPARED_STATEMENTS;
+  private int preparedStatementCacheSize = DEFAULT_PREPARED_STATEMENT_CACHE_SIZE;
+  private int preparedStatementCacheSqlLimit = DEFAULT_PREPARED_STATEMENT_CACHE_SQL_LIMIT;
```
vietj: rename preparedStatementCacheMaxSize

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/LruCache.java
```diff
@@ -0,0 +1,32 @@
+package io.vertx.sqlclient.impl;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * A reusable cache implementation based on LRU(Least Recently Used) strategy.
+ */
+public class LruCache<K, V> extends LinkedHashMap<K, V> {
```
vietj: make this package private

vietj: I think it's actually not need to have this intermediary class, just make the cache class extends `LinkedHashMap`

PR #330, opened 2019-06-18
RowResultDecoder abstraction

### LINKED ISSUE #329: RowResultDecoder abstraction
We use `RowResultDecoder` class to decode rows with `java.util.stream.Collector` support in MySQL and PostgreSQL, such generic class can be made abstract and extracted to `sqlclient` class.

BillyYccc:
fix #329.


vietj:
I don't think it is necessary to have an abstract class for this at the moment, I prefer to keep implementation separate to give more freedom over decoding

BillyYccc:
We can still handle the inbound ByteBuf at will with the abstract method `decodeRow`, the main idea of this is to abstract the collector/rowset handling logic, maybe we can revisit this to see if it's really necessary after we complete the MSSQL implementation.

PR #370, opened 2019-07-23
support for ts types

### LINKED ISSUE #371: Support for text search types
Add support for text search in postgresql (tsquery and tsvector). 

This types can be handled using java `String` as they are usually not really used on the client side. 
Most of the time we only need to pass a string to the functions to_tsvector or to_tsquery so we need to do this type of query : 

```java 
client.preparedQuery("SELECT to_tsvector( $1 ) @@ to_tsquery( $2 )", Tuple.of("fat cats ate fat rats", "fat & rat"),  ar -> {
      if (ar.succeeded()) {
        RowSet rows = ar.result();
        for (Row row : rows) {
          System.out.println("Match : " + row.getBoolean(0));
        }
      } else {
        System.out.println("Failure: " + ar.cause().getMessage());
      }
    });
```

larousso:
fix #371 

Hi, 

Here is a PR to support text search types. 

I have handled tsquery and tsvector as text (not with a dedicated type) because usually this types are calculated from text using functions and I don't think a dedicated type would be useful. 

I have written some tests and I'm not sure I have the same coverage of case that the tests for other types so if you see missing case tell me! 

In the DataType enum I have set `supportsBinary` to `false` because it doesn't work when the value is `true` (word are truncated and I don't know why). I don't find informations on why types are eligible to binary/text encoding.

### DIFF
--- a/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
+++ b/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
@@ -611,12 +611,12 @@ class DataTypeCodec {
       case JSON:
       case JSONB:
         if (value == null ||
-          value == Tuple.JSON_NULL ||
-          value instanceof String ||
-          value instanceof Boolean ||
-          value instanceof Number ||
-          value instanceof JsonObject ||
-          value instanceof JsonArray) {
+            value == Tuple.JSON_NULL ||
+            value instanceof String ||
+            value instanceof Boolean ||
+            value instanceof Number ||
+            value instanceof JsonObject ||
+            value instanceof JsonArray) {
           return value;
         } else {
           return REFUSED_SENTINEL;
@@ -1305,7 +1305,7 @@ class DataTypeCodec {
     binaryEncodeJSON(value, buff);
   }
 
-  private static Object binaryDecodeTsVector(int index, int len, ByteBuf buff) {
+  private static String binaryDecodeTsVector(int index, int len, ByteBuf buff) {
     return buff.getCharSequence(index, len, StandardCharsets.UTF_8).toString();
   }
 
@@ -1313,7 +1313,7 @@ class DataTypeCodec {
     buff.writeCharSequence(String.valueOf(value), StandardCharsets.UTF_8);
   }
 
-  private static Object binaryDecodeTsQuery(int index, int len, ByteBuf buff) {
+  private static String binaryDecodeTsQuery(int index, int len, ByteBuf buff) {
     return buff.getCharSequence(index, len, StandardCharsets.UTF_8).toString();
   }
 
@@ -1321,11 +1321,11 @@ class DataTypeCodec {
     buff.writeCharSequence(String.valueOf(value), StandardCharsets.UTF_8);
   }
 
-  private static Object textDecodeTsVector(int index, int len, ByteBuf buff) {
+  private static String textDecodeTsVector(int index, int len, ByteBuf buff) {
     return buff.getCharSequence(index, len, StandardCharsets.UTF_8).toString();
   }
 
-  private static Object textDecodeTsQuery(int index, int len, ByteBuf buff) {
+  private static String textDecodeTsQuery(int index, int len, ByteBuf buff) {
     return buff.getCharSequence(index, len, StandardCharsets.UTF_8).toString();
   }
   /**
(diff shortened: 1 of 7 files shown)


vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
```diff
@@ -572,22 +592,31 @@ public static Object decodeText(DataType id, int index, int len, ByteBuf buff) {
...
       case JSON:
       case JSONB:
         if (value == null ||
-            value == Tuple.JSON_NULL ||
-            value instanceof String ||
-            value instanceof Boolean ||
-            value instanceof Number ||
-            value instanceof JsonObject ||
-            value instanceof JsonArray) {
+          value == Tuple.JSON_NULL ||
```
vietj: can you avoid reformat ?

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
```diff
@@ -1276,6 +1305,29 @@ private static void binaryEncodeJSONB(Object value, ByteBuf buff) {
     binaryEncodeJSON(value, buff);
   }
 
+  private static Object binaryDecodeTsVector(int index, int len, ByteBuf buff) {
```
vietj: use `String` as return type

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
```diff
@@ -1276,6 +1305,29 @@ private static void binaryEncodeJSONB(Object value, ByteBuf buff) {
...
+    return buff.getCharSequence(index, len, StandardCharsets.UTF_8).toString();
+  }
+
+  private static void binaryEncodeTsVector(String value, ByteBuf buff) {
+    buff.writeCharSequence(String.valueOf(value), StandardCharsets.UTF_8);
+  }
+
+  private static Object binaryDecodeTsQuery(int index, int len, ByteBuf buff) {
```
vietj: use `String` as return type

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
```diff
@@ -1276,6 +1305,29 @@ private static void binaryEncodeJSONB(Object value, ByteBuf buff) {
...
+    return buff.getCharSequence(index, len, StandardCharsets.UTF_8).toString();
+  }
+
+  private static void binaryEncodeTsQuery(String value, ByteBuf buff) {
+    buff.writeCharSequence(String.valueOf(value), StandardCharsets.UTF_8);
+  }
+
+  private static Object textDecodeTsVector(int index, int len, ByteBuf buff) {
```
vietj: use `String` as return type

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
```diff
@@ -1276,6 +1305,29 @@ private static void binaryEncodeJSONB(Object value, ByteBuf buff) {
...
+    return buff.getCharSequence(index, len, StandardCharsets.UTF_8).toString();
+  }
+
+  private static Object textDecodeTsQuery(int index, int len, ByteBuf buff) {
```
vietj: use `String` as return type

vietj:
looks great, can you open an issue that describes more precisely the new feature and also update the documentation that references the new types

larousso:
Yeah sure !

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
larousso: Done

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
larousso: Done

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
larousso: Done

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
larousso: Done

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
larousso: Done

larousso:
Documentation is done

BillyYccc:
Can you please update the [README](https://github.com/eclipse-vertx/vertx-sql-client/blob/3.8/vertx-pg-client/README.adoc) as well?

vietj:
great contribution, thanks!

PR #391, opened 2019-08-12
MySQL LOCAL INFILE support

### LINKED ISSUE #301: MySQL LOCAL INFILE Request support
protocol reference: https://dev.mysql.com/doc/dev/mysql-server/8.0.12/page_protocol_com_query_response_local_infile_request.html

BillyYccc:
fixes #301.


BillyYccc:
I feel support for the files size larger than default MAX_PACKET_SIZE is necessary and that's something missing now.

PR #392, opened 2019-08-13
Add support for SqlResultProperty

BillyYccc:
we still need documentation and examples.


vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLResult.java
```diff
@@ -0,0 +1,15 @@
...
+import io.vertx.codegen.annotations.VertxGen;
+import io.vertx.sqlclient.SqlClient;
+import io.vertx.sqlclient.SqlResultProperty;
+
+/**
+ *
+ */
+@VertxGen
+public interface MySQLResult {
+  SqlResultProperty<Integer> LAST_INSERTED_ID = new SqlResultProperty<Integer>() {
```
vietj: rather on `MySQLClient`, `MySQLResult` is confusing and make think that this class should be used as result specifically for MySQL and it is not

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlResultProperty.java
```diff
@@ -0,0 +1,7 @@
+package io.vertx.sqlclient;
+
+import io.vertx.codegen.annotations.VertxGen;
+
+@VertxGen
+public interface SqlResultProperty<T> {
```
vietj: I think we should name it more generically so it could eventually be reused for another similar case, I think naming it `PropertyKind` is good.

vietj: this should have a `Class<T> type()` that returns the java class

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlResultBase.java
```diff
@@ -43,6 +46,12 @@ public int size() {
     return size;
   }
 
+  @Override
+  @SuppressWarnings("unchecked")
+  public <V> V property(SqlResultProperty<V> property) {
+    return (V) sqlResultProperties.getOrDefault(property, null);
```
vietj: with the addition of `type()` on the property we can do rather:

```
return property.type().cast(sqlResultProperties.get(property));
```

BillyYccc:
I have updated according to the comments.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlResultBase.java
```diff
@@ -18,15 +18,18 @@
...
 import java.util.List;
+import java.util.Map;
 
 public abstract class SqlResultBase<T, R extends SqlResultBase<T, R>> implements SqlResult<T> {
 
   int updated;
   List<String> columnNames;
   int size;
   R next;
+  Map<PropertyKind<?>, Object> sqlResultProperties;
```
vietj: just call it `properties`

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlResultBase.java
```diff
@@ -43,6 +46,11 @@ public int size() {
     return size;
   }
 
+  @Override
+  public <V> V property(PropertyKind<V> property) {
+    return property.type().cast(sqlResultProperties.get(property));
```
vietj: sqlResultProperties might be null because it's lazy initialised

vietj:
missing a test that retrieves with a non existing property (like a fake one for the test) because for now it will throw UnsupportedOperationException.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlResultBase.java
BillyYccc: updated in https://github.com/eclipse-vertx/vertx-sql-client/pull/392/commits/7135bbeae5cfe7e8a038e23ba45fd3c0a75c2b98.

BillyYccc:
> missing a test that retrieves with a non existing property (like a fake one for the test) because for now it will throw UnsupportedOperationException.

the error handling logic is improved in https://github.com/eclipse-vertx/vertx-sql-client/pull/392/commits/fcb6079923a2d631eed6a30606e18b61237d631f.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlResultBase.java
```diff
@@ -43,6 +46,24 @@ public int size() {
...
+  public <V> V property(PropertyKind<V> property) {
+    if (property == null) {
+      throw new IllegalArgumentException("Property can not be null");
+    }
+    if (properties == null) {
+      return null;
+    } else {
+      Object value = properties.get(property);
+      Class<V> type = property.type();
+      if (type == null) {
```
vietj: no need for this,I think it's better to have an NPE

vietj: an NPE on type.cast()

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlResult.java
```diff
@@ -57,6 +57,8 @@
    */
   int size();
 
+  <V> V property(PropertyKind<V> property);
```
vietj: Missing javadoc

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlResultBase.java
BillyYccc: see https://github.com/eclipse-vertx/vertx-sql-client/pull/392/commits/dbb4d080eff0dae1f078efa384d72a1a2aadd162.

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlResult.java
BillyYccc: added in https://github.com/eclipse-vertx/vertx-sql-client/pull/392/commits/3ba7628e94292eecbf57f94a0179a6af15204215.

vietj:
thanks for the contribution

BillyYccc:
Thanks for the review!

PR #403, opened 2019-08-26
MySQL stored procedures support

### LINKED ISSUE #402: MySQL stored procedure support
protocol reference: https://dev.mysql.com/doc/dev/mysql-server/8.0.12/page_protocol_command_phase_sp.html

BillyYccc:
We have already supported the stored procedures via the queries and we do not need to introduce something like `CallableStatement`.

Add tests and documentation for stored procedures - fixes #402

### DIFF
--- a/vertx-mysql-client/src/main/asciidoc/index.adoc
+++ b/vertx-mysql-client/src/main/asciidoc/index.adoc
@@ -14,6 +14,7 @@ scalability and low overhead.
 * RxJava 1 and RxJava 2
 * Direct memory to object without unnecessary copies
 * Java 8 Date and Time
+* Stored Procedures support
 * MySQL utilities commands support
 * Compatible with MySQL 5.6 and 5.7
 * Rich collation and charset support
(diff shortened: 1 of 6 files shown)


vertx-mysql-client/src/main/asciidoc/index.adoc
```diff
@@ -252,6 +253,17 @@ create easily create a string directly from the row set:
 {@link examples.MySQLClientExamples#collector02Example(io.vertx.sqlclient.SqlClient)}
 ----
 
+== MySQL Stored Procedure
+
+This client supports for running stored procedures in queries, the result will be retrieved from the server following the https://dev.mysql.com/doc/dev/mysql-server/8.0.12/page_protocol_command_phase_sp.html[MySQL protocol] without any magic here.
```
vietj: rather "You can run stored procedures in queries. The result will ."

BillyYccc: updated in https://github.com/eclipse-vertx/vertx-sql-client/pull/403/commits/35de77b6698f39ea616f85819fd0069e46216d47.

PR #410, opened 2019-09-08
Simple pojo mapping

vietj:
This provides a few refactor and simplifications in the code base as well as the ability to provide simple row mapping for mapping query results to objects using the underlying JSON mapping facilities or a function to achieve the mapping.


vietj:
@BillyYccc can you have a look ?

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/RowSetImpl.java
```diff
@@ -16,56 +16,70 @@
...
+  static <R> Collector<Row, RowSetImpl<R>, RowSet<R>> mappingCollector(Function<JsonObject, R> f) {
+    return Collector.of(
+      RowSetImpl::new,
+      (set, row) -> {
+        JsonObject o = new JsonObject();
+        int size = row.size();
+        for (int idx = 0;idx < size;idx++) {
+          String key = row.getColumnName(idx);
+          Object value = row.getValue(idx);
+          o.put(key, value);
```
BillyYccc: I don't think this will work for non-JSON types such as `DATETIME` unless we made a transformation here.

vertx-mysql-client/src/test/java/io/vertx/mysqlclient/MySQLQueryTest.java
```diff
@@ -268,6 +274,116 @@ public void testPreparedCollector(TestContext ctx) {
...
+
+  private void testCollectorFailure(TestContext ctx, Throwable cause, Collector<Row, Object, Object> collector) {
+    MySQLConnection.connect(vertx, options, ctx.asyncAssertSuccess(conn -> {
+      conn.query("SELECT * FROM collectorTest WHERE id = 1", collector, ctx.asyncAssertFailure(result -> {
+        ctx.assertEquals(cause, result);
+        conn.close();
+      }));
+    }));
+  }
+
```
BillyYccc: Maybe we can move these new tests and the `Collector` related tests to the `vertx-sql-client` later.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/RowSetImpl.java
vietj: yes I know, actually we will update JsonObject so non json objects are stored as is in the underlying map and the `Instant` to String transformation is done when serializing to a String or a Buffer

vertx-mysql-client/src/test/java/io/vertx/mysqlclient/MySQLQueryTest.java
vietj: yes it would be good thing to do

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/RowSetImpl.java
vietj: I'll try to provide a patch for this in vertx-core

vietj: we should have some integration tests for this though

vietj: I think in practice anything will be serialised for non JSON types but it should work as long as it can be serialized and deserialized. For instance it works with Instant because Vert.x has jackson serializer / deserializer for those. It might create un-necessary copies though.

vietj: storing stuff as non string would only help for Vert.x data object mapping that would use directly the right value from JSON

vietj: so perhaps for now we should only merge all the commits but not the last commit about mapping, given that this commit is quite easy and we give more thought at it

vietj: what we could do is just provide mapping using Row and not JSON, i.e `Function<Row, T>` and allow anyone to have its transformation here without for now the `JsonObject`.

BillyYccc: > so perhaps for now we should only merge all the commits but not the last commit about mapping, given that this commit is quite easy and we give more thought at it

yes that's pragmatic.

> what we could do is just provide mapping using `Row` and not JSON, i.e `Function<Row, T>` and allow anyone to have its transformation here without for now the `JsonObject`.

Maybe we can provide methods converting `Row` to `Data Object` if we can support all Java types as proposed in https://github.com/vert-x3/wiki/wiki/RFC%3A-codegen-any-Java-type-support.

vietj: I think for now I will only push the changes without mapping and discuss more in depth mapping itself in the community because I can see other options for this.

vietj: Basically I don't think we need to use `JsonObject` and we could use instead the `Iterable<Map.Entry<String, Object>>` interface I think, so reusing data object would still work and we would not be coupled that much to JSON.

We would also have `Row` implement the same interface and then we don't need to use JsonObject for this

BillyYccc: yes that's the way we should go for now.

vietj: For now I will only push the changes without the mapping and let this PR open as a draft for this future feature

mystdeim:
Hello! Does anybody have news about this pull-request? I'm really looking forward this feature

vietj:
@mystdeim it will be in Vert.x 4 I think, we need first to have `JsonObject` extends `Map<String, Object>` in vertx-core to have this working

vietj:
@mystdeim you can make your own mapper currently using `Collector` though.

PR #415, opened 2019-09-12
MySQL transaction management fix.

### LINKED ISSUE #280: MySQL transaction support
reference: https://dev.mysql.com/doc/refman/8.0/en/sql-syntax-transactions.html

jeoffreylim:
Fixes #280


jeoffreylim:
Weird, CI compilation error. But it works now.

jeoffreylim:
I see that some of the unit tests can be unified for both MySQL & PG.  However, I am more interested of having the functionality to set transaction isolation level. Is there a plan for that?

vietj:
are your two questions related ? it is not clear to me they seem like distinct questions.

jeoffreylim:
Hi @vietj to give more clarity, while working on the MySQL transaction functionality I see that some of the unit tests can be combined for both MySQL & Postgres. They look like just repeated unit test code for both DB clients.

But I am more interested to work on setting the transaction isolation level. Is there any plans for that one? Even for PostgresSQL there is no API and functionality for it, just some commented out unit test code.

I propose you provide the transaction isolation API definition to the vertx-sql-client module, and I could help out and contribute the transaction isolation functionality for MySQL.

vietj:
I'm not much familiar with the transaction isolation setting can you point out documentation for MySQL and PostgreSQL ?

jeoffreylim:
Here are the APIs for vertx for transaction isolation:

https://vertx.io/docs/apidocs/io/vertx/ext/sql/SQLOptions.html#getTransactionIsolation--
https://vertx.io/docs/apidocs/io/vertx/ext/sql/TransactionIsolation.html

I saw some commented out unit test code in here:

https://github.com/eclipse-vertx/vertx-sql-client/blob/fa8ad2604a7ff128ed2df8208c62f10c927d5a5d/vertx-pg-client/src/test/java/io/vertx/pgclient/PgConnectionTestBase.java#L181

```
@Test
  public void testGetDefaultTx(TestContext ctx) {
    Async async = ctx.async();
    PgClientImpl client = (PgClientImpl) PgClient.create(vertx, options);
    client.connect(c -> {
      SQLConnection conn = c.result();
      conn.getTransactionIsolation(ctx.asyncAssertSuccess(result -> {
        ctx.assertEquals(TransactionIsolation.READ_COMMITTED, result);
        async.complete();
      }));
    });
  }
```

These would pretty much complete functionalities for database transactions.

Heromyth:
If autocommit is set to true (default), the ROLLBACK will not works. So, is it necessary to add this setting to testReleaseConnectionOnRollback in MySQLTransactionTest.java?

jeoffreylim:
> If autocommit is set to true (default), the ROLLBACK will not works. So, is it necessary to add this setting to testReleaseConnectionOnRollback in MySQLTransactionTest.java?

@Heromyth all the unit tests is within a transaction and is not in auto commit mode. The "connector" handler passes a Transaction object to the lambda callback, and is not a Connection.

BillyYccc:
> Hi @vietj to give more clarity, while working on the MySQL transaction functionality I see that some of the unit tests can be combined for both MySQL & Postgres. They look like just repeated unit test code for both DB clients.

I think this could be done to extract the common tests into the TCK later when we have a stable `Transaction` API.

> But I am more interested to work on setting the transaction isolation level. Is there any plans for that one? Even for PostgresSQL there is no API and functionality for it, just some commented out unit test code.

My opinion is we can add this API in the `SqlConnection` interface in another PR maybe, users can set the isolation level with scheduling a SQL query directly for now, and we should focus primarily on making MySQL client compatible with the current `Transaction` API in this PR.

jeoffreylim:
> My opinion is we can add this API in the `SqlConnection` interface in another PR maybe, users can set the isolation level with scheduling a SQL query directly for now, and we should focus primarily on making MySQL client compatible with the current `Transaction` API in this PR.
@BillyYccc I did the best I could to have something equivalent to PostgreSQL "message type ready for query" through MySQL ok packet to intercept transaction errors. This made the Transaction API compatible with MySQL, and the unit tests verifies this. 

Let me know if I could help some more. Thanks!

BillyYccc:
@jeoffreylim  I think it's fine for now, `TxStatus` is PostgreSQL protocol-specific and indeed it's used in `Transaction` to fast-fail the appending queries in a transaction and executes a rollback, an Error packet should be ok for this.

Can you please resolve the conflicting files? You can open other PR/issues to talk about adding new features such as isolation levels setting.

jeoffreylim:
@BillyYccc fixed the conflicting files. Also created a new issue for setting transaction isolation levels #432

vietj:
thanks for the contribution

PR #416, opened 2019-09-13
Fixed copy of direct buffers causes leak.

AndreasKl:
@vietj Did not found a prettier solution for copying the `ByteBuf` to a heap `Buffer` without calling low level parts e.g. `PartialPooledByteBufAllocator`.

### DIFF
--- a/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
+++ b/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
@@ -1079,7 +1079,9 @@ class DataTypeCodec {
   }
 
   private static Buffer binaryDecodeBYTEA(int index, int len, ByteBuf buff) {
-    return Buffer.buffer(buff.copy(index, len));
+    Buffer target = Buffer.buffer(len);
+    target.appendBuffer(Buffer.buffer(buff.slice(index, len)));
+    return target;
   }
 
   private static void binaryEncodeUUID(UUID uuid, ByteBuf buff) {
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/DataTypeCodec.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/DataTypeCodec.java
@@ -66,7 +66,7 @@ class DataTypeCodec {
       case VARSTRING:
       case BLOB:
       default:
-        return textDecodeBlobOrText(charset, columnDefinitionFlags, data);
+        return textDecodeBlobOrText(charset, columnDefinitionFlags, data, length);
     }
   }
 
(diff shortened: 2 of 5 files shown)


vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
```diff
@@ -1079,6 +1079,11 @@ private static void binaryEncodeBYTEA(Buffer value, ByteBuf buff) {
   }
 
   private static Buffer binaryDecodeBYTEA(int index, int len, ByteBuf buff) {
+    if (buff.isDirect()) {
```
vietj: is there a case the ByteBuf is not direct ?

vietj:
can you check if there are other places where this might happen in this codebase ?

AndreasKl:
> can you check if there are other places where this might happen in this codebase ?

I did a simple search for `.copy(` yesterday and this was the only occurrence. However I did not take the time to really analyse what happens with the `ByteBuf` passed in by `io.vertx.pgclient.impl.codec.PgDecoder#channelRead`. I can do an deeper analysis this evening.

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
AndreasKl: Afair, my debug logs showed that the `ByteBuf` is not always direct. I will check that.

vietj: interesting, can you provide stack traces of both cases ? I'm curious to understand the usage pattern

AndreasKl: I logged the result of `buff.copy` and the resulting `Buffer` could be either a direct `ByteBuf` or not, I did not really understand the pattern when which one was chosen. However the input was always a direct buffer. Sorry for the confusion. I'll remove the condition.

AndreasKl:
> > can you check if there are other places where this might happen in this codebase ?
> 
> I did a simple search for `.copy(` yesterday and this was the only occurrence. However I did not take the time to really analyse what happens with the `ByteBuf` passed in by `io.vertx.pgclient.impl.codec.PgDecoder#channelRead`. I can do an deeper analysis this evening.

Looks like the same issue is in `textDecodeBlob` and `binaryDecodeBlob`, will fix these as well.

vietj:
thanks @AndreasKl for spotting this

AndreasKl:
@vietj Fixed the three issues I found. I looked at all methods that decode things and return a `Buffer`. Tests are in place, I'm not happy with the duplication in `MySQLDataTypeTestBase` but I think it is OK for now.

vietj:
@AndreasKl  thanks for the contribution, your investigations and the time you involved in this effort, this is really important for this project

PR #419, opened 2019-09-15
Misc tests clean up

benbenw:
- Remove unused init.sql

Signed-off-by: zorglub <b.wiart@ubik-ingenierie.com>


vietj:
@BillyYccc can you review this ?

BillyYccc:
It's reasonable to do this because actually we're using the `create-postgres.sql` for Postgres testing instead of this SQL script, I think this was copied by accident when we performed the migration.

Issue #300, opened 2019-05-28
MySQL caching_sha2_password authentication method support


BillyYccc:
Add support for this so that we can work with MySQL 8, but before supporting this we need to support TLS handshake first.

protocol reference:

* https://dev.mysql.com/doc/dev/mysql-server/latest/page_caching_sha2_authentication_exchanges.html

* https://dev.mysql.com/doc/refman/8.0/en/caching-sha2-pluggable-authentication.html

PR #420, opened 2019-09-15
MySQL 8 support

BillyYccc:
This PR brings some big changes to the MySQL client, including
* full `caching_sha2_password` authentication method support (with TLS and non-TLS RSA key pair password encryption) - fixes https://github.com/eclipse-vertx/vertx-sql-client/issues/300
* rework the old implementations of authentication(initcommand and change user command)
* support `utf8mb4_0900_ai_ci` collation which is made the default for MySQL 8.0
* update documentation for MySQL/MariaDB versions support


vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLConnectOptions.java
```diff
@@ -145,6 +146,26 @@ public MySQLConnectOptions setSsl(boolean ssl) {
     return this;
   }
 
+  /**
+   * Set the server RSA public key which is mostly used for encrypting password under insecure connections when performing authentication.
+   *
+   * @param serverRsaPublicKey the value of the server RSA public key
+   * @return a reference to this, so the API can be used fluently
+   */
+  public MySQLConnectOptions setServerRsaPublicKey(String serverRsaPublicKey) {
```
vietj: any key should either be a path or a Buffer like in `PemKeyCertOptions`

BillyYccc:
I have reworked the internal implementaions of MySQL to drop using the `InitCommand` due to the lack of extensibilities in https://github.com/eclipse-vertx/vertx-sql-client/pull/420/commits/f8db677682c0786b1c0c88a0ecd99864b0710517.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLConnectOptions.java
BillyYccc: updated in https://github.com/eclipse-vertx/vertx-sql-client/pull/420/commits/6225f912d3f6d37f8afee8fb030b851f0c4bf4e5.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLConnectOptions.java
```diff
@@ -145,6 +146,26 @@ public MySQLConnectOptions setSsl(boolean ssl) {
     return this;
   }
 
+  /**
+   * Set the server RSA public key which is mostly used for encrypting password under insecure connections when performing authentication.
+   *
+   * @param serverRsaPublicKeyOptions the options of the server RSA public key
+   * @return a reference to this, so the API can be used fluently
+   */
+  public MySQLConnectOptions setServerRsaPublicKeyOptions(MySQLServerRsaPublicKeyOptions serverRsaPublicKeyOptions) {
```
vietj: for the sake of simplicity I think we don't need a nested class but only `rsaPublicKeyPath` and `rsaPublicKeyValue` fields in these options

BillyYccc: I have changed this in https://github.com/eclipse-vertx/vertx-sql-client/pull/420/commits/2612e315b322a5c7335dfd56840bdcaa7f1ab824.

vertx-mysql-client/src/test/java/io/vertx/mysqlclient/MySQLTLSTest.java
```diff
@@ -191,4 +192,29 @@ public void testConnFail(TestContext ctx) {
...
+    MySQLConnection.connect(vertx, options, ctx.asyncAssertSuccess(conn -> {
+      conn.query("SELECT current_user()", ctx.asyncAssertSuccess(res1 -> {
+        Row row1 = res1.iterator().next();
+        String username = row1.getString(0);
+        ctx.assertEquals("mysql", username.substring(0, username.lastIndexOf('@')));
+        MySQLConnectOptions changeUserOptions = new MySQLConnectOptions()
+          .setUser("superuser")
+          .setPassword("password")
+          .setDatabase("emptyschema");
+        conn.changeUser(changeUserOptions, ctx.asyncAssertSuccess(v2 -> {
```
vietj: why are we reusing `MySQLConnectionOptions` ? does the `changeUser` method needs to perform a reconnection ? if not then we should have a specific options class for this instead of reusing the connect options

BillyYccc: Change user will perform a re-authentication but not reconnect. The options used here is for reusing a subset of the connect options to authenticate, do you think it’s necessary to define a new class here? I think the code may be duplicated with `SqlConnectOptions` and `MySQLConnectOptions`.

vietj: yes I think it's better to have a new options class usability wise

vietj: perhaps also we should offer to do automatic user change at connection time

BillyYccc: what does `automatic user change at connection time` mean?

BillyYccc: The `CHANGE_USER` API is updated in https://github.com/eclipse-vertx/vertx-sql-client/pull/420/commits/f083dd128daf3a0772bb7b3672a5648c17b48a8f.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLChangeUserOptions.java
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  /**
+   * Add a property for this client, which will be sent to server at the re-authentication.
+   *
+   * @param key   the value of property key
+   * @param value the value of property value
+   * @return a reference to this, so the API can be used fluently
+   */
+  public MySQLChangeUserOptions addProperty(String key, String value) {
```
vietj: I think we need to annotate this with `@GenIgnore` because it generates a duplicate converter called `propertys`

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLChangeUserOptions.java
```diff
@@ -0,0 +1,249 @@
...
+    return this;
+  }
+
+  /**
+   * Get the path of the server RSA public key.
+   *
+   * @return a reference to this, so the API can be used fluently
+   */
+  public String getServerRsaPublicKeyPath() {
+    return serverRsaPublicKeyPath;
```
vietj: do we need to have the server prefix ? because this key has to be for the server right ?

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLChangeUserOptions.java
```diff
@@ -0,0 +1,249 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import static io.vertx.mysqlclient.MySQLConnectOptions.*;
+
+/**
+ * Authentication options for MySQL CHANGE_USER command.
+ */
+@DataObject(generateConverter = true)
+public class MySQLChangeUserOptions {
```
vietj: I think we should call it `MySQLAuthOptions`

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLChangeUserOptions.java
```diff
@@ -0,0 +1,249 @@
...
+
+  private void init() {
+    this.user = DEFAULT_USER;
+    this.password = DEFAULT_PASSWORD;
+    this.database = DEFAULT_SCHEMA;
+    this.charset = DEFAULT_CHARSET;
+    this.properties = new HashMap<>(DEFAULT_CONNECTION_ATTRIBUTES);
+  }
+
+  private void checkParameterNonNull(Object parameter, String message) {
```
vietj: use instead `Objects#requireNonNull` method, it throws `NullPointerException` and this is the correct exception we should throw

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLChangeUserOptions.java
BillyYccc: it's better naming because this is the server public key which is stored locally on the client side.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLChangeUserOptions.java
BillyYccc: updated in https://github.com/eclipse-vertx/vertx-sql-client/pull/420/commits/0d847058c7205b566256252cb111b79f2a61b698.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLChangeUserOptions.java
BillyYccc: added in https://github.com/eclipse-vertx/vertx-sql-client/pull/420/commits/4145bd592180460c67cdbbb7e8efff435ad3d830.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLChangeUserOptions.java
BillyYccc: updated in https://github.com/eclipse-vertx/vertx-sql-client/pull/420/commits/b4967b2c93f2ba463d1106f30a97edf06a25cd26.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/AuthenticationCommandBaseCodec.java
```diff
@@ -0,0 +1,95 @@
...
+            rsaPublicKeyRequest.writeByte(AUTH_PUBLIC_KEY_REQUEST_FLAG);
+            sendNonSplitPacket(rsaPublicKeyRequest);
+          } else {
+            // send encrypted password
+            sendEncryptedPasswordWithServerRsaPublicKey(password, serverRsaPublicKey.toString());
+          }
+        }
+      } else if (flag == FAST_AUTH_STATUS_FLAG) {
+        // fast auth success
+        return;
```
vietj: return statement seems not needed

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/AuthenticationCommandBaseCodec.java
```diff
@@ -0,0 +1,95 @@
...
+      } else {
+        completionHandler.handle(CommandResponse.failure(new UnsupportedOperationException("Unsupported flag for AuthMoreData : " + flag)));
+        return;
```
vietj: return statement seems not needed

BillyYccc: removed in https://github.com/eclipse-vertx/vertx-sql-client/pull/420/commits/31abea90fbec976a664eddbba0682657da355651.

vietj:
thanks for the contribution, great work

BillyYccc:
Thank you for the review!

PR #424, opened 2019-09-21
Wrapped list should implement List interface to avoid mismatched types - fixes #413

### LINKED ISSUE #413: CCE when using `Tuple.wrap(Object[])`
When using `Tuple.wrap(Object[])` I get an instance of `ListTuple` which is not a `java.util.List` (unlike `ArrayTuple`), which leads me to this CCE:

```
java.lang.ClassCastException: io.vertx.sqlclient.impl.ListTuple cannot be cast to java.base/java.util.List
	at io.vertx.sqlclient.impl.SqlClientBase.lambda$preparedQuery$0(SqlClientBase.java:81)
	at io.vertx.sqlclient.impl.SqlConnectionImpl.lambda$schedule$0(SqlConnectionImpl.java:52)
	at io.vertx.sqlclient.impl.SocketConnectionBase.handleMessage(SocketConnectionBase.java:210)
	at io.vertx.sqlclient.impl.SocketConnectionBase.lambda$init$0(SocketConnectionBase.java:81)
```

Due to https://github.com/eclipse-vertx/vertx-sql-client/blob/3.8/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlClientBase.java#L80:

```java
  private <R1, R2 extends SqlResultBase<R1, R2>, R3 extends SqlResult<R1>> C preparedQuery(
    String sql,
    Tuple arguments,
    Function<R1, R2> factory,
    Collector<Row, ?, R1> collector,
    Handler<AsyncResult<R3>> handler) {
    schedule(new PrepareStatementCommand(sql), cr -> {
      if (cr.succeeded()) {
        PreparedStatement ps = cr.result();
        String msg = ps.prepare((List<Object>) arguments);
```

Note that I see the same cast in `preparedBatch`.

BillyYccc:


vietj:
I don't get why this is needed

BillyYccc:
> I don't get why this is needed

this is a fix for #413

vietj:
ok I see, I would rather then instead change `PreparedStatement#prepare(List<Object>)` to PreparedStatement#prepare(Tuple)` then instead of having this PR

PR #436, opened 2019-10-16
User can configure how affected rows are counted

### LINKED ISSUE #434: sql-client: CLIENT_FOUND_ROWS client flag is not set, wheras the flag ist set for JDBC as default
Today I spent about an hour debug an issues, where an "update" statement despite being successful returned zero rows. I'm a postgres guy so my experience with MySQL is limited.

After looking into how this works in JDBC I found out that `https://dev.mysql.com/doc/refman/5.7/en/mysql-affected-rows.html` returns zero, if the update statement was successful but did not update the tuple.

The JDBC driver sets [`CLIENT_FOUND_ROWS`](https://github.com/mysql/mysql-connector-j/blob/36534fa273b4d7824a8668ca685465cf8eaeadd9/src/main/protocol-impl/java/com/mysql/cj/protocol/a/NativeAuthenticationProvider.java#L176) to `true` per default, which differs from the `vertx-mysql-client`:

```
useAffectedRows

Don't set the CLIENT_FOUND_ROWS flag when connecting to the server (not JDBC-compliant, will 
break most applications that rely on "found" rows vs. "affected rows" for DML statements), but does cause 
"correct" update counts from "INSERT ... ON DUPLICATE KEY UPDATE" statements to be returned by 
the server.

Default: false
```

I think that this flag should be the default to yield the same results as the `vertx-postgres-client`.

AndreasKl:
Fixes #434

### DIFF
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLConnectOptions.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLConnectOptions.java
@@ -129,7 +129,7 @@ public class MySQLConnectOptions extends SqlConnectOptions {
    *
    * @return how affected rows are calculated on update/delete/insert.
    */
-  public boolean getUseAffectedRows() {
+  public boolean isUseAffectedRows() {
     return useAffectedRows;
   }
 
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLConnectOptions.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLConnectOptions.java
@@ -36,6 +36,7 @@ public class MySQLConnectOptions extends SqlConnectOptions {
   public static final String DEFAULT_PASSWORD = "";
   public static final String DEFAULT_SCHEMA = "";
   public static final String DEFAULT_CHARSET = "utf8mb4";
+  public static final boolean DEFAULT_USE_AFFECTED_ROWS = false;
   public static final Map<String, String> DEFAULT_CONNECTION_ATTRIBUTES;
   public static final SslMode DEFAULT_SSL_MODE = SslMode.DISABLED;
 
@@ -47,6 +48,7 @@ public class MySQLConnectOptions extends SqlConnectOptions {
 
   private String collation;
   private String charset;
+  private Boolean useAffectedRows;
   private SslMode sslMode;
   private String serverRsaPublicKeyPath;
   private Buffer serverRsaPublicKeyValue;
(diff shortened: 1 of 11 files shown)


vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLConnectOptions.java
```diff
@@ -119,6 +124,29 @@ public MySQLConnectOptions setCharset(String charset) {
     return this;
   }
 
+  /**
+   * Get how affected rows are calculated on update/delete/insert.
+   *
+   * @return how affected rows are calculated on update/delete/insert.
+   */
+  public boolean getUseAffectedRows() {
```
BillyYccc: rather name the method `isUseAffectedRows` conventionally  and recodengen.

vertx-mysql-client/src/test/java/io/vertx/mysqlclient/MySQLClientCapabilitiesTest.java
```diff
@@ -0,0 +1,55 @@
...
+    testAffectedRowsClientCapability(ctx, false, 1);
+  }
+
+  private void testAffectedRowsClientCapability(TestContext ctx, boolean useAffectedRows, int expectedRowCount) {
+    MySQLConnectOptions connectOptions = options.setUseAffectedRows(useAffectedRows);
+    MySQLConnection.connect(vertx, connectOptions, ctx.asyncAssertSuccess(conn -> {
+      conn.query(PREPARE_TESTING_TABLE_DATA, ctx.asyncAssertSuccess(res0 -> {
+        conn
+          .query("UPDATE vehicle SET type = 'bike' WHERE id = 1;", ctx.asyncAssertSuccess(res1 -> {
+            ctx.assertEquals(expectedRowCount, res1.rowCount());
```
BillyYccc: add a `conn.close` here when the checking is done.

vertx-sql-client/src/test/java/io/vertx/sqlclient/tck/PreparedQueryTestBase.java
```diff
@@ -138,6 +138,18 @@ public void testPreparedUpdate(TestContext ctx) {
     }));
   }
 
+  @Test
+  public void testPreparedUpdateWithoutChange(TestContext ctx) {
```
BillyYccc: I don't think we should add this to the TCK, because different database servers by default behave differently to return a rowcount and they are not bound to return the found rows here.

AndreasKl:
Updated the PR

vietj:
can we merge this @BillyYccc ?

BillyYccc:
looks great now 👍

vietj:
thanks @BillyYccc

BillyYccc:
thanks for the contribution! @AndreasKl

BillyYccc:
`useAffectedRows` has been supported and it can be used to enable/disable the `CLIENT_FOUND_ROWS` flag when connecting to the server. `CLIENT_FOUND_ROWS` is enabled by default now so the `SqlResult#rowCount` method returns the found rows instead of affected rows for DML queries by default in MySQL.

PR #472, opened 2019-12-05
MySQL lastInsertId should be represented as Long type instead of Integer

### LINKED ISSUE #471: `MySQLClient.LAST_INSERTED_ID` should be `Long`
According to the [reference](https://dev.mysql.com/doc/refman/8.0/en/mysql-insert-id.html):

```
uint64_t mysql_insert_id(MYSQL *mysql)
```

So, `LAST_INSERTED_ID` should be:

```
PropertyKind<Long> LAST_INSERTED_ID = () -> Long.class;
```

BillyYccc:
Fix #471


BillyYccc:
This is an API change and we need to make it on the breaking change list. @vietj

Issue #504, opened 2020-01-14
MySQL TLS handshake failure could not be reported correctly


BillyYccc:
The CI does not pass in https://travis-ci.org/eclipse-vertx/vertx-sql-client/builds/636987648 and it is due to MySQL TLS tests failure.

This is because of regression in https://github.com/eclipse-vertx/vert.x/commit/3b685e70b3635a2ae5830efd52dbabe673a7cdcd.

MySQL handshake failure will trigger https://github.com/eclipse-vertx/vertx-sql-client/blob/d96e53f80c0066b377c43efbd1c6466bb0d06e51/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/InitialHandshakeCommandCodec.java#L159

which then calls `chctx.fireChannelRead(resp);`, but the connection is closed thus the message could not be processed, so the message handler could not be called. Then the test hangs and fails for being timeout.

BillyYccc:
it should have been fixed in both `3.8` and `master` branch now.

PR #505, opened 2020-01-14
Adapt to use the upgradeToSsl method in 4.0 version

BillyYccc:
Vert.x 4.0 provides an `NetSocket#upgradeToSsl()` with an asynchronous result so we don't need the tricks, the `MySQLSocketConnection#upgradeToSSLConnection` can also be omitted so we save an asyncresult instantiation.

This could get #504 fixed in `master` branch quickly, because currently `InitiateSslHandler` does not handle SSL handshake failure, instread it propagates the exception and get the connection closed.


vietj:
You can merge it @BillyYccc

PR #506, opened 2020-01-14
Get the SSL handshake failure being reported to messageHandler correctly

### LINKED ISSUE #504: MySQL TLS handshake failure could not be reported correctly

BillyYccc:
fix #504

### DIFF
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/InitiateSslHandler.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/InitiateSslHandler.java
@@ -41,7 +41,6 @@ public class InitiateSslHandler extends ChannelInboundHandlerAdapter {
 
   @Override
   public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
-    cause.printStackTrace();
     if (cause instanceof DecoderException) {
       DecoderException err = (DecoderException) cause;
       cause = err.getCause();
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/MySQLEncoder.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/MySQLEncoder.java
@@ -49,6 +49,14 @@ class MySQLEncoder extends ChannelOutboundHandlerAdapter {
     codec.completionHandler = resp -> {
       CommandCodec c = inflight.poll();
       resp.cmd = (CommandBase) c.cmd;
+      /*
+       * a bit hacky but we need this message delivered to the socket messageHandler
+       */
+      if (resp.cause().getMessage().equals("SSL handshake failed")) {
+        socketConnection.handleMessage(resp);
+        return;
+      }
+
       chctx.fireChannelRead(resp);
     };
     inflight.add(codec);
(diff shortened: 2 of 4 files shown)


vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/InitiateSslHandler.java
```diff
@@ -41,7 +41,6 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
 
   @Override
   public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
-    cause.printStackTrace();
```
aguibert: this change will have a merge conflict with your other PR #505, I assume you can just undo this change?

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/MySQLEncoder.java
```diff
@@ -49,6 +49,14 @@ void write(CommandBase<?> cmd) {
     codec.completionHandler = resp -> {
       CommandCodec c = inflight.poll();
       resp.cmd = (CommandBase) c.cmd;
+      /*
+       * a bit hacky but we need this message delivered to the socket messageHandler
+       */
+      if (resp.cause() != null && resp.cause().getMessage().equals("SSL handshake failed")) {
```
aguibert: this does seem a bit hacky, where does the error message come from? Is there any chance it could change or be a localized (non-english) message?

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/InitiateSslHandler.java
BillyYccc: no because this PR is for `3.8` branch.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/MySQLEncoder.java
BillyYccc: The error is from https://github.com/eclipse-vertx/vertx-sql-client/blob/6d38d4b9f442642ecc6b6de425e23d012cd7270b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/InitiateSslHandler.java#L56

Because in Vert.x 3.8 the `NetSocket#upgradeToSsl` method does not have an async result to represent if it fails, so we let the failure propagated in the channel pipeline and gets the connection closed, then we can check the activity of channel to know if the SSL handshake is successful or not. We only need this trick in `3.8` branch which is different what I do in my another PR for `master` branch(i.e. 4.0 version).

vietj:
@BillyYccc can we get a test for this, e.g using a wrong or expired certificate

BillyYccc:
I have added a test which the client connects with a self-signed certificate to make the TLS handshake fail.

BillyYccc:
can we merge this now?

PR #513, opened 2020-02-03
Add Postgres resultset metadata and column metadata support

### LINKED ISSUE #470: [Postgres]Provide access to resultset metadata
Please make enum `DataType` and `PgColumnDesc` public (package `io.vertx.pgclient.impl.codec`), previously they were like that

BillyYccc:
Motivation:

Users can get the metadata of columns and resultset now - fixes #470.

### DIFF
--- a/vertx-pg-client/src/main/asciidoc/index.adoc
+++ b/vertx-pg-client/src/main/asciidoc/index.adoc
@@ -216,6 +216,13 @@ You can fetch generated keys with a 'RETURNING' clause in your query:
 {@link examples.PgClientExamples#returning(io.vertx.sqlclient.SqlClient)}
 ----
 
+You may also want to get the column metadata of the resultset such as the field data type oid or the data type length:
+
+[source,$lang]
+----
+{@link examples.PgClientExamples#resultsetMetadata(io.vertx.sqlclient.SqlClient)}
+----
+
 include::connections.adoc[]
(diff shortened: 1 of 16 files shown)


BillyYccc:
@vietj can you please take a look at this?

vietj:
is there a pressing need for this ? I would like to study how we can do that for all clients, i.e do we need a same API for all databases ?

BillyYccc:
That means we need to build an abstraction to describe resultset metadata but the information provided by various vendor database servers is so different that it's hard to cover all the cases using a unified API.

aguibert:
@vietj There is now a moderate need for this item. Hibernate Reactive needs some sort of metadata similar to what JDBC metadata offers in order to perform the full set of schema management operations.

+1 that we should consider this at a generic API level though. This feature would only be useful to Hibernate Reactive if it were a generic API, in the current form (PG only) it wouldn't be easily usable by upstack libs.

gavinking:
> Hibernate Reactive needs some sort of metadata similar to what JDBC metadata offers in order to perform the full set of schema management operations.

Well, wait. It looks like this issue refers to result set metadata. That's not what Hibernate schema management uses. What it uses is *table* metadata (information schema).

vietj:
closing this as now we have `Columndescritpor` in master

PR #517, opened 2020-02-10
Implement incremental cursor fetch

### LINKED ISSUE #492: [DB2] Fetch results incrementally
The initial implementation of the DB2 client assumes all rows are returned at the same time (default fetch size of 64 rows). One important advantage of a reactive client is the ability to stream results back incrementally. 

Implement functionality that allows results to be fetched in batches (numResults > fetchSize). This work item should also include the ability to configure the fetch size.

aguibert:
Closes #492

### DIFF
--- a/vertx-db2-client/src/main/java/examples/DB2ClientExamples.java
+++ b/vertx-db2-client/src/main/java/examples/DB2ClientExamples.java
@@ -20,6 +20,7 @@ import java.nio.file.Paths;
 import java.sql.CallableStatement;
 import java.sql.Connection;
 import java.sql.DriverManager;
+import java.sql.PreparedStatement;
 import java.sql.Types;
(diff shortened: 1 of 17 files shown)


vertx-db2-client/src/main/java/examples/DB2ClientExamples.java
```diff
@@ -51,11 +54,28 @@
     
     public static void runJDBC() throws Exception {
         try (Connection con = DriverManager.getConnection("jdbc:db2://" + HOST + ":" + PORT + "/" + DB_NAME, DB_USER, DB_PASS)) {
-            runInitSql(con);
```
aguibert: Changes to this file can be ignored, as it is just for local experimentation (in the future I won't check these in)

PR #520, opened 2020-02-20
Postgres testing cleanup

### LINKED ISSUE #385: Testcontainers migration
Migrate our embedded databases for testing to [Testcontainers](https://www.testcontainers.org/) discussed in https://github.com/eclipse-vertx/vertx-sql-client/pull/381#issuecomment-515672725. (Actually we have given a try a long time before)

Pros:
* Single dependency on the `Testcontainers`
* Better support for the new features and new versions
* the same way of using it in testing with different databases
* Easier for extensibility to other databases

Cons:
* Looks like we can not test the Unix socket feature in the Mac OS X(see https://github.com/docker/for-mac/issues/483), but I think it will be fine since that part of code will not change too much.

### LINKED ISSUE #386: Boosting tests
Currently testing is slow and the most notable reason is we frequently start/stop the server and initialize it with an init script during our tests. We can make the server to be shared which means the shared one won't stop until all the tests end.

In order to achieve this, we need to try to make all the tests stateless(which means it can be rerun several times and still pass). 
I have given a try on MySQL and it works perfectly, the Postgres tests need some changes to adapt to this but it should not be too hard.

BillyYccc:
Motivation:

Rework the Postgres testing 

* Migrate the Unix domain socket tests with using the TestContainers - fixes #385 
* Boost the Postgres Testing - fixes #386


vietj:
@BillyYccc this is cool for domain sockets, however this is not supported on OSX (due to https://github.com/docker/for-mac/issues/483) which makes impossible to test on OSX domain sockets which is a blocker for testing domain sockets on OSX.

BillyYccc:
I know that but it could not be avoided if we test Unix domain socket with Docker on the OS X system, so the domain socket tests are skipped if the server is started by TestContainers on OS X.

Or should we deploy a specific database in CI with OS X for this kind of stuff?

vietj:
cleanup

PR #523, opened 2020-02-21
Initial transaction implementation for DB2

### LINKED ISSUE #493: [DB2] Transaction support
Implement the ability to explicitly commit and rollback transactions using the SqlConnection#begin() method

aguibert:
Fixes #493 

By default, DB2 flows the "commit" (`RDBCMM`) command after each SQL UPDATE statement, which is effectively JDBC auto-commit behavior. To account for DB2's implicit transactions I had to add a `boolean autoCommit()` method to the internal ExtendedQuery and SimpleQuery commands, so that the codec implementations can know whether or not they should be flowing an auto-commit after UPDATE statements.  

Also CC @mswatosh for review


BillyYccc:
I think it's db2 specific, are we able to handle it in the db2 implementation instead of the common SPI?

vertx-db2-client/src/main/java/io/vertx/db2client/impl/codec/ExtendedQueryCommandCodec.java
```diff
@@ -121,7 +118,8 @@ private void decodeUpdate(ByteBuf payload) {
         int updatedCount = (int) updateResponse.readExecute();
         // TODO: If auto-generated keys, read an OPNQRY here
         // readOpenQuery()
-        updateResponse.readLocalCommit();
+        if (!cmd.inTransaction())
```
vietj: missing curly braces for statement

vietj:
I believe we should have an approach where we make this behaviour as part of the command/response protocol so this concern is moved closer to the database behaviour and with more flexiliby, creating less coupling between the client API and the database protocol definitions.

This will also make easier to implement for the https://github.com/vert-x3/vertx-jdbc-client/tree/sql-client-api

vertx-db2-client/src/test/java/io/vertx/db2client/DB2TransactionTest.java
```diff
@@ -0,0 +1,209 @@
...
+import io.vertx.ext.unit.TestContext;
+import io.vertx.ext.unit.junit.VertxUnitRunner;
+import io.vertx.sqlclient.PoolOptions;
+import io.vertx.sqlclient.Row;
+import io.vertx.sqlclient.SqlClient;
+import io.vertx.sqlclient.Transaction;
+import io.vertx.sqlclient.Tuple;
+
+@RunWith(VertxUnitRunner.class)
+public class DB2TransactionTest {
```
vietj: could we get these tests in the TCK ?

Issue #490, opened 2020-01-07
[DB2] Connection error path handling


aguibert:
Implement proper handling and messages for connection error paths (e.g. try to connect with invalid username/password).

Related TCK tests are in `DB2ConnectionTest`

aguibert:
having @mswatosh look at this one

PR #529, opened 2020-02-26
Fixing invalid database path and close connection handling

mswatosh:
Fixing the close path for the db2 driver. 
Fixing an issue with the invalid database names, when the db2 server stops responding properly.
All the connection tests for the TCK are passing now.

Related to #490

Signed-off-by: Mark Swatosh <mark.swatosh@gmail.com>

### DIFF
--- a/vertx-db2-client/src/main/java/io/vertx/db2client/impl/codec/InitialHandshakeCommandCodec.java
+++ b/vertx-db2-client/src/main/java/io/vertx/db2client/impl/codec/InitialHandshakeCommandCodec.java
@@ -47,6 +47,12 @@ class InitialHandshakeCommandCodec extends AuthenticationCommandBaseCodec<Connec
     @Override
     void encode(DB2Encoder encoder) {
         super.encode(encoder);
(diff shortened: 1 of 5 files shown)


vertx-db2-client/src/main/java/io/vertx/db2client/impl/codec/InitialHandshakeCommandCodec.java
```diff
@@ -47,6 +47,12 @@
     @Override
     void encode(DB2Encoder encoder) {
         super.encode(encoder);
+        encoder.socketConnection.closeHandler(h -> {
+          if (status == ST_CONNECTING) {
+            //Sometimes DB2 closes the connection when sending an invalid Database name.
+            cmd.fail(new Throwable("Socket closed during connection"));
```
aguibert: should this be a `DB2Exception` instead of `Throwable`? Also, in most cases when this handler runs this `InitialHandshakeCommand` will already be completed, so calling `cmd.fail()` on a command that previously completed normally seems odd

PR #537, opened 2020-03-12
Initial progress on batch queries

### LINKED ISSUE #522: [DB2] Prepared batch statements
Get all tests passing in `DB2PreparedBatchTest` and investigate, implement, and test the rest of the batch API

aguibert:
Fixes #522


aguibert:
@BillyYccc @vietj can you review please? I added some shared TCK tests for the following:
1. Doing a batch SELECT query
2. Verify failure when an empty batch is submitted (had to change behavior in PG for this, it currently no-opped)
3. Verify failure when incorrect number of batch args are supplied (e.g. 2 params in SQL string but only 1 value provided in the `Tuple`)

PR #538, opened 2020-03-13
Sql template

vietj:
SQL templates for Vert.x 4

Latest generated documentation https://gist.github.com/vietj/db18ad21ea242f7260a8a824f6d6d3a8


vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/BatchTemplate.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @GenIgnore
+  default Future<Void> batch(List<T> list) {
+    Promise<Void> promise = Promise.promise();
+    batch(list, promise);
+    return promise.future();
+  }
+
+  @GenIgnore
+  void batch(List<T> list, Handler<AsyncResult<Void>> result);
```
BillyYccc: How about naming it `batchUpdate` since it returns an asynchronous result of `Void` type?

knotenpunkt:
when do you think to merge this PR?^^

And Julien, you should really take a look on my two BUG-issues in the vertx-repository.
The other once, are more or less feature request, wrong labeled, you should adjust them.

vietj:
link to current doc https://gist.github.com/vietj/db18ad21ea242f7260a8a824f6d6d3a8

zenios:
Hi

My only concern is with annotation pollution. To many annotations needs to be added to a class in order to be @RowMapped and @ParamsMapped.

Why don't we have @TemplateParam and @Column added to all getters by default and provide a way to exclude it. Just like lombok does for getter / setter generation

In the end most people map one to one from table to bean

vietj:
@zenios this is a good concern and I believe it is a possible choice given we provide a way to exclude some fields from the mapping. Keep in mind also that such data object can also be mapped to/from JSON using the json code generator too

vietj:
alright folks, this is now ready for review

vertx-sql-client-template/src/main/asciidoc/index.adoc
```diff
@@ -0,0 +1,325 @@
...
+  <artifactId>${maven.artifactId}</artifactId>
+  <version>${maven.version}</version>
+</dependency>
+----
+* Gradle (in your `build.gradle` file):
+
+[source,groovy]
+----
+dependencies {
+  compile '${maven.groupId}:${maven.artifactId}:${maven.version}'
```
jponge, line 25: Use `implementation` rather than `compile`.

jponge: But that's a more general remark. `compile` is still here, but it's recommended to go with `implementation` for dependencies you consume, and `api` for dependencies you want to consume the re-export.

vertx-sql-client-template/src/main/asciidoc/index.adoc
```diff
@@ -0,0 +1,325 @@
...
+
+SQL template produces (by default) a `RowSet<Row>` like a client `PreparedQuery`. In fact the template is a thin
+wrapper for a `PreparedQuery`.
+
+[source,$lang]
+----
+{@link examples.TemplateExamples#queryExample}
+----
+
+When you need to perform an insert or update operation and you do not care of the result, you can use instead {@link io.vertx.sqlclient.template.SqlTemplate#forUpdate}:
```
jponge, line 43: you can use `{@link XYZ}` instead.

vertx-sql-client-template/src/main/asciidoc/index.adoc
```diff
@@ -0,0 +1,325 @@
...
+
+[source,$lang]
+----
+{@link examples.TemplateExamples#insertExample}
+----
+
+== Row mapping
+
+Templates produces {@link io.vertx.sqlclient.Row} as default result.
```
jponge, line 52: By default templates produce `{...}` as result type.

vertx-sql-client-template/src/main/asciidoc/index.adoc
```diff
@@ -0,0 +1,325 @@
...
+
+You can provide a function:
```
jponge, line 54: I'd put `to achieve row level mapping instead:` here with the full function and template call below

vertx-sql-client-template/src/main/asciidoc/index.adoc
```diff
@@ -0,0 +1,325 @@
...
+  <artifactId>vertx-sql-client-template</artifactId>
+  <version>${maven.version}</version>
+</dependency>
+----
+
+This feature can also be used in Gradle:
+
+[source]
+----
+compile "io.vertx:vertx-codegen:${maven.version}:processor"
```
jponge, line 225: `annotationProcessor` scope in Gradle

vertx-sql-client-template/src/main/asciidoc/index.adoc
```diff
@@ -0,0 +1,325 @@
...
+compile "io.vertx:vertx-sql-client-template:${maven.version}"
+----
+
+IDE provides usually support for annotation processors.
```
jponge, line 229: usually provide

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/SqlTemplate.java
```diff
@@ -0,0 +1,145 @@
+package io.vertx.sqlclient.template;
```
jponge, line 1: BTW you should have 2020 license headers in files (IntelliJ copyright update can help)

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/impl/SqlTemplate.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class SqlTemplate {
+
+  private static Pattern PARAM_PATTERN = Pattern.compile(":(\\p{javaUnicodeIdentifierStart}\\p{javaUnicodeIdentifierPart}*)");
```
zenios, line 14: This will not work for postgresql :: cast

This is how spring does it
https://github.com/spring-projects/spring-framework/blob/4ca64684f6d5b7414ca3628a6b372a7c586fd2a6/spring-jdbc/src/main/java/org/springframework/jdbc/core/namedparam/NamedParameterUtils.java#L82

Another solution is to use ${name} via this pattern

```
Pattern.compile("\\$\\{\\s*(.*?)\\s*}");
```

We could also cache the result

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/impl/SqlTemplateImpl.java
```diff
@@ -0,0 +1,108 @@
...
+    return sqlTemplate.mapTuple(paramsMapper.apply(params));
+  }
+
+  @Override
+  public void execute(T parameters, Handler<AsyncResult<R>> handler) {
+
+    queryMapper
+      .apply(client.preparedQuery(sqlTemplate.getSql()))
+      .execute(toTuple(parameters), handler);
+  }
```
zenios, line 82: This can be remove if we use the default method on interface

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/impl/SqlTemplateImpl.java
```diff
@@ -0,0 +1,108 @@
...
+
+  @Override
+  public void executeBatch(List<T> batch, Handler<AsyncResult<R>> handler) {
+    queryMapper.apply(client.preparedQuery(sqlTemplate.getSql()))
+      .executeBatch(batch
+        .stream()
+        .map(this::toTuple)
+        .collect(Collectors.toList()), handler);
+  }
```
zenios, line 98: This can be remove if we use the default method on interface

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlClientInternal.java
```diff
@@ -0,0 +1,17 @@
+package io.vertx.sqlclient.impl;
+
+import io.vertx.sqlclient.SqlClient;
+
+public interface SqlClientInternal extends SqlClient {
+
+  /**
+   * Append a query place holder in the {@code query}.
```
zenios, line 8: placeholder not place holder

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/impl/SqlTemplate.java
vietj: I like this idea

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/SqlTemplate.java
```diff
@@ -0,0 +1,145 @@
...
+
+  /**
+   * Execute a batch query with the {@code batch}.
+   *
+   * <p>Each item in the batch is mapped to a tuple.
+   *
+   * @param batch the batch
+   * @param handler the result handler
+   */
+  void executeBatch(List<I> batch, Handler<AsyncResult<R>> handler);
```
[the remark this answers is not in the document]
vietj: because in practice the underlying query will actually use internally a `Promise<R>` that is created from the handler, so if you pass a promise to an `executeBatch` it will not be rewrapped into another promise.

vertx-sql-client-template/src/main/java/examples/TemplateExamples.java
```diff
@@ -0,0 +1,397 @@
...
+
+  public void paramsUserMapper() {
+    Function<User, Map<String, Object>> PARAMS_USER_MAPPER = user -> {
+      Map<String, Object> params = new HashMap<>();
+      params.put("id", user.id);
+      params.put("firstName", user.firstName);
+      params.put("lastName", user.lastName);
+      return params;
+    };
+  }
```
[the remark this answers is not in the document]
vietj: this is referenced from `index.adoc`

vertx-sql-client-template/src/main/java/examples/TemplateExamples.java
```diff
@@ -0,0 +1,397 @@
...
+
+  public void rowUserMapper() {
+    Function<Row, User> ROW_USER_MAPPER = row -> {
+      User user = new User();
+      user.id = row.getInteger("id");
+      user.firstName = row.getString("firstName");
+      user.lastName = row.getString("lastName");
+      return user;
+    };
+  }
```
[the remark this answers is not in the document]
vietj: this is referenced from `index.adoc`

vietj:
I would prefer to keep `@Column` and `@TemplateParam` distincts because they don't have the same meaning.

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/impl/SqlTemplateImpl.java
vietj: This is important to keep this and delegate to the original query

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/impl/SqlTemplateImpl.java
vietj: This is important to keep this and delegate to the original query

vietj:
@zenios @jponge addressed comments - now the syntax uses `${...}` because using `:` syntax would require to write an SQL parser which is a non goal

zenios:
Will do another later.thanks

vietj:
Thank you.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlClientInternal.java
```diff
@@ -5,13 +5,13 @@
...
+   * Append a query placeholder in the {@code query}.
    *
    * <p>The index starts at {@code 0}
    *
    * @param queryBuilder the builder to append to
-   * @param index the place holder index
+   * @param index the placeholder index
    */
-  int appendQueryPlaceHolder(StringBuilder queryBuilder, int index, int current);
+  int appendQueryPlaceholder(StringBuilder queryBuilder, int index, int current);
```
zenios, line 15: We should document the return value

vertx-sql-client-template/src/main/asciidoc/index.adoc
```diff
@@ -0,0 +1,332 @@
+= Client SQL templates
```
zenios, line 1: SQL Client templates

vertx-sql-client-template/src/main/asciidoc/index.adoc
```diff
@@ -0,0 +1,332 @@
...
+:toc:
+
+SQL client templates is a small library design to facilitate the execution of SQL queries.
```
zenios, line 4: library designed

vertx-sql-client-template/src/main/asciidoc/index.adoc
```diff
@@ -0,0 +1,332 @@
...
+
+== Usage
+
+To use client SQL templates add the following dependency to the _dependencies_ section of your build descriptor:
```
zenios, line 8: To use SQL client templates

vertx-sql-client-template/src/main/asciidoc/index.adoc
```diff
@@ -0,0 +1,332 @@
...
+
+This feature can also be used in Gradle:
+
+[source]
+----
+annotationProcessor "io.vertx:vertx-codegen:${maven.version}:processor"
+compile "io.vertx:vertx-sql-client-template:${maven.version}"
+----
+
+IDE usually provide usually support for annotation processors.
```
zenios, line 236: IDEs usually provide support for annotation processors.

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/SqlTemplate.java
```diff
@@ -0,0 +1,145 @@
...
+import io.vertx.sqlclient.impl.SqlClientInternal;
+import io.vertx.sqlclient.template.impl.SqlTemplateImpl;
+
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collector;
+
+/**
+ * A SQL template.
```
zenios, line 21: An SQL template.

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/SqlTemplate.java
```diff
@@ -0,0 +1,145 @@
...
+ * simple map, a user defined mapping can be used instead given it maps the source to such a map.
+ *
+ * <p>SQL template default results are {@link Row}, a user defined mapping can be used instead, mapping the
+ * result set {@link Row} to a {@link RowSet} of the mapped type.
+ */
+@VertxGen
+public interface SqlTemplate<I, R> {
+
+  /**
+   * Create a SQL template for query purpose consuming map parameters and returning {@link Row}.
```
zenios, line 36: Create an SQL template

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/SqlTemplate.java
```diff
@@ -0,0 +1,145 @@
...
+   * @param template the template query string
+   * @return the template
+   */
+  static SqlTemplate<Map<String, Object>, RowSet<Row>> forQuery(SqlClient client, String template) {
+    io.vertx.sqlclient.template.impl.SqlTemplate sqlTemplate = io.vertx.sqlclient.template.impl.SqlTemplate.create((SqlClientInternal) client, template);
+    return new SqlTemplateImpl<>(client, sqlTemplate, Function.identity(), Function.identity());
+  }
+
+  /**
+   * Create a SQL template for query purpose consuming map parameters and returning void.
```
zenios, line 48: Create an SQL template

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/impl/SqlTemplateImpl.java
```diff
@@ -0,0 +1,108 @@
...
+  public static final Collector<Row, Void, Void> NULL_COLLECTOR = Collector.of(() -> null, (v, row) -> {}, (a, b) -> null);
+
+  protected final SqlClient client;
+  protected final io.vertx.sqlclient.template.impl.SqlTemplate sqlTemplate;
+  protected final Function<T, Map<String, Object>> paramsMapper;
+  protected Function<PreparedQuery<RowSet<Row>>, PreparedQuery<R>> queryMapper;
+
+  public SqlTemplateImpl(SqlClient client,
+                         io.vertx.sqlclient.template.impl.SqlTemplate sqlTemplate,
+                         Function<PreparedQuery<RowSet<Row>>, PreparedQuery<R>> foobar,
```
BillyYccc, line 33: `foobar` should be named `queryMapper`?

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlClientBase.java
```diff
@@ -37,7 +37,13 @@
 import java.util.function.Function;
 import java.util.stream.Collector;
 
-public abstract class SqlClientBase<C extends SqlClient> implements SqlClient, CommandScheduler {
+public abstract class SqlClientBase<C extends SqlClient> implements SqlClientInternal, CommandScheduler {
+
+  @Override
+  public int appendQueryPlaceholder(StringBuilder queryBuilder, int index, int current) {
+    queryBuilder.append("?");
```
BillyYccc, line 44: we need to override this method in `MSSQLConnectionImpl` and `MSSQLPoolImpl` for sql server client.
The syntax is like postgres as:
```
    queryBuilder.append('@').append(1 + index);
    return index;
```

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/generator/RowMapperGen.java
```diff
@@ -0,0 +1,215 @@
...
+  private static String rowType(TypeInfo type) {
+    DataObjectInfo dataObject = type.getDataObject();
+    if (dataObject != null) {
+      return rowType(dataObject.getJsonType());
+    }
+    return type.getName();
+  }
+
+  private static String getter(TypeInfo type) {
+    switch (type.getKind()) {
```
BillyYccc, line 167: it seems we don't support mapping string/numeric datat type to enum type(name/ordinal). I wonder if we could support this here.

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/generator/RowMapperGen.java
```diff
@@ -0,0 +1,215 @@
...
+        case "java.time.OffsetDateTime":
+          return "getOffsetDateTime";
+        case "java.time.temporal.Temporal":
+          return "getTemporal";
+        case "java.util.UUID":
+          return "getUUID";
+        case "io.vertx.core.buffer.Buffer":
+          return "getBuffer";
+      }
+      DataObjectInfo dataObject = type.getDataObject();
```
BillyYccc, line 200: can we add tests for data object type like geometrics data object and numeric data object to check whether it works?

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/generator/RowMapperGen.java
```diff
@@ -0,0 +1,215 @@
...
+      case STRING:
+        return "getString";
+      case JSON_OBJECT:
+        return "getJsonObject";
+      case JSON_ARRAY:
+        return "getJsonArray";
+    }
+    if (type instanceof ClassTypeInfo) {
+      ClassTypeInfo ct = (ClassTypeInfo) type;
+      switch (ct.getName()) {
```
BillyYccc, line 182: MySQL is using `java.time.Duration` type, can we add the mapping here using Row#get(java.time.Duration.class, index)?

vertx-sql-client-template/src/main/asciidoc/index.adoc
vietj: we might actually rename this to simply "SQL templates"

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlClientBase.java
vietj: oh good to know I will do it

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/generator/RowMapperGen.java
vietj: yes, I'm not sure they work, we'll write the tests and figure out after :-)

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/generator/RowMapperGen.java
vietj: I will try adding a specific tests for this using test container and implement it.

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/generator/RowMapperGen.java
vietj: I think we can map it to get/set String easily

vietj: at least for beginning, some might want to provide at some point a custom function for this type

vietj: I've checked in such mapping support in master for PostgreSQL and implement support for templates in this PR.

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/generator/RowMapperGen.java
vietj: I added support for this and specific integration tests

vertx-sql-client-template/src/main/java/io/vertx/sqlclient/template/generator/RowMapperGen.java
vietj: I added support and tests for such types

vietj: Actually they are mapped to generic types with `row.get(Box.class, pos)`

vietj:
Ready for another review

vietj:
Gist with docs has also be updated https://gist.github.com/vietj/db18ad21ea242f7260a8a824f6d6d3a8

vietj:
Any more feedback on this ?

zenios:
@vietj Skipping of fields will not be included on version 1?

vertx-sql-client-templates/src/main/asciidoc/index.adoc
```diff
@@ -0,0 +1,338 @@
...
+
+=== Code generation
+
+Any data object annotated by {@link io.vertx.sqlclient.templates.annotations.RowMapped} or {@link io.vertx.sqlclient.templates.annotations.ParametersMapped}
+will trigger  the generation of a corresponding mapper class.
+
+The _codegen_ annotation processor generates these classes at compilation time. It is a feature of the Java
+compiler so _no extra step_ is required, it is just a matter of configuring correctly your build:
+
+Just add the `io.vertx:vertx-codegen:processor` and `io.vertx:vertx-sql-client-template`
```
BillyYccc, line 208: should be `vertx-sql-client-templates`?

vertx-sql-client-templates/src/main/asciidoc/index.adoc
```diff
@@ -0,0 +1,338 @@
...
+----
+<dependency>
+  <groupId>io.vertx</groupId>
+  <artifactId>vertx-codegen</artifactId>
+  <version>${maven.version}</version>
+  <classifier>processor</classifier>
+</dependency>
+<dependency>
+  <groupId>io.vertx</groupId>
+  <artifactId>vertx-sql-client-template</artifactId>
```
BillyYccc, line 223: should be `vertx-sql-client-templates`?

vertx-sql-client-templates/src/main/asciidoc/index.adoc
```diff
@@ -0,0 +1,338 @@
...
+  <version>${maven.version}</version>
+</dependency>
+----
+
+This feature can also be used in Gradle:
+
+[source]
+----
+annotationProcessor "io.vertx:vertx-codegen:${maven.version}:processor"
+compile "io.vertx:vertx-sql-client-template:${maven.version}"
```
BillyYccc, line 233: `vertx-sql-client-templates`?

vertx-sql-client-templates/src/main/asciidoc/index.adoc
vietj: right

vertx-sql-client-templates/src/main/asciidoc/index.adoc
vietj: right

vertx-sql-client-templates/src/main/asciidoc/index.adoc
vietj: right

PR #539, opened 2020-03-13
Enable MySQL transaction TCK tests

BillyYccc:


BillyYccc:
@aguibert can you take a look at this? I changed the TCK tests a bit, the MySQL transaction TCK tests fail because the `SqlResult#rowCount` does not return rows by `SELECT` query but rather affected rows in MySQL and I use `SqlResult#size` instead.

BillyYccc:
thanks for reviewing

Issue #497, opened 2020-01-07
[DB2] Support for auto-generated keys


aguibert:
Typically databases will use auto-generated or auto-incrementing keys when there are uniqueness constraints for primary keys. It should be possible for users of the DB2 client to easily retrieve these auto-generated keys after issuing a query that auto-generates keys, similar to what the MySQL driver has for `io.vertx.mysqlclient.MySQLClient.LAST_INSERTED_ID`

aguibert:
Closing this since it was sufficiently covered by PR #547

PR #547, opened 2020-03-14
 Support for queries that return auto-generated keys

aguibert:
Part of #497 

Builds on top of commits from PR #537

### DIFF
--- a/vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/ClientTypes.java
+++ b/vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/ClientTypes.java
@@ -15,6 +15,7 @@
  */
 package io.vertx.db2client.impl.drda;
 
+import java.math.BigDecimal;
 import java.sql.Types;
 
 // This enumeration of types represents the typing scheme used by our jdbc driver.
@@ -77,7 +78,7 @@ public class ClientTypes {
     private ClientTypes() {
     }
 
-    static String getTypeString(int type) {
+    public static String getTypeString(int type) {
         switch (type) {
         case BIGINT:
             return "BIGINT";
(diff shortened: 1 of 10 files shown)


vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/DRDAQueryResponse.java
```diff
@@ -116,15 +116,50 @@ public void readOpenQuery() { // @AGG removed callback StatementCallbackInterfac
         endOfSameIdChainData();
     }
     
+    /**
+     * Parse the reply for the Open Query Command. This method handles the
+     * parsing of all command replies and reply data for the opnqry command.
+     */
+    public void readBeginOpenQuery() { // @AGG removed callback StatementCallbackInterface statementI) {
```
aguibert: In this file I'm intentionally leaving the commented out code in place because this is the code that's ported from Apache Derby and I'm keeping the commented out code for easy reference in the near future.

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/ClientTypes.java
```diff
@@ -198,4 +199,48 @@ static public int mapDB2TypeToDriverType(boolean isDescribed, int sqlType, long
...
+        		   clazz == Integer.class ||
+        		   clazz == long.class ||
+        		   clazz == Long.class ||
+        		   clazz == short.class ||
+        		   clazz == Short.class ||
+        		   clazz == BigDecimal.class;
+        case ClientTypes.BINARY:
+        case ClientTypes.BLOB:
+        	return clazz == boolean.class ||
+        		   clazz == Boolean.class ||
```
gjwatts: For BINARY and BLOB, why do we return true if it is a B/boolean?

aguibert: According to[ this DB2 doc](https://www.ibm.com/support/knowledgecenter/SSEPEK_12.0.0/intro/src/tpc/db2z_stringdatatypes.html) a BINARY column is a varying-length binary string where `0 < length <= 255` with a default length of 1. I think the most common use case here is to just store a length=1 binary string which would fit into a B/boolean Java type.

A BLOB is also a variable-length binary string where `0 <= length <= 2G` with a default size of 1M. Normally this will only fit into a byte array, but in some cases it could be length=1.

Overall, the checks in this method don't need to be perfect, they are just client-side checking and if we get some incompatible data the user will get an error message later on, but the error message we can give here will probably be more specific/clear than the error message later on.

PR #549, opened 2020-03-16
Refactor API to reduce surface area

aguibert:
Reduces API surface to a more DSL-style API. 

## Base SQLClient methods
 - `Query<RowSet<Row>> query(String sql)`
 - `PreparedQuery<RowSet<Row>> preparedQuery(String sql)`

## Query<R>
1. Altering settings
1.1 `Query<R> withOptions(QueryOptions options)`
2. Altering result type
2.1 `<T> Query<SqlResult<T>> asCollector(Collector<?, ?, T> collector)` 
3. Terminal operations
3.1 `Future<R> execute()` 
3.2 `void execute(Handler<AsyncResult<R>> handler)`

## PreparedQuery<R> extends Query<R>
1. Altering result type
1.1 `PreparedQuery<Cursor> asCursor()`
1.2 `PreparedQuery<RowStream<Row>> asStream(int fetchSize)`
2. Terminal operations
2.1 `Future<R> execute(Tuple args)`
2.2 `void execute(Tuple args, Handler<AsyncResult<R>> handler)`
2.3 `Future<R> executeBatch(List<Tuple> argsList)`
2.4 `void executeBatch(List<Tuple> argsList, Handler<AsyncResult<RowSet<Row>>> handler)`

## Examples

Basic query
```java
SqlClient client = // ...
client.query("SELECT * FROM users WHERE id='julien'")
  .execute(ar -> {
    // do stuff with results
  });
```

Prepared query with args
```java
SqlClient client = // ...
client.preparedQuery("SELECT * FROM users WHERE id=$1")
  .execute(Tuple.of("julien"), ar -> {
    // do stuff with results
  });
```

Prepared query with a Cursor
```java
SqlClient client = // ...
client.preparedQuery("SELECT * FROM users WHERE id=$1")
  .asCursor()
  .execute(Tuple.of("julien"), ar -> {
    Cursor cursor = ar.result();
    // do stuff with the cursor
  });
```

Prepared query with a Stream
```java
SqlClient client = // ...
client.preparedQuery("SELECT * FROM users WHERE first_name LIKE $1")
  .asStream(50) // fetch 50 rows at a time
  .execute(Tuple.of("julien"), ar -> {
    RowStream<Row> stream = ar.result();
    // do stuff with the stream
  });
```


vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlClient.java
```diff
@@ -43,129 +43,12 @@
...
-   */
-  @GenIgnore
-  <R> SqlClient preparedBatch(String sql, List<Tuple> batch, Collector<Row, ?, R> collector, Handler<AsyncResult<SqlResult<R>>> handler);
-
-  /**
-   * Like {@link #preparedBatch(String, List, Collector, Handler)} but returns a {@code Future} of the asynchronous result
-   */
-  @GenIgnore
-  <R> Future<SqlResult<R>> preparedBatch(String sql, List<Tuple> batch, Collector<Row, ?, R> collector);
+  PreparedQuery<RowSet<Row>> preparedQuery(String sql);
```
BillyYccc: This `PreparedQuery` isn't actually a **prepared** statement, and the `PREPARE` phase is executed lazily?

aguibert: Yes that's correct. I couldn't think of any use case where a user may need to execute some code between the PREPARE and first execution of a prepared statement

vietj: you might want to prepare a query before execution to ensure the query is valid

vietj: e.g you prepare the query in the start phase of the application and then use the prepared query at runtime

vietj:
The `SqlConnection#prepare` methods should not be removed because they allow to prepare and validate a query before it is executed. I don't think it is a problem because the `SqlClient` will still expose the `preparedQuery` method.

I am wondering if we should not rename `executeBatch` into simply `execute` and overload the `execute(Tuple)` method with `execute(List<Tuple>)` which means a batch execution.

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlClient.java
aguibert: do you mean to ensure the query is valid or to ensure the connection is valid?
Checking if the query is valid should be something that is done at application development time or at least in a testing env. 
If it's to check if the connection is valid, some drivers have a `SqlConnection#ping` method (which we should make standard IMO)

aguibert: also, if we want to keep the `prepare` method, what would it return? And how would we clearly distinguish that return type from the return type of `preparedQuery`?

Currently the proposed architecture is a 3-tier model that closely resembles JDBC (which is good because it will help people understand it quickly).
```
1) SqlConnection/Pool/Transaction
   2) Query<RESULT_TYPE> / PreparedQuery<RESULT_TYPE>
      3) RESULT_TYPE (one of RowSet, RowStream, Cursor, or COLLECTOR_TYPE)
```
Which conceptually maps nicely to JDBC:
```
1) Connection
  2) Statement / PreparedStatement / CallableStatement
    3) ResultSet
```

If we allow an intermediate `PreparedQuery` object for a PREPARE phase result, it inserts a new (infrequently used) layer between (1) and (2).

jponge:
> The `SqlConnection#prepare` methods should not be removed because they allow to prepare and validate a query before it is executed. I don't think it is a problem because the `SqlClient` will still expose the `preparedQuery` method.
> 
> I am wondering if we should not rename `executeBatch` into simply `execute` and overload the `execute(Tuple)` method with `execute(List<Tuple>)` which means a batch execution.

I'd keep `executeBatch` because it has clear samantics

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlClient.java
BillyYccc: how about keeping the `prepare` APIs unchanged and replace the `SqlClient` APIs with:
```
public interface SqlClient {
    Statement<RowSet<Row>> createStatement(String sql);

    void close();
}
```

```
public interface Statement<R> {

    <K> Statement<R> addOption(PropertyKind<K> propertyKind);

    <T> Statement<SqlResult<T>> asCollector(Collector<?, ?, T> collector);

    Future<R> query();

    Statement<R> query(Handler<AsyncResult<R>> handler);

    Future<R> execute();

    Statement<R> execute(Handler<AsyncResult<R>> handler);

    Future<R> execute(Tuple params);

    Statement<R> execute(Tuple params, Handler<AsyncResult<R>> handler);

    Future<R> executeBatch(List<Tuple> batch);

    Statement<R> executeBatch(List<Tuple> batch, Handler<AsyncResult<R>> handler);

}
```

In this way the `Statement` interface is only a container for SQL string and could be used for `simple query`, one-shot `prepared query` and `prepared batch`. And the `SqlConnection#prepare` and `PreparedQuery` interface works as before

vietj: I think the `Query` name is fine, no need to use the `Statement` name instead.

aguibert: @BillyYccc What about `Cursor` and `RowStream` result types that only apply to prepared queries? I think it's important to have the return-type-setting `asXXX()` methods on the interface for consistency. This is the main reason why I had the extended `PreparedQuery` interface in the original proposal. If we can apply Cursor and RowStream results to non-prepared statements, then it wouldn't be an issue.

Also, I think it's important to have consistent naming convention on the query execution methods (i.e. `execute()` and `executeBatch()`), it's not really clear that `query()` is so similar to `execute()`, and having a separate `Query` and `PreparedQuery` interface would alleviate that.

vietj:
I did this PR https://github.com/eclipse-vertx/vertx-sql-client/pull/554 , it implements the ideas expressed here.

Key differences:
- `PreparedQuery` does not extend `Query` and instead is renamed `PreparedStatement` and looses its ability to directly execute the query, instead a `query()` method is provided that returns a `Query` (you can see that as a flavour of prefer composition over inheritance).
- Cursor and streams are left apart and a treated as before from `PreparedStatement`

aguibert:
Closing this PR since it we ended up delivering a form of this proposal under PR #554 instead

Issue #499, opened 2020-01-08
[DB2] Binary Data TCK tests


aguibert:
Get the `DB2BinaryDataTypeDecodeTest` and `DB2BinaryDataTypeEncodeTest` tests passing

Sanne:
just wondering, does this issue imply that these tests are not passing, or in unknown status?

aguibert:
@Sanne right now they are just unknown status

PR #550, opened 2020-03-19
Enable binary data type TCK tests

aguibert:
For issue #499

### DIFF
--- a/vertx-db2-client/src/main/java/io/vertx/db2client/impl/codec/DB2ParamDesc.java
+++ b/vertx-db2-client/src/main/java/io/vertx/db2client/impl/codec/DB2ParamDesc.java
@@ -15,8 +15,11 @@
  */
 package io.vertx.db2client.impl.codec;
 
+import java.math.BigDecimal;
+
 import io.vertx.db2client.impl.drda.ClientTypes;
 import io.vertx.db2client.impl.drda.ColumnMetaData;
+import io.vertx.sqlclient.data.Numeric;
 import io.vertx.sqlclient.impl.ErrorMessageFactory;
 import io.vertx.sqlclient.impl.ParamDesc;
 import io.vertx.sqlclient.impl.TupleInternal;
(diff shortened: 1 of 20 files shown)


vertx-db2-client/src/main/java/io/vertx/db2client/impl/codec/DB2ParamDesc.java
```diff
@@ -40,12 +43,29 @@ public String prepare(TupleInternal values) {
...
+        case ClientTypes.BOOLEAN:
+        case ClientTypes.DECIMAL:
+        case ClientTypes.DOUBLE:
+        case ClientTypes.INTEGER:
+        case ClientTypes.REAL:
+        case ClientTypes.SMALLINT:
+        	return clazz == Numeric.class;
+    	}
+    	return false;
+    }
```
gjwatts: Why not just add "return clazz == Numeric.class" to canConvert in ClientTypes?  Is there a case where ClientTypes is a BIT, CHAR or LONGVARCHAR and you want to return a Numeric class only if it is not one of those types?

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/ColumnMetaData.java
```diff
@@ -105,21 +107,48 @@ public void setColumnCount(final int numColumns) {
...
+    	if (sqlxName_ != null && i < sqlxName_.length && sqlxName_[i] != null)
+    		return sqlxName_[i];
     	// Otherwise use column names from SQLDOPTGRP
-    	if (sqlName_ != null && sqlName_.length > 0)
-    		return Collections.unmodifiableList(Stream.of(sqlName_).collect(Collectors.toList()));
-    	return Collections.emptyList();
+    	if (sqlName_ != null && i < sqlName_.length)
+    		return sqlName_[i];
+    	return null;
     }
```
gjwatts: Should we add a check to make sure the passed in i >= 0?  Or are we guaranteed a value that is 0+

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/ListTuple.java
```diff
@@ -19,6 +19,7 @@
 
 import io.vertx.sqlclient.Tuple;
 
+import java.util.Iterator;
```
gjwatts: Remove?

gjwatts:
Should we update copyright dates in java files we change that are outside of DB2?

vertx-db2-client/src/main/java/io/vertx/db2client/impl/codec/DB2ParamDesc.java
aguibert: right, the other types like CHAR, LONGVARCHAR, (and in the future BLOB and CLOB) won't be convertable to Numeric

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/ColumnMetaData.java
aguibert: Good catch, I think this method will eventually be an SPI so I'll add the check

PR #553, opened 2020-03-20
Add tests for table joins and subqueries

### LINKED ISSUE #551: [DB] Support for table joins
SQL queries can join results from two or more tables using:
 - Inner Join
 - Left Outer Join
 - Right Outer Join
 - Full Outer Join

Ensure that all forms of table joins work correctly as described in [the DB2 documentation](https://www.ibm.com/support/knowledgecenter/SSEPEK_10.0.0/intro/src/tpc/db2z_joindatafromtables.html)

aguibert:
This PR mainly just adds tests for various queries which happened to work already.

Also, it exposed an issue where the DB2Codec was reading ahead slightly in order to check a "sequence ID" which was a holdover from the copied MySQL code. The sequence ID only applies to MySQL codec and therefore can be removed from DB2

Fixes #551

### DIFF
--- a/vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/ColumnMetaData.java
+++ b/vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/ColumnMetaData.java
@@ -115,6 +115,8 @@ public class ColumnMetaData {
     }
     
     public String getColumnName(int i) {
+    	if (i < 0)
+    		throw new IllegalArgumentException("Requested column name for negative index: " + i);
     	// Prefer column names from SQLDXGRP if set
     	if (sqlxName_ != null && i < sqlxName_.length && sqlxName_[i] != null)
     		return sqlxName_[i];
--- a/vertx-db2-client/src/test/java/io/vertx/db2client/DB2TestBase.java
+++ b/vertx-db2-client/src/test/java/io/vertx/db2client/DB2TestBase.java
@@ -1,8 +1,13 @@
 package io.vertx.db2client;
 
+import java.util.Collections;
+import java.util.List;
+
 import org.junit.After;
 import org.junit.Before;
 import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.rules.TestName;
 
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
@@ -21,12 +26,17 @@ public abstract class DB2TestBase {
 	protected Vertx vertx;
 	protected Connector<SqlConnection> connector;
 	protected DB2ConnectOptions options;
+	
+	@Rule
+	public TestName testName = new TestName();
 
 	@Before
 	public void setUp(TestContext ctx) throws Exception {
+		System.out.println(">>> BEGIN " + testName.getMethodName());
 		vertx = Vertx.vertx();
 		initConnector();
-		cleanTestTable(ctx);
+		for (String table : tablesToClean())
+			cleanTestTable(ctx, table);
 	}
 
 	@After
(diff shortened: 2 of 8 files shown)


vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/ColumnMetaData.java
```diff
@@ -115,6 +115,8 @@ public void setColumnCount(final int numColumns) {
     }
     
     public String getColumnName(int i) {
+    	if (i < 0)
```
aguibert: this change was requested from the previous PR

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/ListTuple.java
```diff
@@ -19,7 +19,6 @@
 
 import io.vertx.sqlclient.Tuple;
 
-import java.util.Iterator;
```
aguibert: this change was requested from the previous PR

vertx-db2-client/src/test/java/io/vertx/db2client/DB2TestBase.java
```diff
@@ -21,12 +26,17 @@
...
 
 	@Before
 	public void setUp(TestContext ctx) throws Exception {
+		System.out.println(">>> BEGIN " + testName.getMethodName());
 		vertx = Vertx.vertx();
 		initConnector();
-		cleanTestTable(ctx);
+		for (String table : tablesToClean())
+			cleanTestTable(ctx, table);
 	}
```
gjwatts: tablesToClean() always returns an empty list each time, so won't that mean cleanTestTable() will never be run?

vertx-db2-client/src/test/java/io/vertx/db2client/TableJoinTest.java
```diff
@@ -0,0 +1,100 @@
...
+@RunWith(VertxUnitRunner.class)
+public class TableJoinTest extends DB2TestBase {
+	
+	@Test
+	public void testColumnRename(TestContext ctx) {
+		connect(ctx.asyncAssertSuccess(conn -> {
+			conn.query("SELECT immutable.id AS \"IMM ID\"," +
+		                      "immutable.message AS IMM_MSG," +
+					          "Fortune.id AS FORT_ID," +
+		                      "Fortune.message AS \"FORT ID\" FROM immutable " +
```
gjwatts: This is a nit and completely irrelevant to the test itself, but on line 25, I would suggest changing \"FORT ID\" to \"FORT MSG\" to remove confusion with the line above.

vertx-db2-client/src/test/java/io/vertx/db2client/DB2TestBase.java
aguibert: by default yes, but if a test class wants to have tables cleaned they can override the `tablesToClean()` method like this:
```java
	protected List<String> tablesToClean() {
		return Collections.singletonList("someTable");
	}
```

vertx-db2-client/src/test/java/io/vertx/db2client/TableJoinTest.java
aguibert: good catch, I will correct that in a follow-on PR

vertx-db2-client/src/test/java/io/vertx/db2client/DB2TestBase.java
gjwatts: Ok, that makes sense.  Thanks for the explanation.

PR #554, opened 2020-03-24
Fluent queries

vietj:
Fluent queries based on the earlier proposal.

The refactor is done in several steps:

1. `PreparedQuery` becomes `PreparedQuery<T>`, collector overload are removed in favor of `collecting(Collector)` method, a new `mapping(Function<Row, T>)` is added

```java
// Before
conn
   .prepare(SQL)
   .flatMap(preparedQuery -> preparedQuery.execute(tuple, collector);

// After
conn
   .prepare(SQL)
   .flatMap(preparedQuery -> preparedQuery
      .collecting(collector)
      .execute(tuple);
```

2. Add a fluent `Query` interface used in `SqlClient`

```java
// Before
client.query(SQL, collector)
client.preparedQuery(SQL, tuple, collector)
client.preparedBatch(SQL, tuples, collector)

// After
client.createQuery(SQL).collecting(collector).execute();
client.createPreparedQuery(SQL).collecting(collector).execute(tuple);
client.createPreparedQuery(SQL).collecting(collector).batch(tuples);
```

Notes:

- `PreparedQuery` does not extend it (on purpose), only the collector variants are removed
- `createPreparedQuery` returns a `Query` and not a `PreparedQuery` (on purpose)
- no `createPreparedBatch` instead the `batch` method on `Query`

3. Rename `PreparedQuery` to `PreparedStatement`, nothing fancy here, the idea is that `prepare` returns something that is not a query but an explicit prepared statement

```java
// Before
conn.prepare(ar -> {
  if (ar.succeeded()) {
    PreparedQuery<RowSet<Row>> prepared = ar.result();
  }
});

// After
conn.prepare(ar -> {
  if (ar.succeeded()) {
    PreparedStatement<RowSet<Row>> prepared = ar.result();
  }
});
```

4. Remove `PreparedStatement` methods for query execution and instead provide a `query()` method returning `Query<RowSet<Row>>`, `PreparedStatement` is not generic anymore as this is now carried by the `Query` returned by `query()`

```java
// Before
conn.prepare().flatMap(preparedQuery -> preparedQuery.execute(tuple));
conn.prepare().flatMap(preparedQuery -> preparedQuery.execute(collector, tuple));

// After
conn
   .prepare()
   .flatMap(preparedStatement -> 
      preparedStatement
         .query()
         .execute(tuple));
conn
   .prepare()
   .flatMap(preparedStatement -> 
      preparedStatement
         .query()
         .collecting(collector)
         .execute(tuple));
```

5. (not done) rename `SqlClient#createQuery` to `SqlClient#query`, same for `createPreparedQuery`

```java
// Before
client.query(SQL);
client.createQuery(SQL).execute();

// After
client.query(SQL).execute();
```


vietj:
If that's fine, I'll finish this PR, get a review and merge it and back-port for 3.9.0 as this is an important change and it can break in 3.9.0 as it is tech preview.

I left aside the `QueryOptions` part as it can be added later and is not useful in 3.9.

Note that `sql-template` branch is rebased on this change to implement some of its features. In addition the generated `@DataObject` mappers can be used with the `Query#mapping` function to map `Row` to an `@DataObject`.

aguibert:
For this sample code:
```java
// After
client.createQuery(SQL).collecting(collector).execute();
client.createPreparedQuery(SQL).collecting(collector).execute(tuple);
client.createPreparedQuery(SQL).collecting(collector).batch(tuples);
```

Does the last line ending with `batch(tuples)` actually execute a batch query? If so, I think it would be good to have "execute" somewhere in the method name in order to make it clear at a glance that the method is executing a batch query vs. just creating a batch query.

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLConnection.java
```diff
@@ -84,56 +81,25 @@ static void connect(Vertx vertx, MSSQLConnectOptions connectOptions, Handler<Asy
   @Override
   MSSQLConnection preparedQuery(String s, Handler<AsyncResult<RowSet<Row>>> handler);
 
-  /**
-   * {@inheritDoc}
-   */
-  @Fluent
-  @GenIgnore
-  @Override
-  <R> MSSQLConnection preparedQuery(String s, Collector<Row, ?, R> collector, Handler<AsyncResult<SqlResult<R>>> handler);
```
aguibert: can these changes really be ported to 3.9? Isn't this a breaking change that would need to wait until 4.0?

vietj:
@aguibert I think @jponge suggested `executeBatch`

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLPool.java
```diff
@@ -60,56 +58,25 @@ static MSSQLPool pool(Vertx vertx, MSSQLConnectOptions connectOptions, PoolOptio
   @Override
   MSSQLPool preparedQuery(String s, Handler<AsyncResult<RowSet<Row>>> handler);
```
aguibert: I think these one-shot operations should be removed also. I know these are intended as "convenience methods" but my initial impression of using this library was I found it very confusing there were so many different ways to do the same thing. IMO we should only provide the user 1 API path where possible. 
The only benefit of these one-shot commands is that it is slightly less typing:
```java
// One-shot way
pool.preparedQuery(SQL, ar -> this::processRows);

// Regular way
pool.createPreparedQuery().execute(ar -> this::processRows);
```

If we get rid of the one-shot methods, it will significantly reduce the API surface and make the overall API much quicker for users to grasp.

vertx-sql-client/src/main/java/io/vertx/sqlclient/PreparedStatement.java
```diff
@@ -0,0 +1,80 @@
...
+   * Execute the prepared query with a cursor and createStream the result. The createStream opens a cursor
+   * with a {@code fetch} size to fetch the results.
+   * <p/>
+   * Note: this requires to be in a transaction, since cursors require it.
+   *
+   * @param fetch the cursor fetch size
+   * @param args the prepared query arguments
+   * @return the createStream
+   */
+  RowStream<Row> createStream(int fetch, Tuple args);
```
aguibert: Since `cursor()` and `createStream()` are both different flavors of executing queries, it seems that they would be better suited on the `Query` interface. 

Basically there are 4 ways that a user can retrieve results from their query:
 - RowSet<Row> (default way)
 - Function<Row,U>
 - Collector
 - Cursor (prepared statement only)
 - RowStream<Row> (prepared statement only)

I think it would be good if we only had 1 way of executing a query (via `execute` on the `Query` interface) and the result type could be selected prior to execution. For example:
 - (nothing needed to get the default RowSet)
 - `Query#mapping(Function<Row,U>)`
 - `Query#collecting(Collector)`
 - `Query#streaming(int fetchSize)` (proposed)
 - `Query#withCursor()` (proposed)

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLPool.java
vietj: please see step 5.

vertx-sql-client/src/main/java/io/vertx/sqlclient/PreparedStatement.java
vietj: Cursor and stream should be kept away from `Query` as they can require transaction isolation (in PG at least) and they should be handled only from `SqlConnection` and `PreparedStatement` and are not one shot operations (i.e they span several interactions). 

Providing cursor on `Query` would make it available from a pool without acquiring a connection.

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLConnection.java
vietj: yes it is breaking but in 3.x SQL client is in tech preview mode and can break.

aguibert:
yes, we can overload with `execute(List<Tuple>)` but @jponge and I don't think we should do so. Just by looking at the method signature, it's not quite clear that `List<Tuple>` means batch execution, but making the method named `executeBatch` would make that obvious at a glance.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/QueryImpl.java
```diff
@@ -0,0 +1,119 @@
...
+      BiCommand<PreparedStatement, Boolean> abc = new BiCommand<>(new PrepareStatementCommand(sql), ps -> {
+        String msg = ps.prepare((TupleInternal) arguments);
+        if (msg != null) {
+          return Future.failedFuture(msg);
+        }
+        return Future.succeededFuture(builder.createCommand(ps, autoCommit, arguments, handler));
+      });
+      scheduler.schedule(abc, handler);
+    } else {
+      builder.execute(scheduler, sql, autoCommit, singleton, handler);
```
BillyYccc: I'm afraid users would be confused here especially those who have used client side prepared statement before, if someone writes code like this
```
conn.createQuery("SELECT * FROM users WHERE id = ?")
        .execute(Tuple.of(1));
```

They will get wrong sql syntax error but what they expect is executing a query with a param.

And even this kind of query can also be executed
```
conn.createQuery("SELECT * FROM users")
        .execute(Tuple.of(1, 2, 3));
```

vietj: it is not cleat what you mean, you are commenting on `QueryImpl` which is not the API

aguibert:
@tsegismont here is the earlier proposal (it's just another PR made by me)
https://github.com/eclipse-vertx/vertx-sql-client/pull/549

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/QueryImpl.java
BillyYccc: I mean the API allows that a simple query could be executed or batched with parameters but it's implicitly executed without any parameters

vertx-sql-client/src/main/java/io/vertx/sqlclient/PreparedStatement.java
aguibert: > Providing cursor on Query would make it available from a pool without acquiring a connection.

I was suggesting that the proposed `streaming` and `withCursor()` methods would alter the result type once `execute` is eventually invoked, so it would still require a connection. Example usage might look like:
```java
pool.createPreparedQuery(SQL)
       .withCursor()
       .execute(ar -> { 
         Cursor c = ar.result();
         // do stuff with cursor
       });
```
So this way `execute` would be the only terminal operation in the API (for executing queries). Otherwise we have 3 terminal operations (execute, cursor, and stream).

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/QueryImpl.java
aguibert: Can you clarify your concern here @BillyYccc ?

If a user just does `conn.createQuery("SELECT * FROM users WHERE id = ?")` no interaction with the DB would occur until `execute()` is invoked. So the first code example you showed will work fine.

For the second code example, users can already use incorrect SQL and params with the current API like this:
```java
// blows up because SQL has 0 params but 3 were passed in
conn.preparedQuery("SELECT * FROM users", Tuple.of(1, 2, 3));
```

vertx-sql-client/src/main/java/io/vertx/sqlclient/PreparedStatement.java
vietj: As I said, using this from a pool is error prone. A cursor is an advanced feature and the user should use a connection to interact with a Cursor and it does not fit well with the `SqlClient` interface that provides exclusively one-shot operation that do not require any close operation to release the resource.

tsegismont:
@aguibert thanks

So IIUC, the goal is to be able to reuse query objects, in the same way as request objects in the web client?

aguibert:
Sort of, that's certainly an output of this transformation, but the goal I had in mind when initially discussing this w/ Julien is to reduce the API surface area of vertx-sql-client. In its current form there are an overwhelming amount of convenience methods that made it hard for me to wrap my head around the API when I was initially approaching it. 
The catalyst for this transformation was that we will soon need to add a `QueryOptions` object, and adding this to all of the various convenience methods would have doubled the already large API surface. But now that we are switching to a fluent/builder style, we'll only need to add a `Query#withOptions(QueryOptions)` method that users may optionally call.

I think JDBC is a good example of what the finished product might look like -- JDBC is a very concise API with a pretty simple topology (Connection --> Statement --> ResultSet)

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/QueryImpl.java
BillyYccc: I think they are different in terms of a user's intuitions.

My first example shows a scene where users wants to execute a client side prepared statement which is commonly used in JDBC drivers, but we don't support this in all clients so users will get an error of wrong sql syntax, this looks strange for those users but it's OK once users know we don't support this kind of queries.

My second example shows a usecase when users accidently writes a sql of no placeholders but execute with several parameters, the query could be executed and parameters are ignored but in my opinion using simple query with parameters should be forbidden rather than executing the simple query implicitly while ignoring params if we don't support client side prepared statement. 

As for the prepared statement example, it's different because the prepared statement has been pre-compiled, when you execute the prepared statement, the params count and even types will be validated with param metadata locally so users will quickly get an error signal without executing it in reality.

aguibert: Ah, I think I see what you are getting at. For the first example, it is a syntax error because the user is doing `con.createQuery(SQL)` as opposed to `con.createPreparedQuery()`?
This does indeed seem like an issue. Since `Query` is used for both prepared and non-prepared queries.

vertx-sql-client/src/main/java/io/vertx/sqlclient/Query.java
```diff
@@ -0,0 +1,86 @@
...
+   *
+   * @param batch the batch of tuples
+   * @param handler the handler receiving the response
+   */
+  void batch(List<Tuple> batch, Handler<AsyncResult<T>> handler);
+
+  /**
+   * Like {@link #batch(List, Handler)} but returns a {@code Future} of the asynchronous result
+   */
+  Future<T> batch(List<Tuple> batch);
```
aguibert: I think we need to have a separate `PreparedQuery` interface, because the methods that accept a `Tuple` argument are only for prepared queries.
Currently both `createQuery()` and `createPreparedQuery()` return type `Query`. As @BillyYccc pointed out a user could try to send in params to a non-prepared query, which is invalid:
```java
// Error: cannot set params on a non-prepared query
client.createQuery(SQL).execute(Tuple.of("foo"));
```

vietj: I think it's fine like that and we do not need to over engineer this

vietj: My point is that the `PreparedQuery` would be very similar to `Query` with only the `execute` method that changes and we would have two interfaces expressing the same concern for a very small gain.

aguibert: What is the opposition to having a separate `PreparedQuery` interface that extends the base `Query` interface? It seems like a good improvement to make the API more type-safe. We could catch these errors at compile-time as opposed to blowing up at runtime.

vietj: that remains a valid point though, perhaps we could have `PreparedQuery` that extends `Query` and add `execute(Tuple)` and `executeBatch(List<Tuple>)`

jponge: Agreed with @vietj here, and that would make sense to specialize `Query` here.

vertx-db2-client/src/main/java/examples/SqlClientExamples.java
```diff
@@ -151,8 +151,8 @@ public void usingConnections01(Vertx vertx, Pool pool) {
   public void usingConnections02(SqlConnection connection) {
     connection.prepare("SELECT * FROM users WHERE first_name LIKE $1", ar1 -> {
       if (ar1.succeeded()) {
-        PreparedQuery pq = ar1.result();
-        pq.execute(Tuple.of("andy"), ar2 -> {
+        PreparedStatement pq = ar1.result();
+        pq.query().execute(Tuple.of("andy"), ar2 -> {
```
jponge: I like the statement / query / {collector}? / execution API flow.

vietj:
I'll push an update tomorrow with the feedback

vertx-sql-client/src/main/java/io/vertx/sqlclient/Query.java
aguibert: yes, that's essentially what I was suggesting. `PreparedQuery` would extend `Query` and `PreparedQuery` would add the Tuple-overloaded execute methods. 

Sounds like we are on the same page now 👍

vietj:
I pushed the updates, most tests using query/preparedQuery methods had to be updated which is a lot of files.

vietj:
This PR is no longer in draft mode and I would like to get it merged and back-ported to 3.9 - I will start the port based on this current PR

tsegismont:
@aguibert thanks for clarifying!

The rationale sounds good to me, and I like the side effect of being able to reuse the `Query` object. But that means the collector instances can be reused and used concurrently. Do we actually have that guarantee?

vietj:
@tsegismont collectors are factories for producing objects that will own the state

tsegismont:
Sorry @vietj not sure I understand. Let's take this example:

```java
query = client.createQuery(SQL).collecting(Collectors.toList());
```

The query object will be reusable, right? And the collector instance created once. What do you mean with _collectors are factories_? Does the SDK mandate that `Collector` implementations can be reused or used concurrently?

vietj:
Collector instance can be shared as far as I can see how we use them for collecting queries.

Queries do not own any state beside configuration state, i.e once a query is created then execute can be called several times. The query creator should respect how the query was obtained:

- if the query is obtained from a pool, then the context does not matter, the query will call back on the context
- if the query is obtained from a connection, then the callback will likely be done on the connection context

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/QueryBase.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @Override
+  public <U> Query<SqlResult<U>> collecting(Collector<Row, ?, U> collector) {
+    return copy(new SqlResultBuilder<>(SqlResultImpl::new, collector));
+  }
+
+  @Override
+  public <U> Query<RowSet<U>> mapping(Function<Row, U> mapper) {
+    return copy(new SqlResultBuilder<>(RowSetImpl.factory(), RowSetImpl.collector(mapper)));
+  }
```
BillyYccc: we could add null check for these methods

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlClientBase.java
```diff
@@ -175,7 +61,7 @@ public C preparedBatch(String sql, List<Tuple> batch, Handler<AsyncResult<RowSet
     Function<R1, R2> factory,
     Collector<Row, ?, R1> collector,
     Promise<R3> handler) {
```
BillyYccc: this method could be removed

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlClient.java
```diff
@@ -17,15 +17,9 @@
...
 package io.vertx.sqlclient;
 
-import io.vertx.codegen.annotations.Fluent;
-import io.vertx.codegen.annotations.GenIgnore;
 import io.vertx.codegen.annotations.VertxGen;
-import io.vertx.core.AsyncResult;
-import io.vertx.core.Future;
-import io.vertx.core.Handler;
 
 import java.util.List;
```
BillyYccc: useless import

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlClient.java
```diff
@@ -36,136 +30,21 @@
...
-  SqlClient preparedQuery(String sql, Handler<AsyncResult<RowSet<Row>>> handler);
-
-  /**
-   * Like {@link #preparedQuery(String, Handler)} but returns a {@code Future} of the asynchronous result
-   */
-  Future<RowSet<Row>> preparedQuery(String sql);
-
-  /**
-   * Execute the given {@code sql} string using a prepared statement without any parameter, the asynchronous result is represented as a collection of elements transformed by the provided {@link java.util.stream.Collector}.
+   * Create a query.
```
BillyYccc: duplicated documentation

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlClient.java
```diff
@@ -36,136 +30,21 @@
...
-   * Like {@link #preparedQuery(String, Collector, Handler)} but returns a {@code Future} of the asynchronous result
+   * @return the query
    */
-  @GenIgnore
-  <R> Future<SqlResult<R>> preparedQuery(String sql, Collector<Row, ?, R> collector);
+  Query<RowSet<Row>> query(String sql);
 
   /**
-   * Execute the given {@code sql} string using a prepared statement with parameters set in the {@code Tuple}, the asynchronous result is represented as a {@link RowSet}.
+   * Create a prepared query, one of the {@link PreparedQuery#execute}, {@link PreparedQuery#executeBatch}
```
BillyYccc: the created object is actually not a **prepared** query, how about writing as 
> Create a prepared query request, ...

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlClient.java
```diff
@@ -36,136 +30,21 @@
 public interface SqlClient {
 
   /**
-   * Execute a simple query using the given {@code sql} string, the asynchronous result is represented as a {@link RowSet}.
+   * Create a query, the {@link Query#execute} method must be called to execute the query.
```
BillyYccc: how about writing as 
> Create a query request, ...

vietj:
Here is the 3.9 backport https://github.com/eclipse-vertx/vertx-sql-client/pull/559

vietj:
If that's fine I'd like to merge this branch in master and then merge the corresponding PR for 3.9

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlClient.java
vietj: I'm not sure to understand here

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlClient.java
vietj: I don't understand

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlClient.java
BillyYccc: I mean we could clarify in the documentation that this created `PreparedQuery` object is not prepared indeed and is only a request handle for one-shot executing a prepared query.

vietj: I think it does not matter because it is the same from the user perspective as he will get a single callback that encompass both query validation and query execution which is what makes the difference and the doc says that `execute` must be called to execute the query

jponge:
LGTM

aguibert:
was out on vacation today so I couldn't give final comments, but everything looks great now -- thanks for incorporating the feedback @vietj!

vietj:
@aguibert I put some pressure on this because I wanted a back-port to 3.9 for release soon

PR #562, opened 2020-03-31
 Get all DB2 tests passing with DB2 on Z

### LINKED ISSUE #518: Support for DB2 z/OS
Currently the DB2 module is being developed using DB2 LUW (Linux, Unix, Windows). Ensure that all of the DB2 client operations also work using a DB2 z/OS backend (to the extent allowed by the backend)

aguibert:
Closes #518

### DIFF
--- a/vertx-db2-client/src/test/java/io/vertx/db2client/TableJoinTest.java
+++ b/vertx-db2-client/src/test/java/io/vertx/db2client/TableJoinTest.java
@@ -1,10 +1,13 @@
 package io.vertx.db2client;
 
+import static org.junit.Assume.assumeFalse;
+
 import java.util.Arrays;
 
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
+import io.vertx.db2client.junit.DB2Resource;
 import io.vertx.ext.unit.TestContext;
 import io.vertx.ext.unit.junit.VertxUnitRunner;
 import io.vertx.sqlclient.Row;
@@ -76,6 +79,7 @@ public class TableJoinTest extends DB2TestBase {
 
 	@Test
 	public void testFullOuterJoin(TestContext ctx) {
+	    assumeFalse("DB2 on Z does not support operations within ON clause for FULL OUTER JOIN", rule.isZOS());
 		testJoin(ctx, "FULL OUTER JOIN");
 	}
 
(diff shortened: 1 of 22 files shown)


aguibert:
Also closes #499 with the 3rd commit

vertx-db2-client/src/test/java/io/vertx/db2client/tck/DB2TransactionTest.java
```diff
@@ -62,5 +65,11 @@ protected void cleanTestTable(TestContext ctx) {
   protected String statement(String... parts) {
     return String.join("?", parts);
   }
+  
+  @Test
+  public void testDelayedCommit(TestContext ctx) {
+    assumeFalse("DB2 on Z holds write locks on inserted columns with isolation level = 2", rule.isZOS());
```
gjwatts: Why fail the test for Z instead of wrapping it with 
if (!rule.isZOS()) {
  super.testDelayedCommit(ctx);
}

vertx-db2-client/src/test/java/io/vertx/db2client/TableJoinTest.java
```diff
@@ -76,6 +79,7 @@ public void testRightOuterJoin(TestContext ctx) {
 
 	@Test
 	public void testFullOuterJoin(TestContext ctx) {
+	    assumeFalse("DB2 on Z does not support operations within ON clause for FULL OUTER JOIN", rule.isZOS());
```
gjwatts: Why fail the test for Z instead of wrapping it with 
if (!rule.isZOS()) {
  testJoin(ctx, "FULL OUTER JOIN");
}

vertx-db2-client/src/test/java/io/vertx/db2client/tck/DB2TransactionTest.java
aguibert: Here I am using `assumeFalse` which is slightly different than `assertFalse`. The `assumeFalse` method will simply bail out of the test if the assumption is not met, which shows up as a SKIP in the JUnit reports. So it's effectively the same thing as wrapping the test in the `if` statement you suggested, except the test will show up as SKIPPED instead of PASSED in the report.

PR #564, opened 2020-03-31
Avoid using a beta release and use the latest stable

pmlopes:
Signed-off-by: Paulo Lopes <pmlopes@gmail.com>

Motivation:

The current scram optional dependency is a beta release. Since its release a official production release has beed added and more after that that have been properly tested on the jdbc postgres driver.


vietj:
This dependency is good to go

PR #570, opened 2020-04-02
Initial experimentation with DB2 pipelining

aguibert:
Changes in this PR:
 - Added pipelining configurability. Currently it defaults to 1 (pipelining is disabled) which is no behavior change. Doing experimentation locally with raising this but it's still very unstable.
 - Renamed DatabaseMetaData to ConnectionMetaData and moved from the SocketConnection to DB2Encoder level
 - Added handling for OPNQRYFLM and SYNTAXRM code points, which only occur on error paths (Supports #490)


aguibert:
There was one unrelated failure in one of the PG tests, but other than that the Travis build is good

PR #573, opened 2020-04-03
Utilize sections more efficiently

aguibert:
In DB2 there are "packages" which may be small or large in size, containing 64 or 384 "sections" respectively. By default DB2 LUW has 3 small and 3 large packages, and DB2/z has 3 large packages.

Each section can only be used by one query (i.e. SELECT statement) at a time, per connection.

This PR does the following optimizations:
 - Always reuses the same special immediate section (sectionNum=MAX_SECTIONS+1) for non-query statements
 - Moves the SectionManager to a per-connection scope instead of a global singleton


gjwatts:
In the issue description, it says _By default DB2 LUW has 3 small and 3 large "sections", and DB2/z has 3 large "sections"._  Did you mean to say By default DB2 LUW has 3 small and 3 large "**packages**", and DB2/z has 3 large "**packages**".

aguibert:
Yes you're correct @gjwatts, it should say "packages" instead of "sections"

PR #574, opened 2020-04-06
Simplified transaction API improvements

vietj:
This PR attempts to provide a better simplified transaction API.

The current API manages the connection on behalf of the user avoiding the tedious return to a pool:

```java
pool.begin()
  .flatMap(tx -> tx
    .query("INSERT INTO Users (first_name,last_name) VALUES ('Julien','Viet')")
    .execute()
    .flatMap(res1 -> tx
      .query("INSERT INTO Users (first_name,last_name) VALUES ('Emad','Alblueshi')")
      .execute())
    .compose(
      res2 -> tx.commit(), 
      err -> tx.rollback())
  ).onComplete(ar3 -> {
  if (ar3.succeeded()) {
    System.out.println("Transaction succeeded");
  } else {
    System.out.println("Transaction failed " + ar3.cause().getMessage());
  }
});
```

The new API relies on the user providing a function that maps a client to a `Future<T>` on which depends the overall result and the transaction management:

- when the future succeeds the transaction is committed
- when the future fails the transaction is rolled back

The overall result function result is returned as completion of the overall operation

```java
// Acquire a transaction and begin the transaction
pool.withTransaction(client -> client
  .query("INSERT INTO Users (first_name,last_name) VALUES ('Julien','Viet')")
  .execute()
  .flatMap(res -> client
    .query("INSERT INTO Users (first_name,last_name) VALUES ('Julien','Viet')")
    .execute()
    // Map to a message result
    .map("Users inserted"))
).onComplete(ar -> {
  // The connection was automatically return to the pool
  if (ar.succeeded()) {
    // Transaction was committed
    String message = ar.result();
    System.out.println("Transaction succeeded: " + message);
  } else {
    // Transaction was rolled back
    System.out.println("Transaction failed " + ar.cause().getMessage());
  }
});
```

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
@@ -16,7 +16,6 @@
  */
 package io.vertx.sqlclient;
 
-import io.vertx.codegen.annotations.Fluent;
 import io.vertx.codegen.annotations.VertxGen;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
@@ -47,14 +46,6 @@ public interface Transaction {
    * Like {@link #rollback} with an handler to be notified when the transaction rollback has completed
    */
   void rollback(Handler<AsyncResult<Void>> handler);
-
-  /**
-   * Set an handler to be called when the transaction is aborted.
-   *
-   * @param handler the handler
-   */
-  @Fluent
-  Transaction abortHandler(Handler<Void> handler);
 
   /**
    * Rollback the transaction and release the associated resources.
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
@@ -26,7 +26,7 @@ import io.vertx.core.Handler;
  * A transaction that allows to control the transaction and receive events.
  */
 @VertxGen
-public interface Transaction extends SqlClient {
+public interface Transaction {
 
   /**
    * Create a prepared query.
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
@@ -22,7 +22,7 @@ import io.vertx.core.Future;
 import io.vertx.core.Handler;
 
 /**
- * A transaction that allows to control the transaction and receive events.
+ * A transaction.
  */
 @VertxGen
 public interface Transaction {
@@ -53,7 +53,7 @@ public interface Transaction {
   void close();
 
   /**
-   * Return the transaction result, the returned future
+   * Return the transaction completion {@code Future} that
    *
    * <ul>
    *   <li>succeeds when the transaction commits</li>
@@ -62,6 +62,6 @@ public interface Transaction {
    *
    * @return the transaction result
    */
-  Future<Void> result();
+  Future<Void> completion();
 
 }
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
@@ -74,4 +74,17 @@ public interface Transaction extends SqlClient {
    * Rollback the transaction and release the associated resources.
    */
   void close();
+
+  /**
+   * Return the transaction result, the returned future
+   *
+   * <ul>
+   *   <li>succeeds when the transaction commits</li>
+   *   <li>fails with {@link TransactionRollbackException} when the transaction rollbacks</li>
+   * </ul>
+   *
+   * @return the transaction result
+   */
+  Future<Void> result();
+
 }
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
@@ -29,20 +29,6 @@ import io.vertx.core.Handler;
 public interface Transaction {
 
   /**
-   * Create a prepared query.
-   *
-   * @param sql the sql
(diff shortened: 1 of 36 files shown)


knotenpunkt:
im not sure if im understand this here correct, but it seems that you as user dont have any more the controll, to abbort the transaction?

you start the transaction 
....
now you take a look on the clock or something else (extrinsic) and you come to the end, that you dont want to commit the transaction. At this point you want to cancel it.

If my assumption is wrong, than you can ignore this comment, and delete it^^
Maybe i have the assumption, cause im not familiar with rxJava (flatMap looks like that^^)

And Julien, in the meantime you were a few times online in gitter. Not opening...... its not better.
Take a look on it, there you have a big answer.
And you should take a look on my issues in the vertx repositories. A few of them are in deed a feature-requests. **But the two with the "bug-word" in the title, are bugs and you should really take a look on it.**

vietj:
the transaction is rolled back if the future returned by the function fails

knotenpunkt:
but do i have the control over the future? can i on myself fail it?
Thats the question^^

Of course if something in the database-connection itself crashes, then it fails. But if i have 10 queries/updates in the transaction, can i say after update 7 that i want to fail, can i say after update 10/10 that i want to fail.

If yes then all is ok^^

vietj:
there is always possibility to re-map a future to a failure

knotenpunkt:
ah ok, thx, yes that was the thing i wanted to know^^
yes so it seems to be a good draft.

jponge:
It indeed simplifies transactions.

vietj:
more feedback before pushing a PR ?

vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
```diff
@@ -26,7 +26,7 @@
  * A transaction that allows to control the transaction and receive events.
  */
 @VertxGen
-public interface Transaction extends SqlClient {
```
BillyYccc, line 29: Can you clarify is it necessary that `Transaction` does not extend `SqlClient` anymore? I think it becomes more complicated since users have to use the connection API directly if doing some operations on a transaction.

vietj: I removed it for several reasons:

- I wanted to have the ability to have `Transaction` extends `Future<Void>` instead of having a `result()` field
- we might add in the future the `withTransaction` method to `SqlClient` and in this case it does not make sense to have `withTransaction` on `Transaction` in this situation

BillyYccc: how about duplicating the two interfaces `query` and `preparedQuery` on `Transaction` so that we can use `Transaction` instread of using the `connection` reference to schedule a command.

vietj: I believe `Transaction` should remain only there for managing transactions.

vietj: in other words, having `Transaction` extending `SqlClient` was a design mistake I did and I would like to fix it

BillyYccc: ok, it seems that we could also remove `Transaction#prepare` in this interface?

vietj: yes totally, I think I forgot that and I will remove it, thanks for noticing

vietj: I will also likely remove the `abortHandler` as it is now replaced with the fact that the transaction has a Future<Void> that gives the overral transaction result and thus `onComplete` can be called instead.

vietj:
I improved the PR:

- Transaction result() renamed to completion()
- removed leftover methods from SqlClient inheritance (prepared and close methods)
- removed abort handler that does the same than the completion future

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/TransactionImpl.java
```diff
@@ -209,24 +218,12 @@ public void rollback(Handler<AsyncResult<Void>> handler) {
...
+  private <R> ScheduledCommand<R> doQuery(TxCommand<R> cmd, Promise<R> handler) {
     return new ScheduledCommand<>(cmd, handler);
   }
-  
+
   @Override
-  boolean autoCommit() {
-    return false;
+  public Future<Void> completion() {
+    return completionFuture;
```
BillyYccc, line 227: the field could be removed and use `completionPromise.future()` instead.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/TransactionImpl.java
```diff
@@ -58,21 +64,61 @@
...
+  private <R> void wrapAndSchedule(ScheduledCommand<R> scheduled) {
+    CommandBase<R> cmd = scheduled.cmd;
+    if (isComplete(cmd)) {
+      status = ST_COMPLETED;
+      doSchedule(cmd, ar -> {
+        if (ar.succeeded()) {
+          if (cmd == COMMIT) {
+            completionPromise.tryComplete();
+          } else {
+            completionPromise.tryFail(TransactionRollbackException.INSTANCE);
```
BillyYccc, line 87: we can let the `TransactionRollbackException` include the failure context here, like `new TransactionRollbackException(ar.cause)`.

BillyYccc: my bad, it's not possible here

BillyYccc:
will this feature get backported to `3.9` branch as well?

vietj:
no because it does not provide a future oriented programming model

vietj:
we could backport the Transaction interface changes however

PR #575, opened 2020-04-06
Prepared Statement cache rework

### LINKED ISSUE #572: Closed prepared statements are not evicted from the preparedstatement cache correctly
### Version

I think the problem should be both in `3.x` and `master` branches.

### Context

Closed Prepared statements by `PreparedStament#close()` are not evicted from the preparedstatement cache correctly

### Do you have a reproducer?

It's not hard to create a reproducer.
```
        mySQLClient.getConnection(ar1 -> {
            if (ar1.succeeded()) {

                connection.prepare("SELECT * FROM test", ar2 -> {
                    if (ar2.succeeded()) {
                        PreparedStatement preparedStatement = ar2.result();
                        preparedStatement.query()
                                .execute(ar3 -> {
                                    if (ar3.succeeded()) {
                                        RowSet<Row> rowSet = ar3.result();
                                        System.out.println(rowSet.size());

                                        preparedStatement.close();
                                        try {
                                            Thread.sleep(500);
                                            e.printStackTrace();
                                        }

                                        connection.prepare("SELECT * FROM test", ar4 -> {
                                            if (ar4.succeeded()) {
                                                PreparedStatement statement2 = ar4.result();
                                                statement2.query()
                                                        .execute(ar5 -> {
                                                            if (ar5.succeeded()) {
                                                                RowSet<Row> rowSet2 = ar5.result();
                                                                System.out.println(rowSet2.size());
                                                            } else {
                                                                ar5.cause().printStackTrace();
                                                            }
                                                        });
                                            } else {
                                                ar4.cause().printStackTrace();
                                            }
                                        });
                                    } else {
                                        ar3.cause().printStackTrace();
                                    }
                                });

                    } else {
                        ar2.cause().printStackTrace();
                    }
                });
            } else {
                ar1.cause().printStackTrace();
            }
        });
```

BillyYccc:
Motivation:

Rework the prepared statement cache. 

* remove the stmt cached result from the cache when it's closed by `PreparedStatement#close()`,
so when a stmt is evicted automatically by the cache or closed by the user, the entry will both be removed from the cache.  - fix #572 
* In the past implementation, the cache is tied to the lifecycle of prepared statements, so a new class `PreparedStatementCacheManager` is introduced here to seperate the concerns. 
* remove the `statement` from `PrepareStatementCommand` in SPI, it's Postgres specific and I think the handling should be moved into pg-client.

Further thinking(not included in this PR):
And I propose that we don't cache preparedStatement created by `SqlConnection#prepare`, as the lifecycle should be managed by the user instead of being potentially closed by the cache. We can add an `isOneShot` flag in `PrepareStatementCommand` to indicate whether the prepare command request is from `PreparedStatement#prepare` or `SqlClient#preparedQuery`.

It's also useful in Postgres client to decide whether to create a named prepared statement of an unnamed one.
before changes:
`SqlConnection#prepare` - create unnamed prepared statements
`SqlClient#preparedQuery` - create unnamed prepared statements
after changes:
`SqlConnection#prepare` - create named prepared statements
`SqlClient#preparedQuery` - create unnamed prepared statements

Feel free to give your opinions on this and I will get it backported to `3.9` once we agree on this.


vietj:
can you remove the converter generated code from the PR to make it easier to review ?

BillyYccc:
ok I have dropped that commit just now.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/cache/PreparedStatementCacheManager.java
```diff
@@ -0,0 +1,92 @@
...
+import io.vertx.sqlclient.impl.SocketConnectionBase;
+import io.vertx.sqlclient.impl.command.CloseStatementCommand;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Cache manager which manages the lifecycle of all cached prepared statements .
+ */
+public class PreparedStatementCacheManager {
```
vietj, line 27: why does this needs to be renamed from ` PreparedStatementCache` to ` PreparedStatementCacheManager` ?

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/PgEncoder.java
```diff
@@ -179,6 +181,20 @@ void writeClosePortal(String portal) {
     out.setInt(pos + 1, out.writerIndex() - pos - 1);
   }
 
+  void writeClosePreparedStatement(long statementName) {
```
vietj, line 184: I don't see this code called in the PR, am I missing something ?

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/cache/PreparedStatementCacheManager.java
BillyYccc: It includes a `Map` recording inflight requests and manages the real cache which is used for caching `AsyncResult` of the prepare command responses. The `PreparedStatementResultLruCache` is a part of this and could also be pluggable in the future.

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/PgEncoder.java
BillyYccc: This is used for closing Postgres named prepared statements, we could add support for that in another PR as stated in the proposal.

It's not useful in this PR, I will remove this.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/cache/PreparedStatementCacheManager.java
vietj: can it be called cache like before and the map then lruMap ? I don't think we should over engineer this for now and see later in the future whether the needs arise in practice

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/PgEncoder.java
vietj: yes I remember this part and in reality it should be called when one is calling `PreparedStatement.close()` method.

can we have this case tested and implemented with `io.vertx.pgclient.impl.codec.CloseStatementCommandCodec` ? I think it used to work in the former Reactive PG Client.

vietj: in other words calling `PreparedStatement#close()` when caching is disabled should actually release this resource in PostgreSQL and this code should be called.

BillyYccc:
updated the PR and create a new related issue in https://github.com/eclipse-vertx/vertx-sql-client/issues/577.

BillyYccc:
Are we good to go now? I will then backport this into `3.9`.

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/PrepareStatementCommandCodec.java
```diff
@@ -30,8 +30,8 @@
 
   @Override
   void encode(PgEncoder encoder) {
-    encoder.writeParse(new Parse(cmd.sql(), cmd.statement()));
-    encoder.writeDescribe(new Describe(cmd.statement(), null));
+    encoder.writeParse(new Parse(cmd.sql(), 0L));
```
vietj, line 33: why are we replacing `cmd.statement()` with `0L` ? if we do that it means we are actually using the un named statement. I don't get this change

BillyYccc: Because currently the prepared statement closing won't do anything on the wire, see https://github.com/eclipse-vertx/vertx-sql-client/blob/e8893a2680c100d39524901dce2e881ce9c124a2/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/CloseStatementCommandCodec.java#L28-L39

so named prepared statement won't be closed and will cause leaks. So let's make it unnamed statements now and as I said in https://github.com/eclipse-vertx/vertx-sql-client/issues/577 I will support named prepared statement for Postgres in `SqlConnection#prepare`, which could be included in another PR.

vietj: let's complete first the other PR and then we come back to this one

BillyYccc: opened a PR in https://github.com/eclipse-vertx/vertx-sql-client/pull/578

BillyYccc:
closing this and reopen a new clean PR for #572 in https://github.com/eclipse-vertx/vertx-sql-client/pull/597.

PR #578, opened 2020-04-10
Prepared statement created by SqlConnection#prepare should not be cached 

### LINKED ISSUE #577: Prepared statement created by SqlConnection#prepare should not be cached
#### Describe the feature

In the past all prepare statements are cached in the session once created, it's tricky that a user-created object `PreparedStatement` is closed by the cache automatically when the cache is enabled and the entry is evicted.

We should not cache those statements which are not one-shot prepared queries, as the lifecyles of the statments should be managed by the users themselves.

Related changes:
* add a flag in the SPI `PrepareStatementCommand` to indicate whether the prepare command request is from PreparedStatement#prepare or SqlClient#preparedQuery so we know whether the statement should be cached or not

* support named prepared statements for Postgres in `SqlConnection#prepare`, the named statements should also be closed when calling `PreparedStatement#close` by sending a `Close` command to the server.

BillyYccc:
Motivation:

fixes #577 

Conformance:

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/command/PrepareStatementCommand.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/command/PrepareStatementCommand.java
@@ -22,18 +22,25 @@ import io.vertx.sqlclient.impl.PreparedStatement;
 public class PrepareStatementCommand extends CommandBase<PreparedStatement> {
 
   private final String sql;
-  public long statement; // 0 means unamed statement otherwise CString
+  private final boolean auto;
 
-  public PrepareStatementCommand(String sql) {
+  public PrepareStatementCommand(String sql, boolean auto) {
     this.sql = sql;
+    this.auto = auto;
   }
 
   public String sql() {
     return sql;
   }
 
-  public long statement() {
-    return statement;
+  /**
+   * Indicate whether this command is scheduled from {@link io.vertx.sqlclient.SqlClient#preparedQuery(String) one-shot preparedQuery} or {@link io.vertx.sqlclient.SqlConnection#prepare(String)}.
+   * This flag will tell if lifecycle of the statement will be controlled by the client or not.
+   *
+   * @return true if the command is a one-shot prepared query.
+   */
+  public boolean auto() {
+    return auto;
   }
 
 }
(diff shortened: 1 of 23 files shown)


BillyYccc:
this will be backported to `3.9` after being merged

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/command/PrepareStatementCommand.java
```diff
@@ -22,18 +22,25 @@
...
 
-  public long statement() {
-    return statement;
+  /**
+   * Indicate whether this command is scheduled from {@link io.vertx.sqlclient.SqlClient#preparedQuery(String) one-shot preparedQuery} or {@link io.vertx.sqlclient.SqlConnection#prepare(String)}.
+   * This flag will tell if lifecycle of the statement will be controlled by the client or not.
+   *
+   * @return true if the command is a one-shot prepared query.
+   */
+  public boolean auto() {
```
vietj, line 42: I think this should not be public because it is not needed by back-ends, this is an package protected internal data that is relevant only to the API and `SocketConnectionBase`, also the name should be more explicit like `shared` or `cacheable` that will tell whether it the prepared statement result can be shared or not.

BillyYccc: the rename is updated in https://github.com/eclipse-vertx/vertx-sql-client/pull/578/commits/9408a3b5b127a37cc74af466c55ce26d1d07ad89, changing the `public` access modifier will not be possible because they're in different packages.

vietj:
I think we need a specific test to check that the prepared statement is effectively closed.

It seems that once a prepared statement has been created we can check it exists with https://www.postgresql.org/docs/9.2/view-pg-prepared-statements.html .

So we could prepare a statement then check it exists in `pg_prepared_statements` then close the prepared statement and check that `pg_prepared_statements` is empty.

BillyYccc:
yes I have added a test in https://github.com/eclipse-vertx/vertx-sql-client/pull/578/commits/304dfe0f1b1bc313f3b5b9f0c08ef9d23c7e4f7a.

BillyYccc:
The CI fails because of timeout tests, will get this merged if there's no more change requested.

vietj:
thanks for adding the tests, with the specific PG statement table I feel we cover correctly this feature

vietj:
LGTM

BillyYccc:
Thanks for the reviews!

PR #580, opened 2020-04-10
Add handling for ACCSEC error path

aguibert:


vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/DRDAConnectResponse.java
```diff
@@ -1154,6 +1154,191 @@ private void parseACCSECreply(int securityMechanism) {
...
+            parseTYPDEFOVR();
+        } else {
+            parseTYPDEFOVR();
+            parseTYPDEFNAM();
+        }
+
+        NetSqlca netSqlca = parseSQLCARD(null);
+        
+        //Check if the SQLCARD has null SQLException
+        if(netSqlca.getSqlErrmc() == null) {
```
gjwatts, line 1261: Are we guaranteed a non-null value for netSqlca before we call any methods?  I see in parseSQLCAGRP there is the possibility of a null value being returned.  Or is it ok if we throw a NPE at this point?

vertx-db2-client/src/test/java/io/vertx/db2client/tck/DB2ConnectionTest.java
```diff
@@ -1,10 +1,14 @@
 package io.vertx.db2client.tck;
 
 import io.vertx.sqlclient.tck.ConnectionTestBase;
+import io.vertx.db2client.DB2Exception;
```
gjwatts, line 4: Please add a copyright header.

vertx-db2-client/src/main/java/io/vertx/db2client/impl/codec/InitialHandshakeCommandCodec.java
```diff
@@ -54,7 +56,8 @@ void encode(DB2Encoder encoder) {
             //Sometimes DB2 closes the connection when sending an invalid Database name.
             //-4499 = A fatal error occurred that resulted in a disconnect from the data source.
             //08001 = "The connection was unable to be established"
-            cmd.fail(new DB2Exception("Socket closed during connection", -4499,"08001"));           
+            cmd.fail(new DB2Exception("The connection was closed by the database server.", 
+            		SqlCode.CONNECTION_REFUSED, SQLState.CONNECT_UNABLE_TO_CONNECT_TO_SERVER));
```
gjwatts, line 60: So here it sounds like we are able to hit the DB2 server, but it closes the connection because of an invalid DB name.  In the comments section of SQLState.java (around line 1737) they make a note that the 08001 errors are for when the client is unable to establish a connection with the server.  Here it seems we are able to connect to the server and they are rejecting the request because of the invalid DB name.  So my suggestion would be to use SQLState.NET_DATABASE_NOT_FOUND instead (which sends back value of "08004*")

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/DRDAConnectResponse.java
aguibert: A non-null SQLCARD should always be returned if a normal DB error occurs (which excludes situations where the DB just closes the connection). On this code path we're still under "normal" error path situations since we only get to this point if we have previously found and matched a `RDBAFLRM` message which must be followed by a SQLCARD

aguibert:
Part of #490

PR #591, opened 2020-04-13
Fixes for initial compatibility with Hibernate RX

aguibert:


vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java
```diff
@@ -468,6 +469,12 @@ default Buffer getBuffer(int pos) {
     if (val instanceof Buffer) {
       return (Buffer) val;
     }
+    if (val instanceof ByteBuf) {
+      return Buffer.buffer((ByteBuf) val);
+    }
+    if (val instanceof byte[]) {
+      return Buffer.buffer((byte[]) val);
+    }
```
aguibert, line 477: CC @vietj for this change. Do you think this is OK? From experimenting with the API returning null in these getters causes very confusing behavior, so I think it's good to be as lenient as possible here.

vertx-mssql-client/src/test/java/io/vertx/mssqlclient/tck/MSSQLBinaryDataTypeEncodeTest.java
```diff
@@ -57,5 +57,13 @@ public void testDecimal(TestContext ctx) {
...
     super.testDecimal(ctx);
   }
+  
+  @Test
+  @Ignore // TODO: Investigate this issue and enable the test 
+  // (unfortunately it wasn't as simple as removing the NOT NULL constraints in init.sql)
+  @Override
+  public void testNullValues(TestContext ctx) {
+    super.testNullValues(ctx);
+  }
```
aguibert, line 67: CC @BillyYccc for this change. I'm adding `testNullValues` to the common TCK in this PR, but it fails for MSSQL. Can you please investigate this when you have time?

BillyYccc: I think it's not supported in the client for now.

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/DRDAQueryRequest.java
```diff
@@ -1193,11 +1199,17 @@ private Hashtable computeProtocolTypesAndLengths(
...
-                    byte[] ba = (byte[]) inputRow[i];
-                    if (ba == null) {
+//                    byte[] ba = (byte[]) inputRow[i];
+                	int length = -1;
+                	if (inputRow[i] instanceof byte[]) {
+                		length = ((byte[]) inputRow[i]).length;
+                	} else if (inputRow[i] instanceof ByteBuf) {
+                		length = ((ByteBuf) inputRow[i]).readableBytes();
+                	}
+                    if (inputRow[i] == null) {
```
gjwatts, line 1209: If inputRow[i] can be null, should that be checked before lines 1204-107 where it could cause a NPE?  Or is the case where it would be null mutually exclusive from when it would be cast as a byte[] or ByteBuf?

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/DRDAQueryRequest.java
```diff
@@ -1215,11 +1227,17 @@ private Hashtable computeProtocolTypesAndLengths(
...
-                    ba = (byte[]) inputRow[i];
-                    if (ba == null) {
+//                    ba = (byte[]) inputRow[i];
+                	length = -1;
+                	if (inputRow[i] instanceof byte[]) {
+                		length = ((byte[]) inputRow[i]).length;
+                	} else if (inputRow[i] instanceof ByteBuf) {
+                		length = ((ByteBuf) inputRow[i]).readableBytes();
+                	}
+                    if (inputRow[i] == null) {
```
gjwatts, line 1237: Same as above

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/DRDAQueryRequest.java
aguibert: good question, I had to double check on this and the statement `if (null instanceof byte[])` evaluates to `false` so we don't need the extra null check

gavinking:
Thanks!!

PR #592, opened 2020-04-15
Implement proper handling and messages for connection error paths

gjwatts:
Change the default messaging behavior for handling trivial error paths like invalid credentials.

### DIFF
--- a/vertx-db2-client/src/test/java/io/vertx/db2client/DB2ErrorMessageTest.java
+++ b/vertx-db2-client/src/test/java/io/vertx/db2client/DB2ErrorMessageTest.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright (C) 2020 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.vertx.db2client;
+
+import org.junit.Ignore;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import io.vertx.db2client.impl.drda.SQLState;
+import io.vertx.db2client.impl.drda.SqlCode;
+import io.vertx.ext.unit.TestContext;
+import io.vertx.ext.unit.junit.VertxUnitRunner;
+
+@RunWith(VertxUnitRunner.class)
+public class DB2ErrorMessageTest extends DB2TestBase {
+	
+	@Test
+	@Ignore		// TODO - Need to figure out why this is blowing up in parseACCSECreply()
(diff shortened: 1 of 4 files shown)


vertx-db2-client/src/test/java/io/vertx/db2client/DB2ErrorMessageTest.java
```diff
@@ -0,0 +1,91 @@
...
+	}
+	
+	@Test
+	public void testConnectInvalidUsername(TestContext ctx) {
+	  options.setUser("INVALID_USER_FOR_TESTING");
+	  DB2Connection.connect(vertx, options, ctx.asyncAssertFailure(err -> {
+		  ctx.assertTrue(err instanceof DB2Exception, "The error message returned is of the wrong type.  It should be a DB2Exception, but it was of type " + err.getClass().getSimpleName());
+		  DB2Exception ex = (DB2Exception) err;
+		  ctx.assertTrue(ex.getMessage().contains("Invalid credentials"), "The SQL error message returned is not correct.  It should have contained \"Invalid credentials\", but instead it said \"" + ex.getMessage() + "\"");
+		  ctx.assertTrue(ex.getErrorCode() == SqlCode.INVALID_CREDENTIALS, "The SQL error code returned is not correct.  It should have been " + SqlCode.INVALID_CREDENTIALS + ", but it was " + ex.getErrorCode());
```
aguibert, line 48: for these types of assertions instead of doing `assertTrue(expected == actual, message)` you can do `assertEquals(expected, actual, message)` and then you don't need to include the boiler plate text "should have been X but was Y" in each message, because it will be included automatically in the event of a failure, and then any message supplied (optional) will be appended to that

vertx-db2-client/src/test/java/io/vertx/db2client/DB2ErrorMessageTest.java
```diff
@@ -0,0 +1,91 @@
...
+			  ctx.assertTrue(ex.getErrorCode() == SqlCode.INVALID_CREDENTIALS, "The SQL error code returned is not correct.  It should have been " + SqlCode.INVALID_CREDENTIALS + ", but it was " + ex.getErrorCode());
+			  ctx.assertTrue(ex.getSqlState() == SQLState.NET_CONNECT_AUTH_FAILED, "The SQL state returned is not correct.  It should have been " + SQLState.NET_CONNECT_AUTH_FAILED + ", but it was " + ex.getSqlState());
+			  System.out.println(err.getMessage());
+		  }));
+	}
+	
+	@Test
+	@Ignore  // TODO - @GJW I want to validate we would always want to fail a test for a blank user name.  If so, we want to catch this *before* we connect to the DB2 server
+	public void testConnectBlankUsername(TestContext ctx) {
+	  options.setUser("");
```
aguibert, line 70: If we were to validate early in the blank username/password case, we would have it blow up immediately when `setUser/setPassword` is called, instead of waiting until the user tries to connect with the `options`

vertx-db2-client/src/test/java/io/vertx/db2client/DB2ErrorMessageTest.java
gjwatts: Thanks Andy.  I have made these changes and re-pushed the code with just my changes in it.

vertx-db2-client/src/test/java/io/vertx/db2client/DB2ErrorMessageTest.java
gjwatts: I will handle that in a future PR.

PR #597, opened 2020-04-22
Rework the prepared statement cache and evict the statement from the …

### LINKED ISSUE #572: Closed prepared statements are not evicted from the preparedstatement cache correctly

BillyYccc:
…cache when it's closed - fixes #572

Motivation:

fixes #572 based on changes after #577.

Conformance:


vietj:
so here we are again :-)

vietj:
@BillyYccc does it need to be rewieved ?

BillyYccc:
I think it would be better to at least take a quick glance at it :-)

vietj:
yes I will then

gsaqui:
@BillyYccc will this also be put into 3.9.1?

BillyYccc:
yes it will be backported to `3.9`.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/cache/LruCache.java
```diff
@@ -0,0 +1,44 @@
...
+import io.vertx.core.Handler;
+import io.vertx.sqlclient.impl.PreparedStatement;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * A LRU replacement strategy cache based on {@link java.util.LinkedHashMap} for prepared statements.
+ */
+class LruCache extends LinkedHashMap<String, AsyncResult<PreparedStatement>> {
```
vietj, line 24: why do we cache the `AsyncResult<PreparedStatement>` and not the `PreparedStatement` instead ?

BillyYccc: It behaves like it used to be. When one prepare statement request fails and the failure would be also cached.

BillyYccc: I wonder if we might want to improve this because caching failures such as syntax error, network failure, etc... does not help anything from the perspective of users.

vietj: ok, I didn't remember this part.

However then it does not map correctly to the max prepared statement cache size because this cache size should account for valid statements I think. So perhaps we want to remove failure caching.

BillyYccc: I have changed this in https://github.com/eclipse-vertx/vertx-sql-client/pull/597/commits/71d8b670e356418360a48c021993bf41b59e0fd0 now.

vietj:
I think we need also to take in account invalid cached prepared statements if we want to make this right, e.g user creates a prepared statement then somehow the underlying table is changed which makes the prepared statement invalid. The prepared statement cache should be able to recover from this.

Can you add a test that do this and see how the various implementation react and see how we could recover from this ?

BillyYccc:
I don't think we need such kind of mechanisms, because in practice it's rare to use DDL to directly break the table structure and all the existed queries would fail, normally the compability should be took care by the users.

vietj:
somebody reported that it happens for them actually (in forums I think) and PG throws a specific error code when it happens

BillyYccc:
I don't find a relevant topic in the forum, do you have a link to that?

Anyway I believe running an invalid query in the server should always be avoided, we don't have to handle that in the cache.

vietj:
here is the relevant topic : https://groups.google.com/d/msg/vertx/2EQRL2d88E0/auaATWeYBAAJ

I believe that we should handle this kind of error when it happens in postgres and remove the cached query from the cache to have it rebuilt once nexte execution

BillyYccc:
says we have a table `test` and there're already two columns `id`, `name` in the table, here are different situations for MySQL and Postgres:
|  | `SELECT * FROM test`  add a `temp` column | `SELECT * FROM test`   drop `name` column | `SELECT id, name FROM test`  add a `temp` column | `SELECT id, name FROM test`  drop `name` column |
|---------------------------------------------------------------|------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------|--------------------------------------------------|----------------------------------------------------------------------------|
| one-shot preparedQuery with cache MySQL | works fine | works fine | works fine | io.vertx.mysqlclient.MySQLException: Unknown column 'name' in 'field list' |
| preparedStatement MySQL | works fine | works fine | works fine | io.vertx.mysqlclient.MySQLException: Unknown column 'name' in 'field list' |
| one-shot preparedQuery with cache Postgres(unnamed statement) | io.vertx.pgclient.PgException: bind message has 2 result formats but query has 3 columns | io.vertx.pgclient.PgException: bind message has 3 result formats but query has 2 columns | works fine | io.vertx.pgclient.PgException: column "name" does not exist |
| preparedStatement Postgres(named statement) | io.vertx.pgclient.PgException: cached plan must not change result type | io.vertx.pgclient.PgException: cached plan must not change result type | works fine | io.vertx.pgclient.PgException: column "name" does not exist |

I find if using `SELECT *` in the Postgres, once the table is changed the preparedQuery will always fail.
I think we could capture such specific exceptions in the postgres client and sees how to handle this well.

BillyYccc:
should we handle the specific bevior in the PR or open an issue and handle it in another PR?

vietj:
if we provide one-shot queries we must provide a way to recover from these situations where the SQL statement is valid but the cached SQL statement is not valid anymore.

I think we can handle it in this PR because it is related to the cache and its invalidation.

vietj:
Note we don't want to handle this for explicitly prepared statement

BillyYccc:
@vietj can you please take a look at this?

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/InvalidCachedStatementExecutionEvent.java
```diff
@@ -0,0 +1,16 @@
+package io.vertx.sqlclient.impl;
+
+/**
+ * An event signals when an invalid cached prepared statement is executed and gets an error response.
+ */
+public class InvalidCachedStatementExecutionEvent {
+  private final PreparedStatement preparedStatement;
+
+  public InvalidCachedStatementExecutionEvent(PreparedStatement preparedStatement) {
```
vietj, line 9: For now can we move this to the `io.vertx.sqlclient.impl.codec` package and rename to `InvalidCachedStatementEvent` because such detection could happen without the execution of a statement. Instead of the `PreparedStatement` just provide the statement identifier so it can be removed from the cache.

BillyYccc:
updated in https://github.com/eclipse-vertx/vertx-sql-client/pull/597/commits/d92102055fe973c066452dda3cb7bf656d0862eb.

BillyYccc:
Merging this now and will backport it to `3.9`.

PR #598, opened 2020-04-24
Use unique PRDID of JNT

aguibert:
DB2 servers use a 3-char PRDID (product ID) string to identify which type of client they are interacting with, so that extra DB-side features can be enabled/disabled based on what certain clients support.

This PR converts to a new PRDID "JNT" (short for Java Netty client) and also makes a few minor changes in the way DRDA protocol is processed with that PRDID.


aguibert:
Merging this PR in so I can get started on backporting the DB2 module to 3.9

aguibert:
Merging this PR in so I can get started on backporting the DB2 module to 3.9

aguibert:
Merging this PR in so I can get started on backporting the DB2 module to 3.9

PR #604, opened 2020-05-06
Backport the MS SQL Client to Vert.x 3.9.x

cescoffier:
Just adapt the code from the 4.x branch (master) to work with 3.9.x.

Motivation:

There is a need to get the MS SQL client to work against 3.9. This PR is just adapting the code to follow the Vert.x 3.9 conventions.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -106,6 +106,7 @@
     <module>vertx-sql-client</module>
     <module>vertx-pg-client</module>
     <module>vertx-mysql-client</module>
+    <module>vertx-mssql-client</module>
   </modules>
 
 
--- a/vertx-mssql-client/README.adoc
+++ b/vertx-mssql-client/README.adoc
@@ -1,0 +1,11 @@
+= The Reactive MSSQL Client
+
+TODO
+
+== Features TODO
+
+* Cursor support
+* Row streaming
+* RxJava 1 and RxJava 2
+* Polyglot support
+* TLS support
--- a/vertx-mssql-client/pom.xml
+++ b/vertx-mssql-client/pom.xml
@@ -1,0 +1,112 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.vertx</groupId>
+    <artifactId>vertx-sql-client-parent</artifactId>
+    <version>3.9.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>vertx-mssql-client</artifactId>
+
+  <name>Vertx MSSQL Client</name>
+  <url>https://github.com/eclipse-vertx/vertx-sql-client</url>
+  <description>The Reactive MSSQL Client</description>
+
+  <properties>
+    <doc.skip>false</doc.skip>
+    <docs.dir>${project.basedir}/src/main/docs</docs.dir>
+    <generated.dir>${project.basedir}/src/main/generated</generated.dir>
+  </properties>
+
+  <dependencies>
+
+    <!-- Vert.x dependencies -->
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-codegen</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-docgen</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-sql-client</artifactId>
+      <version>${stack.version}</version>
+    </dependency>
+
+    <!-- Testing purposes -->
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>mssqlserver</artifactId>
+      <version>1.12.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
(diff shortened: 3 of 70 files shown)


gavinking:
Oh, that's great!

cescoffier:
The downstream cannot use Vert.x 4.0 milestone or snapshots before the end
of the year. So, without a backport an alternative would need to be found.

gavinking:
@pmlopes are any of those issues blockers for typical usage of the Vert.x client under Hibernate Reactive and especially with Quarkus? (This use case is the motivation for the backport.)

It doesn't *look* to me like they are, though I'm not sure about #605.

> I'd suggest as an alternative that downtream projects interested on this module to start using the `SNAPSHOT` builds on sonatype or the milestone releases instead.

That's not an option for Quarkus, and therefore also not an option for Hibernate Reactive.

vietj:
@gavinking can you use test this branch to see how much it fits your current expectations ?

gavinking:
@vietj I can try it out in standalone mode, I guess, though I'm not the right person to test the Quarkus scenario.

gavinking:
Alright, so I tried to get a very basic test (`SQLServerBasicTest`) working on SQL Server using the 3.9.1 snapshot, and here's the result:

<https://github.com/gavinking/hibernate-rx/tree/sqlserver>

I can see a couple of queries for next sequence values executed and returning successfully, but then as soon as I try to run the first `select` or `insert` query, it hangs waiting for a response from the driver.

I have no clue what the cause might be: something to do with us not using with transactions? Something to do with the connection properties I'm using? Something else?

vietj:
@gavinking can you make a reproducer case ? e.g you add a new test in the sql server tests and then we get an analysis from @BillyYccc

gavinking:
Folks, I have simplified it down to the following:

```java
package org.hibernate.rx;

import io.vertx.core.Vertx;
import io.vertx.ext.unit.Async;
import io.vertx.ext.unit.TestContext;
import io.vertx.ext.unit.junit.VertxUnitRunner;
import io.vertx.mssqlclient.MSSQLConnectOptions;
import io.vertx.mssqlclient.MSSQLPool;
import io.vertx.sqlclient.PoolOptions;
import org.junit.Test;
import org.junit.runner.RunWith;

@RunWith(VertxUnitRunner.class)
public class ConnectionTest {
	@Test
	public void testConnection(TestContext context) {
		MSSQLConnectOptions options = new MSSQLConnectOptions()
				.setHost( "localhost" )
				.setUser( "sa" )
				.setPassword( "reallyStrongPwd123" );
		MSSQLPool pool = MSSQLPool.pool( Vertx.vertx(), options, new PoolOptions() );
		Async async = context.async();
		pool.preparedQuery( "select * from Basic" ).execute(
				ar -> {
					System.out.println( ar.result().rowCount() );
					System.out.println("...done");
					async.complete();
				}
		);
		System.out.println("waiting...");
	}
}
```

This test hangs and never completes. Surely this must be something simple somehow related to my setup.

gavinking:
But the thing that seems strange is that if I replace the query with:

    select next value for hibernate_sequence

Then the test completes successfully. Which is why I wonder if this has something to do with txns.

vietj:
so you mean that "select * from INFORMATION_SCHEMA.SEQUENCES" has a problem ?

gavinking:
> so you mean that "select * from INFORMATION_SCHEMA.SEQUENCES" has a problem ?

Yes, but any table will do. It's not specific to `INFORMATION_SCHEMA.SEQUENCES`.

BillyYccc:
I have tried and found some of the data types in the table are not supported for now such as nvarchar.

gavinking:
It's not just that. Honestly I'm just finding a whole lot of really strange behavior (queries that sometimes work and sometimes don't). But it's really hard to debug because I never get any error.

I *think* the root problem is that it shouldn't hang when an error occurs. It should send the error back to my callback and/or terminate the async operation. This doesn't seem to happen: I never get errors at all, no matter how broken my query is. I can send the query `create` to the database and it hangs.

gavinking:
OK, so apart from the Big problem, which is the failure to report errors, it looks like the other major source of my pain is that it doesn't like inserting `null` into nullable columns.

Here's the test code I'm using to show this. Note that it uses testcontainers to create the db.

```java
package org.hibernate.rx;

import io.vertx.core.Vertx;
import io.vertx.ext.unit.Async;
import io.vertx.ext.unit.TestContext;
import io.vertx.ext.unit.junit.VertxUnitRunner;
import io.vertx.mssqlclient.MSSQLConnectOptions;
import io.vertx.mssqlclient.MSSQLPool;
import io.vertx.sqlclient.PoolOptions;
import org.junit.Test;
import org.junit.runner.RunWith;
import org.testcontainers.containers.MSSQLServerContainer;

@RunWith(VertxUnitRunner.class)
public class ConnectionTest {
	public static final MSSQLServerContainer<?> sqlserver = new MSSQLServerContainer<>()
			.withReuse( true );

	@Test
	public void testConnection(TestContext context) {
		sqlserver.start();
		System.out.println( sqlserver.getJdbcUrl() );
		System.out.println( sqlserver.getFirstMappedPort() );
		MSSQLConnectOptions options = new MSSQLConnectOptions()
				.setHost( "localhost" )
				.setPort( sqlserver.getFirstMappedPort() )
				.setUser( sqlserver.getUsername() )
				.setPassword( sqlserver.getPassword() );
		MSSQLPool pool = MSSQLPool.pool( Vertx.vertx(), options, new PoolOptions() );
		Async async = context.async();
		pool.preparedQuery( "create table WithNull (nullint int)" ).execute(
				ar1 -> {
					System.out.println( ar1.result().rowCount() );
					System.out.println("...inserting...");
					pool.preparedQuery("insert into WithNull (nullint) values (null)").execute(
						ar2 -> {
							System.out.println( ar2.result().rowCount() );
							System.out.println("...querying...");
							pool.preparedQuery("select * from WithNull").execute(
								ar3 -> {
									System.out.println( ar3.result().rowCount() );
									System.out.println("...done");
									sqlserver.stop();
									async.complete();
								}
							);
						}
					);
				}
		);
		System.out.println("creating table...");
	}
}
```

BillyYccc:
I have create an issue for this, see https://github.com/eclipse-vertx/vertx-sql-client/issues/627.
Some unsupported behaviors will throw `UnsupportedOperationException` which may not be captured correctly by the callback, they should be improved and be logged as an error maybe.

Issue #616, opened 2020-05-11
Generic API entry point


aguibert:
#### Describe the feature

In order for other projects (such as Hibernate Reactive) to build on top of the generic vertx-sql-client API, they need a mechanism to initialize a connection/client/pool without a direct dependency on the DB-specific API.

For example, Hibernate Reactive has to do stuff like this:
```java
switch ( uri.getScheme() ) {
	case "postgresql":
		PgConnectOptions pgOptions = new PgConnectOptions()
				.setPort( uri.getPort() )
				.setHost( uri.getHost() )
				.setDatabase( database )
				.setUser( username );
		if (password != null) {
			pgOptions.setPassword( password );
		}
		this.pool = PgPool.pool(Vertx.vertx(), pgOptions, poolOptions);
		break;
	case "mysql":
		MySQLConnectOptions mysqlOptions = new MySQLConnectOptions()
				.setPort( uri.getPort() )
				.setHost( uri.getHost() )
				.setDatabase( database )
				.setUser( username );
		if (password != null) {
			mysqlOptions.setPassword( password );
		}
		this.pool = MySQLPool.pool(Vertx.vertx(), mysqlOptions, poolOptions);
		break;
}
```

With JDBC, each driver is discoverable via the JDK ServiceLoader mechanism, so that no driver-specific API is needed in order to integrate with the driver.

Proposed changes:
- Define a new `io.vertx.sqlclient.Driver` interface, used for creating `Pool` objects given `SqlConnectOptions` objects
- Make `io.vertx.sqlclient.SqlConnectOptions` a concrete class so it can be directly instantiated
- Add static `pool()` methods to `io.vertx.sqlclient.Pool` that detect the Pool implementation from the ServiceLoader. An error is thrown if 0 or > 1 implementations are found
- Add `META-INF/service/io.vertx.sqlclient.Driver` files to each driver

#### Use cases

- Up-stack technologies can more easily integrate with vertx-sql-client if they can just depend on the generic `vertx-sql-client` artifact instead of needing to depend on each specific driver in order to initialize connections/pools.

#### Contribution

I can implement this feature

gavinking:
Agreed, I was thinking this too.

aguibert:
Closed by #618

PR #618, opened 2020-05-11
Driver service for generic API entry point

aguibert:
For #616 

Allows users to configure and use this API without any references to driver-specific APIs.

For example:
```java
SqlConnectOptions options = new SqlConnectOptions()
  .setHost("mydb.com")
  .setPort(1234)
  .setDatabase("mydb")
  .setUser("admin")
  .setPassword("admin");
Pool pool = Pool.pool(options);
pool.getConnection(ar -> {
 // ...
});
```

All existing API paths are uneffected, and if users need to set driver-specific connect options they can still use the driver-specific options classes (e.g. `new PgConnectOptions()`)


vietj:
why instead can't we add a method on ConnectOption that instantiates a driver instead of using an SPI mechanism here ? i.e have

```java
static Pool pool(SqlConnectOptions connectOptions, PoolOptions poolOptions) {
  return connectOptions.createPool(poolOptions);
}
```

aguibert:
@vietj What class are you suggesting that code snippet would be a part of? My goal is to  decouple the API entry point from the driver-specific API. If we were to do something like `connectOptions.createPool(poolOptions)`, then we'd still have the same service loader discovery pattern, but the logic would just be moved inside of the impl for SqlConnectOptions.

vietj:
Ok, I overlooked the fact drivers can create pool when the class `SqlConnectOptions` (which is now not anymore an abstract class) is provided.

In practice will that be the case ? i.e are we assuming HibernateRX does never instantiate database specific options like `DB2Options` ? because if it does in practice then we can proceed as I commented and avoid un necessary stuff.

aguibert:
> In practice will that be the case ? i.e are we assuming HibernateRX does never instantiate database specific options like DB2Options ?

Yes, I believe that will be the case. I suspect it will also be the case for the vast majority of users too, so they will be able to benefit from this too by having a generic entry point no matter what driver they are using.

gavinking:
> In practice will that be the case ? i.e are we assuming HibernateRX does never instantiate database specific options like `DB2Options` ? 

_Traditionally_, Hibernate hasn't stopped you from passing extra properties to the JDBC driver, but it hasn't needed to pass special options itself.

In fact, over time what's happened is that we've moved to having just one configuration item: the JDBC URL, and inferring the database dialect from that. You can of course add whatever options you like to the JDBC URL, but Hibernate will just ignore them and pass them along.

vietj:
I'm pretty sure that in practice some CDI code will create in practice DB2Options using Microprofile config in Quarkus :-)

vertx-sql-client/src/main/java/io/vertx/sqlclient/Driver.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+package io.vertx.sqlclient;
+
+import io.vertx.core.Vertx;
+
+/**
+ * An entry point to the Vertx Reactive SQL Client
+ * Every driver must implement this interface.
+ */
+public interface Driver {
```
vietj, line 25: we should have this in an sub-package, i.e `io.vertx.sqlclient.driver`

aguibert:
Another thing to consider is that this simplifies documentation and examples, because all of the "getting started" code samples can now look the same, and the only difference for various drivers will be the dependencies added to the pom.xml. 

IMO this will be good to have just from a flexibility/portability standpoint, and it's a pretty well-established pattern from other technologies like JDBC and most of JavaEE/MicroProfile.

vertx-sql-client/src/main/java/io/vertx/sqlclient/Pool.java
```diff
@@ -31,6 +37,63 @@
...
+   * @param connectOptions the options used to create the connection pool, such as database hostname
+   * @return the connection pool
+   */
+  static Pool pool(SqlConnectOptions connectOptions) {
+    return pool(connectOptions, new PoolOptions());
+  }
+  
+  /**
+   * Create a connection pool to the database configured with the given {@code connectOptions} and {@code poolOptions}.
+   *
```
vietj, line 53: we need to document that the implementation will look for a pool implementation available in the classpath, return the most specific one or the first one and fail iwhen no implementation is found.

vertx-db2-client/src/main/java/io/vertx/db2client/DB2Driver.java
```diff
@@ -0,0 +1,41 @@
...
+import io.vertx.sqlclient.Driver;
+import io.vertx.sqlclient.Pool;
+import io.vertx.sqlclient.PoolOptions;
+import io.vertx.sqlclient.SqlConnectOptions;
+
+public class DB2Driver implements Driver {
+
+  @Override
+  public Pool createPool(SqlConnectOptions options, PoolOptions poolOptions) {
+    return DB2Pool.pool(new DB2ConnectOptions(options), poolOptions);
```
vietj, line 28: when options is already a `DB2ConnectOptions` we should simply use it instead

aguibert:
@vietj thanks for the review, all review comments are addressed now and it's ready for you to take another look

PR #625, opened 2020-05-13
Tracing support

vietj:
Vert.x 4 supports tracing.

Fixes https://github.com/eclipse-vertx/vertx-sql-client/issues/653

### DIFF
--- a/vertx-pg-client/src/test/java/io/vertx/pgclient/TracingTest.java
+++ b/vertx-pg-client/src/test/java/io/vertx/pgclient/TracingTest.java
@@ -1,0 +1,185 @@
+package io.vertx.pgclient;
+
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+import io.vertx.core.Vertx;
+import io.vertx.core.VertxOptions;
+import io.vertx.core.spi.VertxTracerFactory;
+import io.vertx.core.spi.tracing.TagExtractor;
+import io.vertx.core.spi.tracing.VertxTracer;
+import io.vertx.core.tracing.TracingOptions;
+import io.vertx.ext.unit.Async;
+import io.vertx.ext.unit.TestContext;
+import io.vertx.sqlclient.PoolOptions;
+import io.vertx.sqlclient.RowSet;
+import io.vertx.sqlclient.SqlClient;
+import io.vertx.sqlclient.Tuple;
+import io.vertx.sqlclient.impl.tracing.QueryRequest;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+
+public class TracingTest extends PgTestBase {
+
+  Vertx vertx;
+  VertxTracer tracer;
+  PgPool pool;
+
+  @Before
+  public void setup() throws Exception {
+    super.setup();
+    vertx = Vertx.vertx(new VertxOptions().setTracingOptions(
+      new TracingOptions().setEnabled(true).setFactory(tracingOptions -> new VertxTracer() {
+        @Override
+        public Object sendRequest(Context context, Object request, String operation, BiConsumer headers, TagExtractor tagExtractor) {
+          return tracer.sendRequest(context, request, operation, headers, tagExtractor);
+        }
+
+        @Override
+        public void receiveResponse(Context context, Object response, Object payload, Throwable failure, TagExtractor tagExtractor) {
+          tracer.receiveResponse(context, response, payload, failure, tagExtractor);
+        }
+      }))
+    );
+    pool = PgPool.pool(vertx, options, new PoolOptions());
+  }
+
+  @After
+  public void teardown(TestContext ctx) {
+    vertx.close(ctx.asyncAssertSuccess());
+  }
+
+  @Test
+  public void testTraceSimpleQuery(TestContext ctx) {
+    String sql = "SELECT * FROM Fortune WHERE id=1";
+    testTraceQuery(ctx, sql, Collections.emptyList(), conn -> conn.query(sql).execute());
+  }
+
+  @Test
+  public void testTracePreparedQuery(TestContext ctx) {
+    String sql = "SELECT * FROM Fortune WHERE id=$1";
+    Tuple tuple = Tuple.of(1);
+    testTraceQuery(ctx, sql, Collections.singletonList(tuple), conn -> conn.preparedQuery(sql).execute(tuple));
+  }
+
+  @Test
+  public void testTraceBatchQuery(TestContext ctx) {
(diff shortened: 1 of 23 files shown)


vietj:
Pushed a new approach wiring the tracing in the client layer instead of the connection layer as the tracing cares about the client context before all.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/tracing/SqlTracer.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Tracer for SQL client, wrapping the generic tracer.
+ */
+public class SqlTracer {
+
+  enum RequestTags {
+
+    // Generic
+    PEER_ADDRESS("peer.address", q -> q.tracer.address),
+    PEER_SERVICE("peer.service", q -> "todo"),
```
BillyYccc, line 22: `peer.service` is not used for now

vertx-pg-client/src/test/java/io/vertx/pgclient/TracingTest.java
```diff
@@ -0,0 +1,185 @@
...
+  public void testTraceQuery(TestContext ctx, String expectedSql, List<Tuple> expectedTuples, Function<SqlClient, Future<?>> fn) {
+    AtomicBoolean called = new AtomicBoolean();
+    AtomicReference<Context> requestContext = new AtomicReference<>();
+    AtomicReference<Context> responseContext = new AtomicReference<>();
+    Object expectedPayload = new Object();
+    tracer = new VertxTracer<Object, Object>() {
+      @Override
+      public <R> Object sendRequest(Context context, R request, String operation, BiConsumer<String, String> headers, TagExtractor<R> tagExtractor) {
+        QueryRequest query = (QueryRequest) request;
+        ctx.assertEquals(query.sql(), expectedSql);
```
BillyYccc, line 91: in the assertions order of the expected value and actual value is reversed which reports wrong error messages when tests fail.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/tracing/SqlTracer.java
vietj: I think that peer.service is something the user can provide, I think we might simply have this part of the TracingOptions when creating Vert.x and leave it the tracer actually set it

vertx-pg-client/src/test/java/io/vertx/pgclient/TracingTest.java
vietj: right :-)

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/tracing/QueryRequest.java
```diff
@@ -0,0 +1,29 @@
...
+public class QueryRequest {
+
+  final SqlTracer tracer;
+  final String sql;
+  final List<Tuple> tuples;
+
+  public QueryRequest(SqlTracer tracer, String sql, List<Tuple> tuples) {
+    this.tracer = tracer;
+    this.sql = sql;
+    this.tuples = tuples;
```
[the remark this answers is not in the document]
vietj: aren't tuples the prepared statement parameters ?

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/tracing/SqlTracer.java
```diff
@@ -0,0 +1,94 @@
...
+  private final VertxTracer tracer;
+  private final String address;
+  private final String host;
+  private final int port;
+  private final String user;
+  private final String database;
+
+  public SqlTracer(VertxTracer tracer, SqlConnectOptions options) {
+    this.tracer = tracer;
+    this.address = options.getHost() + ":" + options.getPort();
```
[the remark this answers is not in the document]
vietj: likely yes, but we cannot use it in options because it's not an option object

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/tracing/SqlTracer.java
vietj: I think actually in our case we should not use this tag

vietj: from https://docs.instana.io/ecosystem/opentracing/ "Set the peer.service tag if you don't know the remote side is instrumented. In general, it is not required as long as the remote side is monitored by Instana and, in that case, it will automatically have a service name set. If peer.service is set in the client call and the server is instrumented and has a custom service name, the results are undefined." so we don't need it as it's a backend not service

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/tracing/QueryRequest.java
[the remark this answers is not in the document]
BillyYccc: I mean we could make sending statement parameters as part of payload an **option** for users. Because the parameter might be a big BLOB or text value, sending them over the network would  not be a good choice and there's hardly any benefit on this.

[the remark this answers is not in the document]
vietj: this actually is not logged by the tracers, if you look at it, a tracer would need to cast the request to a QueryRequest and do something with parameters. It seems that by default tracers don't care about parameters. So we only provide opportunity for a tracer to downcast to a `QueryRequest` and do something with them if they way. If that's a blob then it is up to the tracer to care about the payload issue, not the SQL client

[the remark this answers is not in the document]
BillyYccc: ok I see how this works in `vertx-tracing`

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/tracing/SqlTracer.java
BillyYccc: +1 I think in our SQL client usecase we don't need to add this.

BillyYccc:
`db.url` is required in OpenTelemetry spec https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md

It's not hard to support this after https://github.com/eclipse-vertx/vertx-sql-client/pull/644,
we could get `SqlConnectOptions` transformed into a short url like `[dbkind]://[username]@[host]:[port]/[databasename]`

vietj:
please don't name that kind :-)

vietj:
kind is typically the name you use when you can't find a good name :-)

BillyYccc:
I couldn't find an appropriate name for that either so I use it to show the usage 😅

vietj:
@BillyYccc as for naming and tags, I believe we need to have in master a definition of the tags we are going to use commonly in this project so each tracing integration can do a mapping of such tag to its own tagging system.

There are chances we will use the opentracing and/or opentelemetry list and then figure out. And then we will revisit this project for naming if required.

vietj:
I think the right name is `dbms` or `rdbms`

Issue #630, opened 2020-05-14
Allow general pipelining of connection initialization flows


aguibert:
#### Describe the feature

Currently the low-level getConnection flow looks like this:

```java
client.getConnection(ar -> {
  SqlConnection con = ar.result();
  // At this point a DB round-trip has already occurred. Usually to authenticate with the DB
});
```

Over the wire, that means we have done an entire DB round-trip before starting any queries on a new connection:
```
SEND: INIT_MSG
RECV: INIT_MSG_REPLY
SEND: QUERY_01
RECV: QUERY_01_REPLY
```

Databases such as DB2 allow commands to be "chained" in the same packet, which can reduce the number of round-trips in situations like this:
```
SEND: INIT_MSG & QUERY_01
RECV: INIT_MSG_REPLY & QUERY_01_REPLY
```
This is basically the same thing as the pipelining feature, except that the driver code is putting two commands into the same packet, rather than it being a net socket level optimization.

The implications of this change are that the initial getConnection() command cannot accurately tell if getting a connection will succeed or not (e.g. invalid credentials) and must consider itself "successful" before we really know if the connection will be valid or not. For example:
```java
client.getConnection(ar -> {
  SqlConnection con = ar.result();
  // at this point we have a connection, but we wouldn't know if it's good or not
  // until we try to use it below...
  con.query("<sql>").execute(ar2 -> {
    // If a connection could not be established for some reason, we would get 
    // that error here since this is the first use of the connection
  });
});
```

Overall I think this will be a good change (if all DBs support this "chaining" concept). It redefines what "getting a connection" means, but it seems to follow reactive principles more closely (i.e. nothing happens until you subscribe/use the thing)

#### Use cases

Database round-trips are by far the biggest performance factor for a DB driver, so whenever we can cut out round-trips it will help. However, this feature would only help if we are getting a new connection. It will help connection pools populate slightly faster, but once the connection pool has already been saturated then this doesn't have any added benefit.

#### Contribution

I can prototype this feature, but wanted to discuss on an issue first.

vietj:
@aguibert I am not sure this is an optimisation we want to do for a few reasons.

1. we used to have something like this for `Transaction` (i.e you would start a transaction but in practice you don't know if it really started until you do the first operation) and it was changed so that when you start a transaction you know you can proceed further along.

[...]

PR #631, opened 2020-05-15
Lazily flow connection init commands

aguibert:
Draft implementation of #630


aguibert:
Cancelling this PR since we aren't going to go with this approach

PR #633, opened 2020-05-15
Fix NPE in generic connection init path

aguibert:


BillyYccc:
it seems there's something wrong with the mssql client, the properties are required at present, I will investigate if I could make sending those properties optional.

aguibert:
@BillyYccc I added another commit that works around the issue in the test, but you are right, setting properties should not be required for basic usage

aguibert:
@BillyYccc I think I've fixed the issue properly now -- can you have a look?

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLConnectOptions.java
```diff
@@ -61,6 +61,9 @@ public MSSQLConnectOptions(JsonObject json) {
   
   public MSSQLConnectOptions(SqlConnectOptions other) {
     super(other);
+    if (getProperties().isEmpty()) {
```
BillyYccc, line 64: I think we should not set a default properties mapping here but check those options at connection level.
Can you take a look at https://github.com/eclipse-vertx/vertx-sql-client/pull/636?

If you're fine with that, you could merge it and rebase.

PR #640, opened 2020-05-17
Add support for customizing transaction start - fixes #432

### LINKED ISSUE #432: Support for setting Transaction Isolation
Reference: 

https://vertx.io/docs/apidocs/io/vertx/ext/sql/SQLOptions.html#getTransactionIsolation--
https://vertx.io/docs/apidocs/io/vertx/ext/sql/TransactionIsolation.html

BillyYccc:
Motivation:

fixes #432 

Conformance:

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Pool.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Pool.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
 
+import io.vertx.codegen.annotations.GenIgnore;
 import io.vertx.codegen.annotations.VertxGen;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
@@ -152,18 +153,13 @@ public interface Pool extends SqlClient {
    */
   <T> Future<T> withTransaction(Function<SqlClient, Future<T>> function);
 
-  /**
-   * Like {@link #withTransaction(Function, Handler)} but provides a customized way to start the transaction
-   * so that you could configure the transaction such as isolation levels or transaction characteristics at the start.
-   *
-   * @param startTransactionSql specify the SQL which is used to start the transaction
-   */
-  <T> void withTransaction(String startTransactionSql, Function<SqlClient, Future<T>> function, Handler<AsyncResult<T>> handler);
+  @GenIgnore
+  // FIXME codegen
+  <T> void withTransaction(TransactionOptions txOptions, Function<SqlClient, Future<T>> function, Handler<AsyncResult<T>> handler);
 
-  /**
-   * Like {@link #withTransaction(String, Function, Handler)} but returns a {@code Future} of the asynchronous result
-   */
-  <T> Future<T> withTransaction(String startTransactionSql, Function<SqlClient, Future<T>> function);
+  @GenIgnore
+  // FIXME codegen
+  <T> Future<T> withTransaction(TransactionOptions txOptions, Function<SqlClient, Future<T>> function);
 
   /**
    * Close the pool and release the associated resources.
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Pool.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Pool.java
@@ -38,7 +38,7 @@ import java.util.function.Function;
  */
 @VertxGen
 public interface Pool extends SqlClient {
-  
+
   /**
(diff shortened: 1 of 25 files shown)


vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/TransactionImpl.java
```diff
@@ -62,9 +63,9 @@
     }
   }
 
-  Future<Transaction> begin() {
+  Future<Transaction> begin(String startTxSql) {
```
vietj, line 66: startTxSql should be replace by an enum as each database might use something different to do this

vietj: also user don't want to know about the SQL for this :-)

BillyYccc:
I have rethinked and add a prototype here(the code may be a little messy but it can show the basic design)

The primary idea is to provide a simple API to configure transaction options at the start the transaction, and it would only support those common defined behaviors.
So now we have an unified API to configure transaction isolation levels and transaction access modes, specific database sql modifiers are not included in the API such as MySQL `WITH CONSISTENT SNAPSHOT` and Postgres `DEFERRABLE`.

Even though we have the same API, the sql to start a transaction in different databases is different, therefore we have to build the begin tx SQL at the codec level for different databases.
Since MySQL it does not support specifying the isolation level in the start tx sql like Postgres, a SQL `SET TRANSACTION ISOLATION LEVEL <...>` needs to be queried before the transaction start so that the isolation level is correctly set. I think we would go in the same way to solve this kind of issue for DB2/SQL Server as well.

If you're fine with this API, I will proceed to clean the code and work at this.

gavinking:
LGTM.

The ANSI standard options are `ISOLATION LEVEL READ UNCOMMITTED`, `ISOLATION LEVEL READ COMMITTED`, `ISOLATION LEVEL REPEATABLE READ`, `ISOLATION LEVEL SERIALIZABLE`, and `READ ONLY`, `READ WRITE`. 

ANSI SQL also defines `DIAGNOSTICS SIZE`, but I've never seen that supported in any database.

vertx-sql-client/src/main/java/io/vertx/sqlclient/TransactionOptions.java
```diff
@@ -0,0 +1,38 @@
+package io.vertx.sqlclient;
+
+import io.vertx.codegen.annotations.DataObject;
+
+//TODO codegen
+public class TransactionOptions {
+
+  public static final TransactionOptions DEFAULT_TX_OPTIONS = new TransactionOptions();
+
+  private TransactionIsolationLevel transactionIsolationLevel;
```
vietj, line 10: simply _isolationLevel_

vertx-sql-client/src/main/java/io/vertx/sqlclient/TransactionOptions.java
```diff
@@ -0,0 +1,38 @@
...
+  private TransactionAccessMode transactionAccessMode;
```
vietj, line 11: simply _accessMode_

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/command/StartTxCommand.java
```diff
@@ -0,0 +1,12 @@
+package io.vertx.sqlclient.impl.command;
+
+import io.vertx.sqlclient.TransactionOptions;
+
+public class StartTxCommand<R> extends TxCommand<R> {
+  public TransactionOptions transactionOptions;
+
+  public StartTxCommand(Kind kind, R result, TransactionOptions transactionOptions) {
```
vietj, line 8: rather here directly the access mode and the isolation level as _ TransactionOptions_ is mutable

vertx-sql-client/src/main/java/io/vertx/sqlclient/Pool.java
```diff
@@ -152,6 +153,14 @@ static Pool pool(Vertx vertx, SqlConnectOptions connectOptions, PoolOptions pool
    */
   <T> Future<T> withTransaction(Function<SqlClient, Future<T>> function);
 
+  @GenIgnore
+  // FIXME codegen
```
vietj, line 157: what is the codegen error here ?

vietj:
That makes some new classes in the top level package, I think we should have them in a `transaction` child package instead

vertx-sql-client/src/main/java/io/vertx/sqlclient/Pool.java
BillyYccc: it should be ok, I just don't add codegen annotations for a POC prototype, will add it later

Issue #524, opened 2020-02-22
MySQL geometry data type support


BillyYccc:
https://dev.mysql.com/doc/refman/8.0/en/spatial-types.html

PR #641, opened 2020-05-17
Add automatic module names

afloarea:
Signed-off-by: afloarea <adrianfloarea07@yahoo.com>

Motivation:

Add automatic module names as per [#524](https://github.com/vert-x3/issues/issues/524)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,6 +56,7 @@
     <tls.connection.uri/>
     <unix.socket.directory/>
     <unix.socket.port/>
+    <jar.manifest>${project.basedir}/src/main/resources/META-INF/MANIFEST.MF</jar.manifest>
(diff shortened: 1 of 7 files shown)


vertx-db2-client/pom.xml
```diff
@@ -34,6 +34,7 @@
   <properties>
     <doc.skip>false</doc.skip>
     <docs.dir>${project.basedir}/src/main/docs</docs.dir>
+    <jar.manifest>${project.basedir}/src/main/resources/META-INF/MANIFEST.MF</jar.manifest>
```
vietj, line 37: can we try having this in the parent pom of the project using inheritance ?

afloarea: done

PR #642, opened 2020-05-18
PgException getter and getMessage with all fields, PgErrorCodes

### LINKED ISSUE #274: Expose more details from PgException
:wave: Hey @vietj, I submitted a PR that got merged recently ( #272 ). It appears to now be defunct and not included in the most recent master?

Those additional error details are critical for complex query error handling.

I see that `PgException` just has 3 strings now. Can we not just expose the internal error type?

julianladisch:
Support debugging database errors by including all fields in PgException.

Fixes #274.

* Create a getter method for each field returned by PostgreSQL.

* The getMessage() method (overriding the RuntimeException method)
  has been returning the message field only. Now it returns all
  fields for complete information in logs. It creates a serialized
  JsonObject to support parsing the String, for example in
  microservice integration tests where it is returned in the
  HTTP body.

* To get only the message field use the new getErrorMessage()
  method.

* PgConnectionTest.java and PgPoolTestBase.java have been using
  PgException#getMessage() to assert that the query was cancelled.
  This method is much more verbose now. Change these asserts to
  check the SQLSTATE error code returned by PgException#getCode().
  "Applications that need to know which error condition has occurred
  should usually test the error code, rather than looking at the
  textual error message." (Quote from
  https://www.postgresql.org/docs/current/errcodes-appendix.html )

The https://github.com/vert-x3/vertx-mysql-postgresql-client
client returns all fields in GenericDatabaseException#getMessage(),
PgException should provide the same amount of data.


vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/Response.java
```diff
@@ -1,209 +1,58 @@
...
 /**
  *
  * <p>
  * A common response message for PostgreSQL
  * <a href="https://www.postgresql.org/docs/9.5/static/protocol-error-fields.html">Error and Notice Message Fields</a>
- *
- * @author <a href="mailto:emad.albloushi@gmail.com">Emad Alblueshi</a>
  */
+@VertxGen
+public interface Response {
```
BillyYccc, line 23: I think we don't actually need this API, we can put the fields in the `PgException`. In addition I'm not sure if all the fields are necessary to be exposed directly in the API.

vietj: agreeing with you, let's have fields on exception

vertx-pg-client/src/test/java/io/vertx/pgclient/impl/codec/ResponseHelper.java
```diff
@@ -0,0 +1,45 @@
...
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.pgclient.impl.codec;
+
+import io.vertx.pgclient.PgException;
+
+public final class ResponseHelper {
```
vietj, line 16: rather reuse the existing `io.vertx.pgclient.impl.util.Util` class that contains already similar static methods

julianladisch: The Response methods setMessage, setSeverity, etc. are package-private.
Response is in main/java/io/vertx/pgclient/impl/codec/Response.java.
ResponseHelper is a test helper located in test/java/io/vertx/pgclient/impl/codec/ResponseHelper.java. It is in test/... because it is needed for testing only. It is in io/vertx/pgclient/impl/codec to have access to the package-private methods of Response.java.
Moving the ResponseHelper methods into main/io/vertx/pgclient/impl/Util.java fails because of the package-private access restriction, and having a test helper needed for testing only should not got into main/...

vertx-pg-client/src/main/java/io/vertx/pgclient/PgErrorCodes.java
```diff
@@ -0,0 +1,306 @@
...
+
+package io.vertx.pgclient;
+
+/**
+ * SQLSTATE codes used by PostgreSQL
+ * (<a href="https://www.postgresql.org/docs/current/errcodes-appendix.html">PostgreSQL Error Codes</a>,
+ * <a href="https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/backend/utils/errcodes.txt;hb=HEAD">
+ * errcodes.txt</a>)
+ */
+public class PgErrorCodes {
```
vietj, line 20: I think this should be instead an enum

julianladisch: I don't think so.
The list of error codes is never complete. New PostgreSQL versions may add new codes, and any application can use custom codes: `do $$ begin raise sqlstate '99999'; end $$;`
Sub-classing PgErrorCodes allows to add custom codes.
An enum cannot be extended to support more codes.
There is no need to get a human readable name for a code at runtime because PgException contains a message and a detail field. PgErrorCodes provides a human readable name for an expected code used in a unit test or when to handle a specific error.
"You should use enum types any time you need to represent a fixed set of constants." (quote from https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html )

vietj: ok, then I think we should not have this list or error codes in the code

aguibert:
There is also a stray file checked into the repo root called `y` which was added to this PR by accident

julianladisch:
Thanks Andrew, I've removed the file `y`.

vertx-pg-client/src/main/java/io/vertx/pgclient/PgErrorCodes.java
julianladisch: I've removed the file PgErrorCodes.java.

PR #647, opened 2020-05-20
Add transaction accessor in `Connection` #646

### LINKED ISSUE #646: Add transaction accessor in `Connection`
Once a transaction has been created with `Connection.begin` it will be associated with the connection, so its implementation keeps it in a field.

It can be useful to add an accessor for it, so that we can check if a connection is in a transaction already, and to obtain the transaction without having to hang on to the connection/transaction tuple.

FroMage:
Fixes #646


vietj:
@FroMage you need to sign your commit

BillyYccc:
Note after a transaction is opened, the `tx` field is set to null only when the connection gets the response of COMMIT or ROLLBACK. I'm not sure it will reflect the transaction state correctly in command pipelining mode.

PR #650, opened 2020-05-21
DB2 support for LocalTime and BigDecimal

aguibert:
Also enhances the TCK tests a bit to ensure we are testing the `Row.getXXX()` code paths. Previously we were only testing `Row.getValue(int)` and `Row.getValue(String)`


aguibert:
For copyright dates we only have to update them if there are significant enough changes to the file. If the changes are considered "trivial" then legally we're not supposed to bump the year.

PR #658, opened 2020-05-26
rowDesc: add map for columns && indexes

anguisa:
Signed-off-by: Olga Ivanova <olga.a.ivanova@inbox.ru>

Motivation:

Improve performance of getting values by column names (search in map instead of list).

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/RowDesc.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/RowDesc.java
@@ -31,13 +31,13 @@ import java.util.stream.IntStream;
 public class RowDesc {
 
   private final List<String> columnNames;
-  private final Map<String, Integer> columns; // key - column name, value - column index
+  private final Map<String, Integer> columns; // key - column name, value - column index (if multiple columns with same names then last)
 
   public RowDesc(List<String> columnNames) {
     this.columnNames = columnNames;
     this.columns = Collections.unmodifiableMap(IntStream.range(0, columnNames.size())
         .boxed()
-        .collect(Collectors.toMap(columnNames::get, Function.identity())));
+        .collect(Collectors.toMap(columnNames::get, Function.identity(), (v1, v2) -> v2)));
   }
 
   public int columnIndex(String columnName) {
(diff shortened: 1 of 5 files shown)


vietj:
did you measure the performance in practice ? did you hit a bottleneck ?

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/RowDesc.java
```diff
@@ -26,16 +31,20 @@
 public class RowDesc {
 
   private final List<String> columnNames;
+  private final Map<String, Integer> columns; // key - column name, value - column index (if multiple columns with same names then last)
 
   public RowDesc(List<String> columnNames) {
     this.columnNames = columnNames;
+    this.columns = Collections.unmodifiableMap(IntStream.range(0, columnNames.size())
```
vietj, line 38: please simply create a `HashMap` that is modifiable (as it remains interna) and then a simple loop to put stuff in the map without lambdas and collectors

vietj:
did you encounter a performance issue to need this change ? in practice this will instantiate a map and I am not convinced that creating and populating a map here is a good trade

anguisa:
> did you measure the performance in practice ? did you hit a bottleneck ?
Initially these changes were made to conform vertx-sql-common and vertx-jdbc where JsonObject with map is used in ResultSet.
I tried to measure the performance. But the results were almost the same for 50 columns in sql query. In our projects we don`t have bigger queries. Maybe these changes are redundant.

anguisa:
> did you encounter a performance issue to need this change ? in practice this will instantiate a map and I am not convinced that creating and populating a map here is a good trade

I agree, for usage in sql queries it is useless. Only waste of memory.
Should I close my PR?

P.S. It is something wrong with io.vertx.pgclient.TracingTest. Sometimes it fails: 
testTraceBatchQuery(io.vertx.pgclient.TracingTest)  Time elapsed: 0.04 sec  <<< FAILURE!
java.lang.AssertionError: Expected io.vertx.core.impl.EventLoopContext@64e56c23 got null
	at io.vertx.ext.unit.impl.TestContextImpl.reportAssertionError(TestContextImpl.java:362)
	at io.vertx.ext.unit.impl.TestContextImpl.assertEquals(TestContextImpl.java:252)
	at io.vertx.ext.unit.impl.TestContextImpl.assertEquals(TestContextImpl.java:245)
	at io.vertx.pgclient.TracingTest.lambda$null$4(TracingTest.java:118)

vietj:
let's close this for now and see if that is ever an issue reported.

I think the tracing test is racy and it seems that @BillyYccc fixed it

PR #659, opened 2020-05-26
replace SQL template placeholder $ with # instead - fixes #655

### LINKED ISSUE #655: Reconsider SQL template placeholder as it does not work with string template
Currently the SQL template supports SQL syntax like "SELECT * FROM users WHERE id=${id}" and replaces the SQL template holder(${}) with a real SQL placeholder such as ?,$1, etc...

This can work perfectly in Java, however languages like Kotlin and Javascript(ECMAScript 6) supports string template grammar which will interpolate variable into the SQL string, therefore the SQL template couldn't be used.

I propose to use #{} instead of ${} as the sql template placeholder, it enhances the semantic that the sql template statement execution is not a literal value interpolation as well. With this change we need also to check whether it works because Postgres uses # as the bitwise XOR operator.

BillyYccc:
Motivation:

Fixes #655

Conformance:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,10 @@ jobs:
       name: "Postgres 10"
       jdk: openjdk8
       script: mvn -q clean verify -B -Dembedded.postgres.version=10.6 --projects vertx-sql-client,vertx-pg-client
+    - stage: test
+      name: "SQL Template"
+      jdk: openjdk8
+      script: mvn -q clean verify -B --projects vertx-sql-client,vertx-sql-client-templates
     - stage: deploy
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -51,6 +51,10 @@ jobs:
       name: "Postgres 10"
       jdk: openjdk8
       script: mvn -q clean verify -B -Dembedded.postgres.version=10.6 --projects vertx-sql-client,vertx-pg-client
+    - stage: test
```
vietj, line 54: duded I missed that actually CI was not running SQL client templates

PR #662, opened 2020-05-27
Make recycling pooled connection events emitted on the eventloop context associated with connection pool

BillyYccc:
Motivation:

available pooled connection deque in `ConnectionPool` might poll a null value which might cause the client hangs.

Conformance:

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/ConnectionPool.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/ConnectionPool.java
@@ -242,6 +242,10 @@ public class ConnectionPool {
         while (waiters.size() > 0) {
           if (available.size() > 0) {
             PooledConnection proxy = available.poll();
+            if (proxy == null) {
+              // available is empty?
+              return;
+            }
             Handler<AsyncResult<Connection>> waiter = waiters.poll();
             waiter.handle(Future.succeededFuture(proxy));
           } else {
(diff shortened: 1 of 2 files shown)


BillyYccc:
@vietj can you please take a look at this? I think it might fix the cause of CI failing of the PG pool tests.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/ConnectionPool.java
```diff
@@ -242,6 +242,10 @@ private void check() {
         while (waiters.size() > 0) {
           if (available.size() > 0) {
             PooledConnection proxy = available.poll();
+            if (proxy == null) {
```
vietj, line 245: did you check this case can happen ? this code is single threaded, i.e always the same thread shall access the pool (perhaps it does not and in that case it's a bug), so the expression `available.size() > 0` being true implies that `proxy != null` is true as well

BillyYccc: Yes the connection pool actions should be all emitted on the eventloop context associated with the pool so it should be impossible for that to happen.

But I found it incorrect for the closing method which might cause different threads access the states in connection pool. The close method is not called on the associated context
https://github.com/eclipse-vertx/vertx-sql-client/blob/04f1f415e274a2e2d7a911067aaf9422d292799d/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/ConnectionPool.java#L171-L186

BillyYccc:
Fixes #666

PR #663, opened 2020-05-27
Decimal precision fix

aguibert:


pom.xml
```diff
@@ -48,6 +48,8 @@
 
   <properties>
     <stack.version>4.0.0-SNAPSHOT</stack.version>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
```
aguibert, line 52: CC @vietj to make sure you are OK with this change

BillyYccc: Why do we need this as it's already defined in https://github.com/vert-x3/vertx-parent/blob/master/pom.xml?

aguibert: I thought I ran into an incompatible class version change and this fixed it, but I just double checked and it looks like we were indeed compiling to 1.8 before, so I must have been mistaken -- I'll remove this

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/DRDAQueryRequest.java
```diff
@@ -1399,7 +1409,7 @@ private Hashtable computeProtocolTypesAndLengths(
                     lidAndLengths[i][0]--;
                 }
             }
-            return overrideMap;
+            return overrideMap; // @AGG this is never used
```
aguibert, line 1412: Intentionally leaving this comment here so I remember to clean it up once I do the global cleanup of the drda package

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/NetSqlca.java
```diff
@@ -123,25 +123,31 @@ public static void throwSqlError(NetSqlca sqlca) {
...
+	   	    		throw new DB2Exception("The database '" + sqlErrmc + "' provided was not found", sqlca.sqlCode_, sqlca.sqlState_);
 	   	    	else
-	   	    		throw new DB2Exception("The database provided was not found", SqlCode.DATABASE_NOT_FOUND, sqlca.sqlState_);
+	   	    		throw new DB2Exception("The database provided was not found", sqlca.sqlCode_, sqlca.sqlState_);
+	   	    case SqlCode.DATA_TYPE_INVALID_ATTR:
+	   	        if (sqlErrmc.length() > 0)
+	   	            throw new DB2Exception("The statement cannot be processed. The data type definition '" + sqlErrmc + 
+	   	                "' specifies an invalid attribute such as precision, lenght, or scale.", sqlca.sqlCode_, sqlca.sqlState_);
+	   	        else
+	   	            throw new DB2Exception("The statement cannot be processed. A data type definition specifies an invalid attribute such as precision, lenght, or scale.", sqlca.sqlCode_, sqlca.sqlState_);
```
gjwatts, line 150: Same typo (copy/paste) :-)

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/DRDAQueryRequest.java
gjwatts: Thanks for letting me know

PR #665, opened 2020-05-28
Initial database metadata API

### LINKED ISSUE #645: Database metadata SPI
#### Describe the feature

Upstack libraries (i.e. Hibernate/JPA) typically require some sort of global database metadata in order to determine what types of features/sql syntax are supported.

In JDBC they have 2 tiers of metadata APIs:
- [DatabaseMetaData](https://docs.oracle.com/en/java/javase/11/docs/api/java.sql/java/sql/DatabaseMetaData.html)
- [ResultSetMetadata](https://docs.oracle.com/en/java/javase/11/docs/api/java.sql/java/sql/ResultSetMetaData.html)

We already have a very primitive representation of RS metadata through `Row#getColumnName(String)` and `Row#getColumnIndex(int)`.

The minimum amount of metadata needed for an equivalent of `DatabaseMetaData` would be obtaining the DB major/minor version.

#### Use cases

- Hibernate needs to be able to check database major/minor versions to determine what features the backend DB supports
- It could also be useful for each driver to check DB major/minor version as part of a support statement. For example, I have only tested the DB2 driver with DB2 11.5, and the DB2 driver should probably throw an error on anything less than 11.5 because it hasn't been tested.

aguibert:
Closes #645 

This allows a new `DatabaseMetaData` object to be retrieved from any `SqlConnection` via `getDatabaseMetaData()`.

Currently the DB metadata has 3 properties:
 - DB product name (String)
 - DB major version (int)
 - DB minor version (int)

CC @gavinking to make sure this is what you need in HR


gavinking:
Great, LGTM.

FYI, in `org.hibernate.dialect.Database` you can see an example of how this info is used.

vertx-sql-client/src/main/java/io/vertx/sqlclient/spi/DatabaseMetaData.java
```diff
@@ -0,0 +1,26 @@
+package io.vertx.sqlclient.spi;
+
+import io.vertx.codegen.annotations.VertxGen;
+
+/**
+ * Contains static metadata about the backend database server 
+ */
+@VertxGen
+public interface DatabaseMetaData {
```
vietj, line 9: `DatabaseMetadata` , metadata is a single word (https://en.wikipedia.org/wiki/Metadata)

vertx-sql-client/src/main/java/io/vertx/sqlclient/spi/DatabaseMetaData.java
```diff
@@ -0,0 +1,26 @@
...
+  
+  /**
+   * @return The product name of the backend database server
+   */
+  public String getDatabaseProductName();
```
vietj, line 14: simply `databaseProductName()`

vertx-sql-client/src/main/java/io/vertx/sqlclient/spi/DatabaseMetaData.java
```diff
@@ -0,0 +1,26 @@
...
+  
+  /**
+   * @return The major version of the backend database server
+   */
+  public int getDatabaseMajorVersion();
```
vietj, line 19: simply `majorVersion()`, no need to prefix with `database` as we are in a class prefixed `Database`

vertx-sql-client/src/main/java/io/vertx/sqlclient/spi/DatabaseMetaData.java
```diff
@@ -0,0 +1,26 @@
...
+  
+  /**
+   * @return The minor version of the backend database server
+   */
+  public int getDatabaseMinorVersion();
```
vietj, line 24: simply `minorVersion()`, no need to prefix with `database` as we are in a class prefixed `Database`

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlConnection.java
```diff
@@ -88,5 +89,7 @@
    * @param handler the completion handler
    */
   void close(Handler<AsyncResult<Void>> handler);
+  
+  DatabaseMetaData getDatabaseMetaData();
```
vietj, line 93: rename to `databaseMetadata()` - I'm wondering if that should not be directly on `SqlClient`

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLDatabaseMetaData.java
```diff
@@ -0,0 +1,47 @@
...
+      String releaseNumberString = releaseVersion.substring(0, indexOfFirstSeparator);
+      microVersion = Integer.parseInt(releaseNumberString);
+    } else {
+      microVersion = Integer.parseInt(versionNumbers[2]);
+    }
+  }
+
+  @Override
+  public String getDatabaseProductName() {
+    return "MySQL"; // TODO: Should this return MariaDB sometimes?
```
BillyYccc, line 30: I think I need to rewrite the parser sometime later as it does not consider the situations of different database version semantics.

aguibert: Ok, at the moment it didn't seem like there was any MariaDB-specific code in the MySQL driver, aside from URL parsing.

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlConnection.java
BillyYccc: Usually exposing such metadata is an internal feature for those users, I think it's better to keep the `SqlClient` API simple.

vertx-sql-client/src/main/java/io/vertx/sqlclient/spi/DatabaseMetaData.java
aguibert: This will be a fun review comment to fix on a case since I'm on a case insensitive OS 🙃

aguibert:
@gavinking thanks for the pointer. I am trying to adjust the values returned by `DatabaseMetadata#getProductName()` so they match the resolution code in `Database#resolveDialect()` and I noticed that for DB2 it seems like either an AS400 dialect is chosen or the oldest DB2 dialect is chosen:
https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/Database.java#L47-L75

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlConnection.java
aguibert: The SqlClient is more focused on one-shot operations, whereas obtaining DB metadata isn't really a one-shot operation, it's mainly something that would make sense to obtain after you've obtained a connection (like in JDBC).
Also, I'm considering the new metadata class as an SPI, which also supports @BillyYccc 's comment about keeping it off of the "golden path" API usage

gavinking:
@aguibert Well it's probably better to look at the new version of that class that I worked on for Hibernate 6:

<https://github.com/hibernate/hibernate-orm/blob/wip/6.0/hibernate-core/src/main/java/org/hibernate/dialect/Database.java>

aguibert:
Thanks, I see you are doing some parsing of the raw version string so I'll add that to the Metadata API also: `String DatabaseMetadata#fullVersion()`

aguibert:
@vietj @BillyYccc thanks for reviewing, I've applied the review comments and also added a new method `String fullVersion()` that gives access to the full/raw version string supplied by the DB. I also added proper version parsing for MariaDB since it has a different version scheme than MySQL

vertx-sql-client/src/main/java/io/vertx/sqlclient/spi/DatabaseMetaData.java
vietj: sometimes I do ahve this too, here is how I deal with:

1/ rename the file to something else like Foo
2/ commit
3/ rename Foo to the old name with case sensitivity change
4/ commit
5/ squash commits

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLDatabaseMetadata.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+  
+  public boolean isMariaDB() {
+    return isMariaDB;
+  }
+
+  @Override
+  public String productName() {
+    return isMariaDB ? "MariaDB" : "MySQL";
```
BillyYccc, line 47: There're a wide range of database servers supporting MySQL C/S protocol such as Percona, Vitess, TiDB... and their reponse version strings differ much. Don't worry about this, I will fix this in another PR since the version string is also used to control the EOF header flag setting in MySQL.

vietj: I think we should in the doc clearly tells which database are tested and that other is a best effort support

BillyYccc: There is something like what you describe in https://vertx.io/docs/vertx-mysql-client/java/#_mysql_and_mariadb_version_support_matrix 

I got a repo for testing basic features with more databases in https://github.com/BillyYccc/vertx-mysql-client-compatibility-testkit which is so-called best effort support

Issue #675, opened 2020-06-04
Prepared query execution flow redesign


vietj:
Currently one shot prepared query is a composite execution of a `PrepareStatementCommand` and `ExtendedQueryCommand`. The first command will prepare the statement and the later will execute it.

When prepare statement are cached, the `PrepareStatementCommand` is only executed once (assuming the statement is not evicted from the cache).

When prepare statement are not cached then it is unclear what should happen and likely this depends on the database.

# What we have now

### What is happening

Todo (study)

### What should happen

A two steps executions issues the following commands

- `PrepareCommand`
  - PARSE
  - DESCRIBE
  - SYNC
- `ExtendedPreparedQuery`
  - BIND
  - EXECUTE
  - SYNC

One shot prepared query should be executed using un-named prepared statement, so the sequence above is replaced by:

- PARSE
- BIND
- EXECUTE
- SYNC

### What is happening

two-step executions(i.e. prepare->execute) & one-shot preparedQuery

* PrepareCommand
send `COM_STMT_PREPARE` to server

* ExtendedQueryCommand
locally bind the parameters and set the metadata
send `COM_STMT_EXECUTE` to server

* CloseStatementCommand
send `COM_STMT_CLOSE` to server

### What should happen

There is no one-shot prepared-query protocol specific command in MySQL. so we can't change the command flow.

### What is happening

two-step executions(i.e. prepare->execute) & one-shot preparedQuery

* PrepareCommand
do nothing but store the SQL string in the client

* ExtendedQueryCommand
locally bind the parameters and set the metadata
send `sp_executesql` request to the server

* CloseStatementCommand
do nothing because we use `sp_executesql`

### What should happen

two-step executions(i.e. prepare->execute)

* PrepareCommand
send `sp_prepare` request to the server

* ExtendedQueryCommand
send `sp_execute` request to the server

* CloseStatementCommand
send `sp_unprepare` request to the server

one-shot preparedQuery
send `sp_executesql` request to the server (using T-SQL)

### What is happening

* PrepareCommand
Send `PREPARE` and `DESCRIBE` commands

* ExtendedQueryCommand
Send `OPEN_QUERY` passing statement params in bound according to column metadata retrieved from `PREPARE` and `DESCRIBE`

* CloseStatementCommand
No-op

### What should happen

[...]

PR #677, opened 2020-06-05
Rework prepared queries

vietj:
fixes 

- https://github.com/eclipse-vertx/vertx-sql-client/issues/675
- https://github.com/eclipse-vertx/vertx-sql-client/issues/682
- https://github.com/eclipse-vertx/vertx-sql-client/issues/685


BillyYccc:
Does this imply the statement auto-closing after one-shot preparedQuery is handled by the codec layer?

vietj:
@BillyYccc yes it is implied

vietj:
@BillyYccc @aguibert I would like to know if the above is fine and then we can start finishing this PR together and after back-port it to 3.9

vietj:
It might be possible actually to avoid changes in the codec by moving the two step execution logic in the socket instead of moving to the codec which would avoid changing code in the current implementations of the codecs, I'll keep you informed of the progress.

vietj:
I updated this PR with an intermediate change that does not require modification of current codecs and is actually more satisfying. I'll update the proposal issue to be in sync with this change.

vietj:
so this should not involve contributions from @aguibert and @BillyYccc

vietj:
Updated with prepared query caching simplification and corrections.

vietj:
I'm moving this PR to review

vietj:
To ease the review I will push the first commit that refactor the extended query commands in a single command to master as it is not related to the PR and the goal was to simplify and help this change

vietj:
this is ready for review

vietj:
Pushed a fix for the incorrect pipelining count that was found by a case provided by @BillyYccc

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SocketConnectionBase.java
```diff
@@ -70,7 +72,8 @@ public SocketConnectionBase(NetSocketInternal socket,
     this.socket = socket;
     this.context = context;
     this.pipeliningLimit = pipeliningLimit;
-    this.psCache = cachePreparedStatements ? new PreparedStatementCache(this, preparedStatementCacheSize) : null;
+    this.paused = false;
+    this.psCache = cachePreparedStatements ? new PreparedStatementCache(preparedStatementCacheSize) : null;
     this.preparedStatementCacheSqlLimit = preparedStatementCacheSqlLimit;
```
BillyYccc, line 77: we should use this option in the cache to make it work like before

BillyYccc: I mean `preparedStatementCacheSqlLimit` here

vietj: why is this feature useful ? can we just instead remove it ?

vietj:
I back ported the commits @BillyYccc

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SocketConnectionBase.java
vietj: I added a generic filter for this, see the latest commit

vietj:
@BillyYccc added support for SQL cache filtering, I think now we are good and we can do a final pass on the review

vietj:
it was a collaboration as usual @BillyYccc :-)

PR #678, opened 2020-06-05
Add error message and test for sqlcode -601

aguibert:


vertx-db2-client/src/test/java/io/vertx/db2client/DB2ErrorMessageTest.java
```diff
@@ -261,6 +261,24 @@ public void testInsertIntoGeneratedAlwaysColumn(TestContext ctx) {
...
+    public void testDuplicateObject(TestContext ctx) {
+      DB2Connection.connect(vertx, options, ctx.asyncAssertSuccess(conn -> {
+        conn.query("CREATE TABLE Fortune (\n" + 
+            "  id       integer       NOT NULL GENERATED AS IDENTITY (START WITH 1, INCREMENT BY 1),\n" + 
+            "  message  varchar(2048),\n" + 
+            "  PRIMARY KEY  (id)\n" + 
+            ")").execute(ctx.asyncAssertFailure(err -> {
+            ctx.assertTrue(err instanceof DB2Exception, "The error message returned is of the wrong type.  It should be a DB2Exception, but it was of type " + err.getClass().getSimpleName());
+            DB2Exception ex = (DB2Exception) err;
+            assertContains(ctx, ex.getMessage(), "The object with the name '");
```
gjwatts, line 275: In Windows, this test (line 275) seems to fail because of the trailing single quote ' (at least removing it made the test pass).
On the Mac, this test passed fine without modification.

vertx-db2-client/src/test/java/io/vertx/db2client/DB2ErrorMessageTest.java
```diff
@@ -261,6 +261,24 @@ public void testInsertIntoGeneratedAlwaysColumn(TestContext ctx) {
...
+            assertContains(ctx, ex.getMessage(), "Fortune'");
```
gjwatts, line 276: On both Windows and Mac, this test (line 276) failed for me because of the mixed case of the word `Fortune`.  In the error message that comes back, it shows the table name as being all uppercase FORTUNE.  Changing the test case to look for the word FORTUNE allowed the test to pass.  Strangely enough, this test succeeds in both Windows and Mac when leaving the trailing single quote at the end of FORTUNE' (which seems to cause problems in the previous test).

aguibert: Good catch, didn't have time to run this test locally before I opened the PR. Got it fixed now and CI is green

Issue #700, opened 2020-06-30
Prepare statement with parameters types


vietj:
The creation of prepared statements requires in some cases to provide the type of parameters, e.g PostgreSQL can optionally use them and provide an adequate binding (like the query `SELECT CONCAT("hello ", $1)`) otherwise it would fail.

We should prepare statements without types and upon specific failures (e.g `42P18` _INDETERMINATE DATATYPE_) retry with the types.

The preparation/execution flow should be reworked to do this. This affects prepared queries and one shot queries.

PR #703, opened 2020-06-30
Prepare statement with parameters types

vietj:
Implements https://github.com/eclipse-vertx/vertx-sql-client/issues/700


vietj:
@BillyYccc can you have a look ?

vietj:
it's currently missing type binding and tests (i.e for now it only does for String to VARCHAR and Integer to INT4)

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/Connection.java
```diff
@@ -23,10 +23,14 @@
 
 public interface Connection extends CommandScheduler  {
 
+  default boolean isLazyException(Throwable t) {
```
BillyYccc, line 26: what does lazy exception mean?

vietj: it means it checks whether the exception is a lazy statement, I'll try finding a better name, for now this PR is not polished at all but it provides most of what we want to do.

vietj: something like isIncompletePreparedStatementException(T) would be better likely

PR #708, opened 2020-07-02
Unix Domain Socket support in ContainerPgRule

tsegismont:
Replaces embedded Pg with container testing for Unix domain sockets


tsegismont:
@vietj can you please take a look and try on your machine? Thank you

vietj:
this isn't something we want to do because it cannot be tested on OS X since Docker for Mac is not able to share domain sockets, see https://github.com/docker/for-mac/issues/483

tsegismont:
Ok, let's put this on hold then

vietj:
yes, that's unfortunate, you can give your +1 on the Docker GitHub issue to make it progress.

BillyYccc:
This is in fact duplicated of https://github.com/eclipse-vertx/vertx-sql-client/pull/520. 
I wonder if we could set up a specific database with unix socket in the CI for tests.

tsegismont:
Ha, I had not noticed @BillyYccc , sorry.

@vietj on second thoughts, do we absolutely need an automated test for BSD and the likes? There is nothing specific in our code, we just pull a different dependency. And CI validates it works on Linux.

What are the chances something goes wrong with kqueue specifically? And if we get a bug report, there is still the option to start a server installed from `brew` and test manually in the background.

PR #710, opened 2020-07-03
connectionReleaseDelay PoolOption

julianladisch:
Close idle connections that wait in the pool for connectionReleaseDelay
milliseconds. Defaults to 0 = keep them forever.

The option name is taken from vertx-mysql-postgresql-client:
https://github.com/vert-x3/vertx-mysql-postgresql-client/blob/3.9.1/vertx-mysql-postgresql-client-jasync/src/main/java/io/vertx/ext/asyncsql/impl/pool/AsyncConnectionPool.java#L68

Implementation
If connectionReleaseDelay is > 0:
When a connection is put into the pool create a timer
that closes the connection and removes it from the pool: expire().
If a connection in the pool is acquired before the timer
fires the timer is cancelled: cancelIdleTimer().

In PoolOptions
* add connectionReleaseDelay option
* fix and unit test equals() and hashCode()

In ConnectionPool
* add constructor that takes PoolOptions to avoid long parameter lists
* add package-private allSize() for unit testing
* rename and move release(PooledConnection) to PooledConnection.addToPool()
  to avoid a name that is misleading if connectionReleaseDelay exists

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/PoolOptions.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/PoolOptions.java
@@ -19,6 +19,7 @@ package io.vertx.sqlclient;
 
 import io.vertx.codegen.annotations.DataObject;
 import io.vertx.core.json.JsonObject;
+import java.util.Objects;
 
 /**
  * The options for configuring a connection pool.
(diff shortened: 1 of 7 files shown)


vietj:
the pool should be reimplemented before 4.0 or around, so I would rather do that in the new pool rather than adding this here

julianladisch:
We use and will use the old pool in production and therefore would like to have this merged upstream (= in eclipse-vertx/vertx-sql-client) to avoid maintaining a fork.

vertx-sql-client/src/main/java/io/vertx/sqlclient/PoolOptions.java
```diff
@@ -101,21 +129,20 @@ public JsonObject toJson() {
 
   @Override
   public boolean equals(Object o) {
-    if (this == o) return true;
-    if (!(o instanceof PoolOptions)) return false;
-    if (!super.equals(o)) return false;
```
craigmcnally, line 106: Was this intentionally removed?

julianladisch: No, the first two line have been reformatted, the third line has been changed and reformatted.

julianladisch:
@vietj Is the code quality of this pull request too low, or is the scheduled pool re-implementation the only reason?

julianladisch:
@vietj What needs to be changed to get this merged to eclipse-vertx/vertx-sql-client?

vietj:
at least `connectionReleaseDelay` that might not be part of the next options for this pool

vietj:
it is rescheduled reimplementation based on the vertx HTTP pool but that would be in 4.1 or 4.2

Issue #615, opened 2020-05-10
Java enum Row/Tuple support for DB2


vietj:
See https://github.com/eclipse-vertx/vertx-sql-client/issues/612

aguibert:
Delivered in PR #715

PR #715, opened 2020-07-15
Add support for enums in Row/Tuple for DB2

gjwatts:
Add support for enums in Row/Tuple for DB2

Issue #615

### DIFF
--- a/vertx-db2-client/src/main/java/examples/DB2ClientExamples.java
+++ b/vertx-db2-client/src/main/java/examples/DB2ClientExamples.java
@@ -15,14 +15,12 @@
  */
 package examples;
 
-import java.math.BigDecimal;
 import java.util.Map;
 import java.util.stream.Collector;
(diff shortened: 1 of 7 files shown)


vertx-db2-client/src/main/java/examples/DB2ClientExamples.java
```diff
@@ -15,14 +15,13 @@
  */
 package examples;
 
-import java.math.BigDecimal;
 import java.util.Map;
 import java.util.stream.Collector;
 import java.util.stream.Collectors;
 
+import examples.Scratch.Days;
```
aguibert, line 22: looks like this import is breaking the build (since Scratch.java is not checked in)

gjwatts: Yes, I missed that one.  Thanks for catching that for me!  I have corrected it and pushed out the fixed version.

PR #720, opened 2020-07-29
Tweak prepared statement execution work flow, this patch brings two c…

### LINKED ISSUE #719: Cached prepared statement do not bind the parameters before execution
I can reproduce this with Hibernate Reactive: https://github.com/hibernate/hibernate-reactive/blob/c78ee19330a655ae221e855d7cc739d2dc0e6acb/hibernate-reactive-core/src/test/java/org/hibernate/reactive/types/SerializableExceptionTest.java

What's happening is that if I run the same query twice with different parameters, the first time the client will try to infer the data type from the parameters and then used it for the subsequent queries.
Here what's happening in my case:
We have a table `Basic` with a blob column. The first time I run an insert the Blob value for the column is null. The client will select `DataType.NULL` instead of `DataType.BLOB`. The second time I run an insert the blob parameter is not null and the client will still use `DataType.NULL` instead of `DataType.BLOB`.

This happens with version 3.9.2 and only with MySQL. It was working with 3.9.1.

The PR on Hibernate Reactive  if interested: https://github.com/hibernate/hibernate-reactive/pull/293

BillyYccc:
…hanges:

1. Parameters should always be prepared when executing prepared statement, which will apply to all the clients
2. MySQL prepared statement prepare() method does not bind parameters any more, the binding phase is MySQL specific behavior and will happen before statement execution

fixes #719

### DIFF
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/ExtendedBatchQueryCommandCodec.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/ExtendedBatchQueryCommandCodec.java
@@ -51,6 +51,7 @@ class ExtendedBatchQueryCommandCodec<R> extends ExtendedQueryCommandBaseCodec<R,
       String bindMsg = statement.bindParameters(param);
       if (bindMsg != null) {
         completionHandler.handle(CommandResponse.failure(bindMsg));
+        return;
       }
       sendBatchStatementExecuteCommand(statement, param);
       batchIdx++;
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/ExtendedBatchQueryCommandCodec.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/ExtendedBatchQueryCommandCodec.java
@@ -47,6 +47,11 @@ class ExtendedBatchQueryCommandCodec<R> extends ExtendedQueryCommandBaseCodec<R,
     if (batchIdx < params.size()) {
       sequenceId = 0;
       Tuple param = params.get(batchIdx);
+      // binding parameters
+      String bindMsg = statement.bindParameters(param);
+      if (bindMsg != null) {
+        completionHandler.handle(CommandResponse.failure(bindMsg));
+      }
       sendBatchStatementExecuteCommand(statement, param);
       batchIdx++;
     }
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SocketConnectionBase.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SocketConnectionBase.java
@@ -161,6 +161,14 @@ public abstract class SocketConnectionBase implements Connection {
           paused = true;
           inflight++;
           cmd = prepareCmd;
+        } else {
+          // cached prepared statement needs to prepare the parameters before execution
+          String msg = queryCmd.prepare();
+          if (msg != null) {
+            inflight--;
+            queryCmd.fail(new NoStackTraceThrowable(msg));
+            return;
+          }
         }
       }
       written++;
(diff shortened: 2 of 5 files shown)


BillyYccc:
The changes should also be applied in `4.0` branch.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SocketConnectionBase.java
```diff
@@ -161,6 +161,14 @@ private void checkPending() {
           paused = true;
           inflight++;
           cmd = prepareCmd;
+        } else {
+          // cached prepared statement needs to prepare the parameters before execution
+          String msg = queryCmd.prepare();
+          if (msg != null) {
```
vietj, line 167: can we add a general test for this ?

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/ExtendedBatchQueryCommandCodec.java
```diff
@@ -47,6 +47,11 @@ private void doExecuteBatch() {
     if (batchIdx < params.size()) {
       sequenceId = 0;
       Tuple param = params.get(batchIdx);
+      // binding parameters
+      String bindMsg = statement.bindParameters(param);
+      if (bindMsg != null) {
+        completionHandler.handle(CommandResponse.failure(bindMsg));
```
vietj, line 53: don't we need a return statement after this failure ?

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SocketConnectionBase.java
vietj: is this related to the actual bug reported ?

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SocketConnectionBase.java
```diff
@@ -161,6 +161,14 @@ private void checkPending() {
...
+            inflight--;
+            queryCmd.fail(new NoStackTraceThrowable(msg));
+            return;
```
vietj, line 170: why is return needed here ?

vietj:
I can see this tested for MySQL, does it apply for all databases ?

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SocketConnectionBase.java
BillyYccc: we need this which was overlooked before, because the parameters of new execution need to be prepared for cached one-shot prepared queries otherwise the parameters would be prepared only once for the first time and never be prepared later.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/ExtendedBatchQueryCommandCodec.java
BillyYccc: fixed in https://github.com/eclipse-vertx/vertx-sql-client/pull/720/commits/612738b570cb98434d2bc223c793f2dfc81b553b

BillyYccc:
Adding a `prepare` procedure in the common package is what we left over before and we need to add this which will be useful for all databases(especially for postgres because of the data type checking).

Changes in the MySQL is also required because we need to separate `prepare` and MySQL `binding` because in the past the binding is in `prepare` and will cause the MySQL client sending wrong binded data type to the server when executing.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SocketConnectionBase.java
vietj: It seems this patch of connection base seems more general than this specific issue

BillyYccc:
superseded by changes in https://github.com/eclipse-vertx/vertx-sql-client/pull/727

Issue #726, opened 2020-08-10
Command pipeline execution should not be reentrant


vietj:
The command pipeline execution can reenter the internal `checkPending` method when a synchronous failure in the `checkPending` (such as a prepared query validation error) triggers a new command. The execution remains correct, i.e the new command will not be executed before other commands but this execution should not be reentrant and the command pipeline execution should be performed by the execution stack frame exclusively.

Issue #725, opened 2020-08-10
Prepared query command execution does not validate query parameters when obtained from the cache


vietj:
Prepared query execution does not validate query parameters when obtained from the cache. They are only validated for the query populating the cache.

PR #727, opened 2020-08-10
Command execution pipeline fixes - see #725 - see #726

vietj:


vietj:
@BillyYccc here are 2 bug fixes related to #719 and #720 - the tests pass for all databases but not MySQL, I suspect this should is related to the other changes you did in #720 and I would be glad if you could fix them for this PR and then we will re evaluate #720 to see if there is more to fix - can you fix the failing MySQL tests with new commits appended to this PR ?

BillyYccc:
I have ported the changes here to make MySQL work now.

vietj:
thanks @BillyYccc is that enough for fixing #719 ?

PR #732, opened 2020-08-12
(Reproducer)[PostgreSQL] Prepared Statement Name Overflow (StringLongSequence)

julianladisch:
StringLongSequence generates the name for the PostgreSQL prepared
statement. StringLongSequence returns the hex numbers 0000000 to 000FFFF,
then starts from 0000000 again.

Using 0000000 a second time on the same connection causes this exception:

io.vertx.pgclient.PgException:
{ "message": "prepared statement \"0000000\" already exists",
  "severity": "ERROR", "code": "42P05", "file": "prepare.c",
  "line": "475", "routine": "StorePreparedStatement"
}


julianladisch:
Reopening because upstream doesn't have applied the fix.

PR #733, opened 2020-08-13
Close prepared statements in SqlClientExamples

julianladisch:
Adding a PreparedStatement.close call to all examples that use
prepared statements helps developers by providing a blueprint
that deallocates the resources taken by the prepared statement.


vietj:
it actually is documented that the PS should be closed, it should I think be emphasized that the PS should be closed in the documentation with a warning in the Prepared Queries section

gavinking:
Rmmmm, wait, I don't think we ever close prepared statements in HR. Do we need to?

BillyYccc:
You don't need to close it if you're using one-shot prepared queries.

gavinking:
OK, thanks 🙏

vietj:
indeed, if you are using one-shot prepared queries, you only need to
release the connection when you are done with it.

the implementation will take care of preparing and caching statements for
you.

PR #735, opened 2020-08-13
Use guard clauses for readability in SqlClientExamples

julianladisch:
Replace

```
method(ar -> {
  if (ar.succeeded()) {
    // use
    // ar.result()
    // to
    // do
    // something
  } else {
    // error handling
  }
});
```

by

```
method(ar -> {
  if (ar.failed()) {
    // error handling
    return;
  }
  // use
  // ar.result()
  // to
  // do
  // something
});
```

This improves readability because
* the main task "to do something" has reduced indentation
  highlighting the main execution flow
* the error handling is moved up to the method call it belongs to

References:
https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html
https://refactoring.guru/replace-nested-conditional-with-guard-clauses
https://stackoverflow.com/questions/4887212/shall-i-use-guard-clause-and-try-to-avoid-else-clause

### DIFF
--- a/vertx-sql-client/src/main/java/examples/SqlClientExamples.java
+++ b/vertx-sql-client/src/main/java/examples/SqlClientExamples.java
@@ -19,7 +19,6 @@ package examples;
 import io.vertx.core.Vertx;
 import io.vertx.sqlclient.Cursor;
 import io.vertx.sqlclient.Pool;
-import io.vertx.sqlclient.PoolOptions;
 import io.vertx.sqlclient.PreparedStatement;
 import io.vertx.sqlclient.Row;
 import io.vertx.sqlclient.RowSet;
@@ -39,12 +38,12 @@ public class SqlClientExamples {
     client
       .query("SELECT * FROM users WHERE id='julien'")
       .execute(ar -> {
-      if (ar.succeeded()) {
-        RowSet<Row> result = ar.result();
-        System.out.println("Got " + result.size() + " rows ");
-      } else {
+      if (ar.failed()) {
         System.out.println("Failure: " + ar.cause().getMessage());
+        return;
       }
+      RowSet<Row> result = ar.result();
+      System.out.println("Got " + result.size() + " rows ");
     });
   }
 
@@ -53,12 +52,12 @@ public class SqlClientExamples {
     client
       .preparedQuery("SELECT * FROM users WHERE id=$1")
       .execute(Tuple.of("julien"), ar -> {
-      if (ar.succeeded()) {
-        RowSet<Row> rows = ar.result();
-        System.out.println("Got " + rows.size() + " rows ");
-      } else {
+      if (ar.failed()) {
         System.out.println("Failure: " + ar.cause().getMessage());
+        return;
       }
+      RowSet<Row> rows = ar.result();
+      System.out.println("Got " + rows.size() + " rows ");
     });
   }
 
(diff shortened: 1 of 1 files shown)


gavinking:
Counterpoint: this is a code style I actually strongly dislike because it means the compiler catches fewer errors in the flow of control. By not clearly indicating to the compiler your intent that these two branches be mutually exclusive, you open up the possibility that you can "fall out" of the error-handling block and start executing the regular code, and the compiler won't catch that.

In general, the compiler will catch more errors if the function exit points are at a *higher* level of nesting. If you're worried that your function has two many levels of nesting, it's almost surely because your function is doing two many things, and Extract Method is your friend.

julianladisch:
@gavinking Can you give an example where the compiler catches an error and this code style change prevents it? Can you point us to some Java book or Java website where this is discussed?

gavinking:
Examples are quite trivial to construct:

```java
if ( error!=null ) {
    switch (error.code) {
        case TYPE1: return something();
        case TYPE2: return somethingElse();
        case TYPE3: return somethingDifferent();
        //....
        case TYPE1001: return phew();
        //ooops forgot about TYPE 503
    }
}

runStatement("drop table universe");
return success();
```

vs:

```java
if ( error!=null ) {
    switch (error.code) {
        case TYPE1: return something();
        case TYPE2: return somethingElse();
        case TYPE3: return somethingDifferent();
        //....
        case TYPE1001: return phew();
        //ooops forgot about TYPE 503
    }
}
else {
    runSqlStatement("drop table universe");
    return success();
}
```

The second code example results in a compile time error. The first code example happily deletes your data because you forgot a `case` of your `switch`.

> Can you point us to some Java book or Java website where this is discussed?

Haha, naw, sadly Java people are pretty allergic to thinking and writing about type safety.

julianladisch:
Your example is for a method with a non-void return type. This pull request is about lambdas with void return type. Can you provide us with an example for a `Handler<AsyncResult>` lambda?
IDEs and static code analyzers issue a warning or a failure for a case statement without default, can you pick a different type of coding error?

julianladisch:
I do not request to replace _all_ if-else-clauses by guard clauses. This needs to be decided on a case-by-case basis. This is an example from this pull request where I intentionally didn't replace the if-else-clause because it is more easy to understand than the guard clause:

                  if (ar3.succeeded()) {
                    System.out.println("Transaction succeeded");
                  } else {
                    System.out.println("Transaction failed " + ar3.cause().getMessage());
                  }
                  // Return the connection to the pool
                  conn.close();

gavinking:
Well in the case of a `void` method, OK, true, the compiler doesn't help you *as much*, but I still think that _even in that case_, guard clauses are almost as bad. It's extremely easy when editing or merging code to accidentally eliminate one of those `return` statements by mistake, and the compiler will *never* catch a missing `return` in a `void` method.

The great thing about `else {}` is that it makes completely clear and explicit which branches of code are mutually exclusive. And it's almost impossible to delete an `else` by accident, even when merging code, because you have matched braces and indentation.

gavinking:
I mean, I don't really want to get this picky, but I just don't see how it's possibly an improvement to change:

```java
public void usingConnections01(Vertx vertx, Pool pool) {

    pool.getConnection(ar1 -> {
      if (ar1.succeeded()) {

        connection
          .query("SELECT * FROM users WHERE id='andy'")
          .execute(ar2 -> {
          if (ar1.succeeded()) {
            connection
              .query("SELECT * FROM users WHERE id='julien'")
              .execute(ar3 -> {
              // Do something with rows and return the connection to the pool
              connection.close();
            });
          } else {
            // Return the connection to the pool
            connection.close();
          }
        });
      }
    });
  }
```

To:

```java
  public void usingConnections01(Vertx vertx, Pool pool) {

    pool.getConnection(ar1 -> {
      if (ar1.failed()) {
        // error handling
        return;
      }

      connection
        .query("SELECT * FROM users WHERE id='julien'")
        .execute(ar2 -> {
        if (ar2.failed()) {
          // Return the connection to the pool
          connection.close();
          return;
        }
        connection
          .query("SELECT * FROM users WHERE id='paulo'")
          .execute(ar3 -> {
          // Do something with rows and return the connection to the pool
          connection.close();
        });
      });
    });
  }
```

The second code example can be broken by removing two tokens, and you won't notice until runtime, assuming you even have tests for the exceptional code path.

vietj:
thanks we could eventually use the new onSuccess / onFailure instead

PR #737, opened 2020-08-17
Unsigned long HexSequence for name of PgPreparedStatement

julianladisch:
Replace StringLongSequence by HexSequence
based on an unsigned long counter
to avoid duplicate prepared statement name in
PgPreparedStatement.

StringLongSequence wraps around to 0000000 after 000FFFF.
HexSequence wraps around to 0000000 after 000FFFFFFFFFFFFFFFF.

Example:

```
conn.prepare("SELECT 1", ctx.asyncAssertSuccess(ps1 -> {
  for (int i=1; i<=0x10000; i++) {
    conn.prepare("SELECT 2", ctx.asyncAssertSuccess(ps2 -> {
      ps2.query().execute(ctx.asyncAssertSuccess(done -> ps2.close()));
    }));
  }
}));
```

This example has a first prepared statement that is not closed.
0x10000 other prepared statements are created and closed, they
use the names 0000001 to 000FFFF and finally 0000000. The
0000000 is still in use by the first prepared statement and
causes this exception:

"message": "prepared statement "0000000" already exists",
"severity": "ERROR", "code": "42P05", "file": "prepare.c",
"line": "475", "routine": "StorePreparedStatement"

The new HexSequence uses the full number range of an
unsigned long making it unlikely that a name still is in use.

Fixes #732.


julianladisch:
@BillyYccc @vietj What needs to be changed to get this merged?

BillyYccc:
I think It needs to be reviewed first.

vietj:
to simplify this, I think instead we should keep the sequence as a simple counter and when the statement is prepared then this counter is encoded into a byte[] with the same encoding used

julianladisch:
Changed to use a simple unsigned long counter.

vietj:
I will review the PR

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/HexSequence.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  /**
+   * @return a copy of the next hex value, terminated by a zero byte.
+   */
+  public byte[] next() {
+    String hex = Long.toUnsignedString(i, 16).toUpperCase(Locale.ROOT);
+    i++;
+    switch (hex.length()) {
+    case 1: return ("000000" + hex + "\0").getBytes(StandardCharsets.US_ASCII);
```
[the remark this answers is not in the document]
vietj: for the hex conversion we can rely on branchless conversion https://gist.github.com/stryku/b8177a66f79540f912ca8b7909ffccda

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/PrepareStatementCommandCodec.java
```diff
@@ -38,7 +38,7 @@ void encode(PgEncoder encoder) {
       statement = encoder.nextStatementName();
     } else {
       // Use unnamed prepared statements that don't need to be closed
-      statement = 0L;
+      statement = new byte[] { 0 };
```
vietj, line 41: we should use a constant for this

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/StringLongSequence.java
```diff
@@ -21,6 +21,9 @@
 
   private short count;
 
+  /**
```
vietj, line 24: is that class still used ?

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/HexSequence.java
```diff
@@ -0,0 +1,86 @@
...
+      return (byte)('0' + c);
+    } else {
+      return (byte)('A' + c - 10);
+    }
+  }
+
+  /**
+   * A copy of the next hex value, terminated by a zero byte.
+   */
+  public byte[] next() {
```
vietj, line 68: @franz1981 do you mind having a look ?

franz1981: correctness or perf?

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/PrepareStatementCommandCodec.java
julianladisch: Replaced by EMPTY_STRING constant.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/StringLongSequence.java
julianladisch: No, removed.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/HexSequence.java
vietj: correctness, it's not critical

vietj: but it does not mean it should be slower than necessary :-)

vietj:
it's on my todo list :-)

PR #738, opened 2020-08-17
Add --release compiler flag when building on JDK 9+ to avoid bytecode incompat

aguibert:
Fixes #736

PR #739, opened 2020-08-17
Backport: Add --release compiler flag when building on JDK 9+ to avoid bytecode incompat

aguibert:
Backport of #738

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,6 +101,28 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  
+  <profiles>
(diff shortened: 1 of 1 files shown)


vietj:
maybe this should be in vert-parent ?

aguibert:
@vietj yes I think that would be ideal, I am not sure how to verify this change doesn't break any of the projects that depend on vertx-parent though

aguibert:
Cancelling this PR -- fixed in vertx-parent as Julien suggested

PR #750, opened 2020-08-31
Column descriptor

### LINKED ISSUE #657: SqlResult.columnTypes()
In Hibernate Reactive we need a way to obtain information about the type of a column in a `RowSet`.

Please [see my comment here](https://github.com/hibernate/hibernate-reactive/pull/180/files/b505f9530eddf59dfc173f24fa50689b0bb2da3a#r429576118).

vietj:
Close #657


vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/datatype/DataType.java
```diff
@@ -8,34 +8,35 @@
...
+  FLOAT(ColumnDefinition.ColumnType.MYSQL_TYPE_FLOAT, Float.class, Float.class, JDBCType.REAL),
+  NUMERIC(ColumnDefinition.ColumnType.MYSQL_TYPE_NEWDECIMAL, Numeric.class, Numeric.class, JDBCType.NUMERIC), // DECIMAL
+  STRING(ColumnDefinition.ColumnType.MYSQL_TYPE_STRING, Buffer.class, String.class, JDBCType.VARCHAR), // CHAR, BINARY
+  VARSTRING(ColumnDefinition.ColumnType.MYSQL_TYPE_VAR_STRING, Buffer.class, String.class, JDBCType.VARCHAR), //VARCHAR, VARBINARY
+  TINYBLOB(ColumnDefinition.ColumnType.MYSQL_TYPE_TINY_BLOB, Buffer.class, String.class, JDBCType.BINARY),
+  BLOB(ColumnDefinition.ColumnType.MYSQL_TYPE_BLOB, Buffer.class, String.class, JDBCType.BINARY),
+  MEDIUMBLOB(ColumnDefinition.ColumnType.MYSQL_TYPE_MEDIUM_BLOB, Buffer.class, String.class, JDBCType.BINARY),
+  LONGBLOB(ColumnDefinition.ColumnType.MYSQL_TYPE_LONG_BLOB, Buffer.class, String.class, JDBCType.BINARY),
+  DATE(ColumnDefinition.ColumnType.MYSQL_TYPE_DATE, LocalDate.class, LocalDate.class, JDBCType.DATE),
+  TIME(ColumnDefinition.ColumnType.MYSQL_TYPE_TIME, Duration.class, Duration.class, JDBCType.DATE),
```
BillyYccc, line 32: `TIME` should be mapped to `JDBCType.TIME`

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/datatype/DataType.java
```diff
@@ -8,34 +8,35 @@
...
+  DATETIME(ColumnDefinition.ColumnType.MYSQL_TYPE_DATETIME, LocalDateTime.class, LocalDateTime.class, JDBCType.DATE),
```
BillyYccc, line 33: `DATETIME` should be mapped to `JDBCType.TIMESTAMP`

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/datatype/DataType.java
```diff
@@ -8,34 +8,35 @@
...
+  YEAR(ColumnDefinition.ColumnType.MYSQL_TYPE_YEAR, Short.class, Short.class, JDBCType.DATE),
```
BillyYccc, line 34: `YEAR` should be mapped to `JDBCType.SMALLINT`

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/datatype/DataType.java
```diff
@@ -8,34 +8,35 @@
...
+  TIMESTAMP(ColumnDefinition.ColumnType.MYSQL_TYPE_TIMESTAMP, LocalDateTime.class, LocalDateTime.class, JDBCType.DATE),
```
BillyYccc, line 35: `TIMESTAMP` should be mapped to `JDBCType.TIMESTAMP`

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/datatype/DataType.java
```diff
@@ -8,34 +8,35 @@
...
+  BIT(ColumnDefinition.ColumnType.MYSQL_TYPE_BIT, Long.class, Long.class, JDBCType.OTHER),
```
BillyYccc, line 36: `BIT` should be mapped to `JDBCType.BIT`

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/datatype/DataType.java
```diff
@@ -8,34 +8,35 @@
...
+  INT8(ColumnDefinition.ColumnType.MYSQL_TYPE_LONGLONG, Long.class, Long.class, JDBCType.BIGINT),
+  DOUBLE(ColumnDefinition.ColumnType.MYSQL_TYPE_DOUBLE, Double.class, Double.class, JDBCType.DOUBLE),
+  FLOAT(ColumnDefinition.ColumnType.MYSQL_TYPE_FLOAT, Float.class, Float.class, JDBCType.REAL),
+  NUMERIC(ColumnDefinition.ColumnType.MYSQL_TYPE_NEWDECIMAL, Numeric.class, Numeric.class, JDBCType.NUMERIC), // DECIMAL
+  STRING(ColumnDefinition.ColumnType.MYSQL_TYPE_STRING, Buffer.class, String.class, JDBCType.VARCHAR), // CHAR, BINARY
+  VARSTRING(ColumnDefinition.ColumnType.MYSQL_TYPE_VAR_STRING, Buffer.class, String.class, JDBCType.VARCHAR), //VARCHAR, VARBINARY
+  TINYBLOB(ColumnDefinition.ColumnType.MYSQL_TYPE_TINY_BLOB, Buffer.class, String.class, JDBCType.BINARY),
+  BLOB(ColumnDefinition.ColumnType.MYSQL_TYPE_BLOB, Buffer.class, String.class, JDBCType.BINARY),
+  MEDIUMBLOB(ColumnDefinition.ColumnType.MYSQL_TYPE_MEDIUM_BLOB, Buffer.class, String.class, JDBCType.BINARY),
+  LONGBLOB(ColumnDefinition.ColumnType.MYSQL_TYPE_LONG_BLOB, Buffer.class, String.class, JDBCType.BINARY),
```
BillyYccc, line 30: we should use `JDBCType.BLOB` for MySQL `BLOB` data types

BillyYccc (commented):
I wonder if we could build an implementation of `java.sql.SQLType` and include all data types of `java.sql.Types` instead of directly using the `java.sql.JDBCType`? I think it would be better for enriching the data types such as `GEOMETRY` and `JSON` so we don't have to represent them as `OTHER`.

aguibert (approved):
LGTM, will try to contribute DB2 to this branch by end of week

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
```diff
@@ -46,85 +48,85 @@
...
+  TIME(1083, true, LocalTime.class, JDBCType.DATE),
+  TIME_ARRAY(1183, true, LocalTime[].class, JDBCType.DATE),
+  TIMETZ(1266, true, OffsetTime.class, JDBCType.DATE),
+  TIMETZ_ARRAY(1270, true, OffsetTime[].class, JDBCType.DATE),
+  TIMESTAMP(1114, true, LocalDateTime.class, JDBCType.DATE),
+  TIMESTAMP_ARRAY(1115, true, LocalDateTime[].class, JDBCType.DATE),
+  TIMESTAMPTZ(1184, true, OffsetDateTime.class, JDBCType.DATE),
+  TIMESTAMPTZ_ARRAY(1185, true, OffsetDateTime[].class, JDBCType.DATE),
+  INTERVAL(1186, true, Interval.class, JDBCType.DATE),
+  INTERVAL_ARRAY(1187, true, Interval[].class, JDBCType.DATE),
```
gavinking, line 92: Some of the above mappings don't look right to me. `TIME` should be `TIME`, `TIMESTAMP`  should be `TIMESTAMP`, and `TIMESTAMPTZ` should probably be `TIMESTAMP_WITH_TIMEZONE` (but I'm not sure about that, we need to check).

And `INTERVAL` I'm not sure about but I'm pretty sure it's not a `DATE`.

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
```diff
@@ -46,85 +48,85 @@
...
+  VARBIT(1562, true, Object.class, JDBCType.OTHER),
+  VARBIT_ARRAY(1563, true, Object[].class, JDBCType.OTHER),
+  CHAR(18, true, String.class, JDBCType.VARCHAR),
+  CHAR_ARRAY(1002, true, String[].class, JDBCType.VARCHAR),
+  VARCHAR(1043, true, String.class, JDBCType.VARCHAR),
+  VARCHAR_ARRAY(1015, true, String[].class, JDBCType.VARCHAR),
+  BPCHAR(1042, true, String.class, JDBCType.VARCHAR),
+  BPCHAR_ARRAY(1014, true, String[].class, JDBCType.VARCHAR),
+  TEXT(25, true, String.class, JDBCType.VARCHAR),
+  TEXT_ARRAY(1009, true, String[].class, JDBCType.VARCHAR),
```
gavinking, line 78: Perhaps `TEXT` could be `LONGVARCHAR`, I'm not sure.

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
gavinking: And `TIMETZ` is `TIME_WITH_TIMEZONE`.

vietj:
I don't think it buys much and add complexity.

vertx-sql-client/src/main/java/io/vertx/sqlclient/desc/ColumnDescriptor.java
```diff
@@ -0,0 +1,30 @@
...
+package io.vertx.sqlclient.desc;
+
+import io.vertx.codegen.annotations.VertxGen;
+
+import java.sql.JDBCType;
+
+@VertxGen
+public interface ColumnDescriptor {
+
+  String name();
```
aguibert, line 26: I don't think it's necessary to have `name()` (or a `ColumnDescriptor` at all for that matter) since the column name is already available through `SqlResult#columnNames()`. Instead, could we just add a `SqlResult#jdbcTypes()` method?

vietj:
we might add more metadata to this interface later like a Class<?> that
would be the class of the mapped type.

gavinking:
Or whether the column is nullable. Or the native column type name.

vietj:
I will take in account all comments done on incorrect mappings and get this in 3.9.3 hopefully and then port to master

vertx-db2-client/src/test/java/io/vertx/db2client/tck/DB2BinaryDataTypeDecodeTest.java
```diff
@@ -61,7 +65,31 @@ public void testDouble(TestContext ctx) {
...
+  @Override
+  public void testDecimal(TestContext ctx) {
+    // Override to expecting JDBCType.DECIMAL instead of NUMERIC
+    testDecodeGeneric(ctx, "test_decimal", Numeric.class, JDBCType.DECIMAL, Numeric.parse("12345"));
+  }
+  
+  @Override
+  public void testTime(TestContext ctx) {
+    // Override to expecting JDBCType.TIME instead of DATE
+    testDecodeGeneric(ctx, "test_time", LocalTime.class, JDBCType.TIME, LocalTime.of(18, 45, 2));
```
aguibert, line 92: I'm still a bit unclear on how the `JDBCType` will be used. Is it OK that I'm overriding some of the expected JDBCTypes for DB2 here? Or is it important that all DBs are consistent with one another for this use case? CC @gavinking

gavinking: Well we’re going to use it to decide what kind of Java object to read from that column of the result set. For example, if you return `Types.VARCHAR`, we will read strings.

aguibert: So for `testChar`, in DB2 the column is defined as `CHAR(6)`. To me it seemed correct that we classify that as `JDBCType.CHAR`, but since a char[6] is basically the same thing as `VARCHAR/String`, should I have it return `JDBCType.VARCHAR` for a `CHAR(6)` column?

gavinking: I would just try to copy what the JDBC driver does.

vietj: @gavinking what can we assume among these ?

- same java type => the same SQL type?
- same java type <= the same SQL type?
- same java type <=> the same SQL type?
- no relationship ?

gavinking: No it's nonsimple. If you look at the appendices of the JDBC spec you'll see that it's a many-to-many mapping between Java types and SQL spec types. And then the Hibernate built-in types only loosely follow that mapping.

But that's not what we're trying to capture here. What we're going for here is a mapping from database types to equivalent ANSI SQL types. Now that is also nonsimple, but it's *simpler*, and we can look at what mapping the JDBC drivers use for inspiration.

gavinking: (But of course Hibernate does expect some consistency: if you return `TIMESTAMP`, it's going to expect to get a `LocalDateTime` when it reads the result set, and if you return `TIMESTAMP_WITH_TIMEZONE`, it's going to expect to get an `OffsetDateTime`.)

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/datatype/DataType.java
```diff
@@ -8,34 +8,35 @@
...
-  NULL(ColumnDefinition.ColumnType.MYSQL_TYPE_NULL, Object.class, Object.class), // useful for mariadb prepare statement response
-  UNBIND(-1, Object.class, Object.class); // useful for binding param values
+  INT1(ColumnDefinition.ColumnType.MYSQL_TYPE_TINY, Byte.class, Byte.class, JDBCType.TINYINT),
+  INT2(ColumnDefinition.ColumnType.MYSQL_TYPE_SHORT, Short.class, Short.class, JDBCType.SMALLINT),
+  INT3(ColumnDefinition.ColumnType.MYSQL_TYPE_INT24, Integer.class, Integer.class, JDBCType.INTEGER),
+  INT4(ColumnDefinition.ColumnType.MYSQL_TYPE_LONG, Integer.class, Integer.class, JDBCType.INTEGER),
+  INT8(ColumnDefinition.ColumnType.MYSQL_TYPE_LONGLONG, Long.class, Long.class, JDBCType.BIGINT),
+  DOUBLE(ColumnDefinition.ColumnType.MYSQL_TYPE_DOUBLE, Double.class, Double.class, JDBCType.DOUBLE),
+  FLOAT(ColumnDefinition.ColumnType.MYSQL_TYPE_FLOAT, Float.class, Float.class, JDBCType.REAL),
+  NUMERIC(ColumnDefinition.ColumnType.MYSQL_TYPE_NEWDECIMAL, Numeric.class, Numeric.class, JDBCType.NUMERIC), // DECIMAL
```
gavinking, line 24: @vietj we should check what the JDBC driver does here. the whole `NUMERIC` vs `DECIMAL` thing has always been a source of problems since the drivers are pretty inconsistent here.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/datatype/DataType.java
```diff
@@ -8,34 +8,35 @@
...
+  STRING(ColumnDefinition.ColumnType.MYSQL_TYPE_STRING, Buffer.class, String.class, JDBCType.VARCHAR), // CHAR, BINARY
```
gavinking, line 25: I'm not sure, but perhaps this should be `CHAR`, though Hibernate wouldn't care much.

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
```diff
@@ -46,85 +47,85 @@
...
-  HSTORE(33670, true, Object.class),
-  OID(26, true, Object.class),
-  OID_ARRAY(1028, true, Object[].class),
-  VOID(2278, true, Object.class),
-  UNKNOWN(705, false, String.class),
-  TS_VECTOR(3614, false, String.class),
-  TS_VECTOR_ARRAY(3643, false, String[].class),
-  TS_QUERY(3615, false,  String.class),
-  TS_QUERY_ARRAY(3645, false,  String[].class);
+  BOOL(16, true, Boolean.class, JDBCType.BIT),
```
gavinking, line 50: I believe this should be `BOOLEAN`.

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
```diff
@@ -46,85 +47,85 @@
...
+  BOOL_ARRAY(1000, true, Boolean[].class, JDBCType.BIT),
```
gavinking, line 51: I suppose this should be `ARRAY`.

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
```diff
@@ -46,85 +47,85 @@
...
+  INT2_ARRAY(1005, true, Short[].class, Number[].class, JDBCType.SMALLINT),
+  INT4(23, true, Integer.class, Number.class, JDBCType.INTEGER),
+  INT4_ARRAY(1007, true, Integer[].class, Number[].class, JDBCType.INTEGER),
+  INT8(20, true, Long.class, Number.class, JDBCType.BIGINT),
+  INT8_ARRAY(1016, true, Long[].class, Number[].class, JDBCType.BIGINT),
+  FLOAT4(700, true, Float.class, Number.class, JDBCType.REAL),
+  FLOAT4_ARRAY(1021, true, Float[].class, Number[].class, JDBCType.REAL),
+  FLOAT8(701, true, Double.class, Number.class, JDBCType.DOUBLE),
+  FLOAT8_ARRAY(1022, true, Double[].class, Number[].class, JDBCType.DOUBLE),
+  NUMERIC(1700, false, Numeric.class, Number.class, JDBCType.NUMERIC),
```
gavinking, line 62: Again we should check what the JDBC driver says about numeric vs decimal.

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
```diff
@@ -46,85 +47,85 @@
...
+  NUMERIC_ARRAY(1231, false, Numeric[].class, Number[].class, JDBCType.NUMERIC),
+  MONEY(790, true, Object.class, null),
+  MONEY_ARRAY(791, true, Object[].class, null),
+  BIT(1560, true, Object.class, JDBCType.OTHER),
```
gavinking, line 66: Shouldn't this be `BIT`?

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
```diff
@@ -46,85 +47,85 @@
...
+  BIT_ARRAY(1561, true, Object[].class, JDBCType.OTHER),
+  VARBIT(1562, true, Object.class, JDBCType.OTHER),
+  VARBIT_ARRAY(1563, true, Object[].class, JDBCType.OTHER),
+  CHAR(18, true, String.class, JDBCType.VARCHAR),
```
gavinking, line 70: Should be `CHAR`.

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
```diff
@@ -46,85 +47,85 @@
...
+  FLOAT4_ARRAY(1021, true, Float[].class, Number[].class, JDBCType.REAL),
+  FLOAT8(701, true, Double.class, Number.class, JDBCType.DOUBLE),
+  FLOAT8_ARRAY(1022, true, Double[].class, Number[].class, JDBCType.DOUBLE),
+  NUMERIC(1700, false, Numeric.class, Number.class, JDBCType.NUMERIC),
+  NUMERIC_ARRAY(1231, false, Numeric[].class, Number[].class, JDBCType.NUMERIC),
+  MONEY(790, true, Object.class, null),
+  MONEY_ARRAY(791, true, Object[].class, null),
+  BIT(1560, true, Object.class, JDBCType.OTHER),
+  BIT_ARRAY(1561, true, Object[].class, JDBCType.OTHER),
+  VARBIT(1562, true, Object.class, JDBCType.OTHER),
```
gavinking, line 68: Also `BIT`, if i'm not mistaken.

gavinking (changes requested):

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
vietj: right I missed that

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
vietj: I will have a look

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
vietj: so if that's an array, how do you get the array component ? I'm assuming in case of array that we return the component type of the array.

we could add on ColumnDescriptor an isArray or isMultivalued boolean

vertx-db2-client/src/test/java/io/vertx/db2client/tck/DB2BinaryDataTypeDecodeTest.java
vietj: ok

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
gavinking: TBH I have never used a SQL array type in my life, so i'm just not sure, but it would not surprise me if there was no good way, other than calling `Array.getBaseType()`.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/datatype/DataType.java
vietj: @BillyYccc can you check that against the JDBC driver ?

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
vietj: I added a ColumnDescriptor.isArray method to check whether it's an array and the SQL type will be in this case the component type of the array.

gavinking: OK

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
vietj: I checked pg and it is `JDBCType.NUMERIC`

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/datatype/DataType.java
vietj: it is DECIMAL in this case and we should correct this, see https://github.com/mysql/mysql-connector-j/blob/dbd865c9fc68701efc70197ea1e65002520bed91/src/main/protocol-impl/java/com/mysql/cj/protocol/a/NativeProtocol.java#L1423 (both constants unsigned and signed then reference the Types.DECIMAL = 3 constant)

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/datatype/DataType.java
vietj: let's keep it this way for now and we'll correct it if that's needed

vietj:
I will port this to master

Issue #608, opened 2020-05-07
[MSSQL]datatype support


BillyYccc:
currently we only support a small set of all the datatypes, we should expand the support and improve the current data type implmentation(such as `NUMERIC`).

BillyYccc:
## Data type support checklist:

Note some data types have both fixed-length & vairable-length column, we need to support them all.

#### Numeric

- [ ] NUMERIC & DECIMAL (supported but do need some changes) (**HIGH PRIORITY**)
- [x] BIT
- [x] float & real
- [x] int, bigint, smallint, tinyint
- [ ] money & smallmoney

#### String & Binary

- [ ] binary & varbinary
- [x] char & varchar
- [ ] nchar & nvarchar (**HIGH PRIORITY**)
- [ ] ntext, text & image

#### date & time

- [x] date
- [ ] datetime
- [ ] datetime2 (**HIGH PRIORITY**)
- [ ] datetimeoffset (**HIGH PRIORITY**)
- [ ] smalldatetime
- [x] time

#### other

xml, GUID, spatial data types, etc...

gavinking:
I agree that `numeric`/`decimal`, and `datetime2` are very high priority.

After that I would vote for `image`, `text`, and `varbinary`.

tsegismont:
@BillyYccc about _NUMERIC & DECIMAL (supported but do need some changes)_: which changes are needed? I couldn't find any comment or todos in the code

BillyYccc:
I remember the scales are not handled in a right way.

tsegismont:
@BillyYccc would you mind filing a separate issue for this please?

tsegismont:
OK I'll take care of this.

Le ven. 26 mars 2021 à 15:44, Billy Yuan ***@***.***> a
écrit :

JonasTaulienSolutions:
Hello, just wanted to let you know, that we had to switch to the (legacy) jdbc-client because `varchar(max)` is currently not supported.

tsegismont:
@JonasTaulienSolutions was `varchar(max)` the only one missing for your project?

JonasTaulienSolutions:
Currently, its only `varchar(max)` that is missing for us. 
But in the next months we will probably need support for some more. If this happens I will write another message here :)

Yaytay:
We have a lot of GUID columns in our data, would it take much to add support for them?
Thanks.

tsegismont:
[...]

PR #751, opened 2020-08-31
Support for datetime2 data type

tsegismont:
See #608

### DIFF
--- a/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedQueryCommandCodec.java
+++ b/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedQueryCommandCodec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2021 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -25,6 +25,7 @@ import io.vertx.sqlclient.data.Numeric;
 import io.vertx.sqlclient.impl.command.ExtendedQueryCommand;
 
 import java.time.LocalDate;
+import java.time.LocalDateTime;
 import java.time.LocalTime;
 import java.time.temporal.ChronoUnit;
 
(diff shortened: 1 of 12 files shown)


vietj (commented):
I can see many reformatting changes because your IDE is configured differently for rewritten (type == String.class) as `(type==String.class)` can you remove those un-necessary changes ?

tsegismont:
My IDE settings are correct, those lines were re-formatted because IDEA formats changed method bodies. It's just a few of them in a couple of methods.

tsegismont:
@vietj anything else beyond formatting?

vietj:
@tsegismont you seem to have peculiar IDE settings. Most vertx code base uses ` == ` , so that is a concern because the code you will be committing will always conflict and rewrite the code base, other committers (like me) will have their IDE reformat the other and it will be very confusing, so you should change your IDE settings to match this rule.

tsegismont:
@vietj I use IDEA with default settings + the plugin for `.editorconfig` file. Not sure why but spaces around `==` got unchecked. I've updated my config and the PR.

Anything else beyond formatting?

vietj (approved):

vietj:
can you rebase @tsegismont ?

tsegismont:
@vietj just rebased on `master`. Can I merge when the GH checks are all green or do you want to take another look?

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedQueryCommandCodec.java
```diff
@@ -207,6 +208,8 @@ private void encodeParamValue(ByteBuf payload, Object value) {
       encodeDateNParameter(payload, (LocalDate) value);
     } else if (value instanceof LocalTime) {
       encodeTimeNParameter(payload, (LocalTime) value, (byte) 6);
+    } else if (value instanceof LocalDateTime) {
+      encodeDateTimeNParameter(payload, (LocalDateTime) value, (byte) 6);
```
BillyYccc, line 212: Any opinions about the scale? I'm not sure if I ever got it right to hardcode it as 6 to not truncating any nano seconds in the `LocalTime` parameter.

tsegismont: I filed #752 to make sure we analyze and take any action needed.

PR #755, opened 2020-09-03
Row to JSON conversion

### LINKED ISSUE #754: Row to JSON conversion
To allow anemic use cases using JSON we shall add a `toJson()` method on `Row`.

vietj:
close #754

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/Utils.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/Utils.java
@@ -1,0 +1,42 @@
+package io.vertx.sqlclient.impl;
+
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+
+import java.lang.reflect.Array;
+import java.util.ArrayList;
+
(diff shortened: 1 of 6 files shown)


vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/Utils.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.lang.reflect.Array;
+import java.util.ArrayList;
+
+import static io.vertx.sqlclient.Tuple.JSON_NULL;
+
+/**
+ * Bunch of generic utils.
+ */
+public final class Utils {
```
jponge, line 15: You might as well use an interface and a static method. I know we old Java-ers don't think of interfaces for static methods in the first place, but you can actually do it 😸

jponge (approved):

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/Utils.java
vietj: I don't have any preferences, some use a Java enum as well

I think the Java language is missing the keyword _singletion_ :-)

PR #782, opened 2020-09-26
Coerce row values to string for Row#getString

aguibert:
Fixes a confusing scenario that a user reported here: https://github.com/quarkusio/quarkus/issues/11035


vietj:
I am not sure this is something we want to do for free

vietj:
my point is that people should know the type of the row they are retrieving and they should use the correct getter otherwise they should use `getValue().toString()` instead

julianladisch:
I agree with Julien.
The javadoc should be extended to make this clear: #802

julianladisch:
`getValue().toString()` may cause NPE.
Should we add `getValueAsString` that returns null for null and toString() otherwise?

PR #791, opened 2020-10-22
Fix method summary of Transaction.completion javadoc

julianladisch:
Fix method summary of Transaction.completion javadoc

* "and" is missing in the method summary of Transaction.completion on
  https://vertx-web-site.github.io/docs/apidocs/io/vertx/sqlclient/Transaction.html

* "rollbacks" should be "rolls back".

Old text:

"Return the transaction completion Future that succeeds when the transaction
commits fails with TransactionRollbackException when the transaction
**rollbacks**"

New text:

"Return the transaction completion Future that succeeds when the transaction
commits **and** fails with TransactionRollbackException when the transaction
**rolls back**"

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
@@ -51,8 +51,8 @@ public interface Transaction {
    * Return the transaction completion {@code Future} that
    *
    * <ul>
-   *   <li>succeeds when the transaction commits</li>
-   *   <li>fails with {@link TransactionRollbackException} when the transaction rollbacks</li>
+   *   <li>succeeds when the transaction commits and</li>
+   *   <li>fails with {@link TransactionRollbackException} when the transaction rolls back</li>
    * </ul>
    *
    * @return the transaction result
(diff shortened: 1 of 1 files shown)


vietj:
why is it a typo ?

gavinking:
I'm not sure that's a typo, but "rollback" isn't a verb. It should be "when the transaction rolls back" ;-)

vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
```diff
@@ -51,8 +51,8 @@
    * Return the transaction completion {@code Future} that
    *
    * <ul>
-   *   <li>succeeds when the transaction commits</li>
-   *   <li>fails with {@link TransactionRollbackException} when the transaction rollbacks</li>
+   *   <li>succeeds when the transaction commits and</li>
```
vietj, line 54: why is _and_ added ? it's a bullet list

julianladisch: "and" is missing in the method summary of Transaction.completion on
https://vertx-web-site.github.io/docs/apidocs/io/vertx/sqlclient/Transaction.html because all browsers that I've tested don't show any bullet list: Firefox, Opera, Chrome, Safari
transaction-bullet-point

vietj: ok, then I think best is that you also remove the UL/LI markup in your PR

julianladisch: Removed.

vietj: thanks

PR #810, opened 2020-11-27
DataTypeCodec with Locale cannot parse '4714-11-24 00:00:00 BC'

julianladisch:
If the Java default Locale is set to Locale.GERMAN or Locale.JAPANESE
DataTypeCodec fails with
`DateTimeParseException: Text '4714-11-24 00:00:00 BC' could not be parsed`.

The German and Japanese strings are
`4714-11-24 00:00:00 v. Chr.`
`4714-11-24 00:00:00 紀元前`

Use Locale.ROOT for parsing to ignore any default Locale.


vietj:
can you provide a PR backporting to 3.9 ?

PR #816, opened 2020-11-29
Add json element accessor for row

### LINKED ISSUE #815: Provide accessors of JSON element for Row and Tuple
We can prodive accessors of JSON element for `Row` and `Tuple` APIs so that users could retrieve a JSON value which might be a number, string, object, array or null.

BillyYccc:
fixes #815


vietj:
I'm wondering if we should not just call this `getJson` instead ?

currently JsonArray and JsonObject do have a `getValue` method.

what's your opinion @pmlopes ?

vietj:
I've done a quick research and we use the term _json_ most of the time when we need this meaning:

- WebClient `HttpRequest#sendJson(Object, ...)`
- Schema: `Object getJson()`
- Schema Router: `addJson(String, Object)`
- Buffer : `toJson()`
- JsonPointer : `Object queryJson()`
- data object `toJson()`
- `RoutingContext#json(Object)`

so we should use `getJson`

vertx-sql-client/src/main/java/io/vertx/sqlclient/Row.java
```diff
@@ -188,6 +188,29 @@ default String getString(String column) {
...
 
+  /**
+   * Get a JSON value for the given {@code column}, the JSON value might be one of the following types:
+   * <ul>
+   *   <li>String</li>
+   *   <li>Number</li>
+   *   <li>JsonObject</li>
+   *   <li>JsonArray</li>
+   *   <li>Boolean</li>
+   *   <li>Null</li>
```
vietj, line 199: `Null` is not a type, instead it should be `might be {@code null} or one of the following types:`

vietj (changes requested):
A couple of concerns to address

vietj:
and of course I know it's not a change you did and it was already in the API. So it's a fix before 4.0.0 is released.

vietj:
Actually now I remember why I named this "getJsonElement", the main reason is that the `getJsonArray` is conflicting with the actual `getJsonArray` naming....

vietj:
so we need to settle on a decision to keep `element` or use `value` I think

vietj:
one possibility is to rename array (which are rarely used in practice) to `getArrayOfXXX` which would give us : `getArrayOfJson` / `getArrayOfJsonArray`

BillyYccc:
I think `getArrayOfXXX` seems inconsistent which is not in the same pattern with the `getIntegerArray`, `getStringArray`...
I'm inclined to use the `getJsonElement` compared with `getJsonValue`.

BillyYccc:
We could also override the `toString` method for the `JSON_NULL` object which will return "null" string.

vietj:
@BillyYccc I updated master and 3.9 and renamed to `getArrayOf` so now you can rename to `getJson(K)` and `getArrayOfJson(K)`

vietj:
@BillyYccc can you rebase and rename to `getJson` and `getArrayOfJson` ?

vertx-sql-client/src/main/java/io/vertx/sqlclient/Row.java
```diff
@@ -665,6 +687,29 @@ default BigDecimal getBigDecimal(String column) {
...
+   * @return the {@code column} value
+   * @throws NoSuchElementException when the {@code column} does not exist
+   */
+  @GenIgnore(GenIgnore.PERMITTED_TYPE)
+  default Object[] getArrayOfJson(String column) {
+    int pos = getColumnIndex(column);
+    if (pos == -1) {
+      throw new NoSuchElementException("Column " + column + " does not exist");
+    }
+    return getArrayOfJson(pos);
```
vietj, line 710: should be `getArrayOfJsons`

vietj (approved):

vietj:
thanks @BillyYccc this PR raised a very relevant couple of issues to fix before 4.0.0 or live with it

PR #817, opened 2020-11-29
add json element accessor for row

### LINKED ISSUE #815: Provide accessors of JSON element for Row and Tuple

BillyYccc:
fixes #815


vertx-sql-client/src/main/java/io/vertx/sqlclient/Row.java
```diff
@@ -841,4 +863,27 @@ default BigDecimal getBigDecimal(String column) {
...
+   *   <li>JsonArray</li>
+   *   <li>Boolean</li>
+   * </ul>
+   *
+   * @param column the column name
+   * @return the {@code column} value
+   * @throws NoSuchElementException when the {@code column} does not exist
+   */
+  @GenIgnore(GenIgnore.PERMITTED_TYPE)
+  default Object[] getArrayOfJson(String column) {
```
vietj, line 881: should be `getArrayOfJsons`

vietj (approved):

PR #819, opened 2020-11-30
Use AtomicBoolean for connection pool check

adamdickmeiss:
Motivation:

Getting NPEs or hangs in ConnectionPool.check function. Sometimes, waiters.poll returns null which indicates undesired concurrent access.

Using an atomic boolean for this solves this issue for us.

Conformance:

Already signed this.

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/ConnectionPool.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/ConnectionPool.java
@@ -32,6 +32,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


vietj:
Normally checkInProgress is only used from the same event loop thread

vietj:
Do you have a reproducer for your case or more info like the stack trace ?

adamdickmeiss:
I'll try to nail it down.. What I say is that adding synchronized to relevant sections touching waiters dequeue does seem to solve it for me. AFAIK, we don't use any threading on our own..

adamdickmeiss:
I can confirm - as I expected - that ConnectionPool.check is called with different ThreadIDs.

adamdickmeiss:
Added https://github.com/eclipse-vertx/vertx-sql-client/issues/823 so discussion can continue there. THx

PR #820, opened 2020-11-30
Use AtomicBoolean for connection pool check

adamdickmeiss:
Signed-off-by: Adam Dickmeiss <adam@indexdata.dk>

Motivation:

Getting hangs or NPEs in ConnectionPool.check under heavy usage. NPE is seen because waiters.poll returns null item, despite the fact that it seemingly is non-empty in earlier check. Using AtomicBoolean to check if already in progress solves it for us.


adamdickmeiss:
Further tests shows that this does not fix the issue we're facing, so closing.

vietj:
indeed it should not be related, can you post more details ?

Issue #823, opened 2020-12-01
NPE or hangs in ConnectionPool


adamdickmeiss:
### Version

vertx-sq-client 4.0.0-SNAPSHOT   commit 4a90fc5f1c30041b6eea3304760f0cb50acde9c3

### Context

While perform a large batch import test in https://github.com/folio-org/mod-inventory-storage of FOLIO we get NPEs or hangs.. For the NPE we get traces like these:

```
java.lang.NullPointerException: null
        at io.vertx.sqlclient.impl.pool.ConnectionPool.check(ConnectionPool.java:260) ~[vertx-sql-client-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
        at io.vertx.sqlclient.impl.pool.ConnectionPool.release(ConnectionPool.java:245) ~[vertx-sql-client-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
        at io.vertx.sqlclient.impl.pool.ConnectionPool.access$200(ConnectionPool.java:44) ~[vertx-sql-client-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
        at io.vertx.sqlclient.impl.pool.ConnectionPool$PooledConnection.doClose(ConnectionPool.java:198) ~[vertx-sql-client-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
        at io.vertx.sqlclient.impl.pool.ConnectionPool$PooledConnection.lambda$close$0(ConnectionPool.java:179) ~[vertx-sql-client-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
        at io.vertx.core.impl.AbstractContext.dispatch(AbstractContext.java:96) ~[vertx-core-4.0.0.CR2.jar:4.0.0.CR2]
        at io.vertx.core.impl.WorkerContext.lambda$emit$0(WorkerContext.java:78) ~[vertx-core-4.0.0.CR2.jar:4.0.0.CR2]
        at io.vertx.core.impl.WorkerContext.execute(WorkerContext.java:113) ~[vertx-core-4.0.0.CR2.jar:4.0.0.CR2]
        at io.vertx.core.impl.WorkerContext.emit(WorkerContext.java:77) ~[vertx-core-4.0.0.CR2.jar:4.0.0.CR2]
[...]

PR #824, opened 2020-12-01
Possible fix for #823

adamdickmeiss:
Access to available/waiters/all synchronized.

Signed-off-by: Adam Dickmeiss <adam@indexdata.dk>

Motivation:

Solves #823 for us. There's no tests in this PR.


vietj:
actually it should not be made synchronized, instead we should check we are on the correct thread in the connection pool.

currently the code assumes that the SQL client is created using an event-loop context and your stack trace seems to use it from a worker verticle thread.

vietj:
see https://github.com/eclipse-vertx/vertx-sql-client/issues/825

PR #826, opened 2020-12-01
DataTypeCodec with Locale cannot parse '4714-11-24 00:00:00 BC'

julianladisch:
If the Java default Locale is set to Locale.GERMAN or Locale.JAPANESE
DataTypeCodec fails with
`DateTimeParseException: Text '4714-11-24 00:00:00 BC' could not be parsed`.

The German and Japanese strings are
`4714-11-24 00:00:00 v. Chr.`
`4714-11-24 00:00:00 紀元前`

Use Locale.ROOT for parsing to ignore any default Locale.

Back-ported from master branch to 3.9 branch.
(cherry picked from commit a6dba023704e922b02ce6c7dd8e1bb6c23acb417)


vietj:
thanks @julianladisch

PR #830, opened 2020-12-02
Fix the incorrect execution flow for handling MySQL infile request empty file

### LINKED ISSUE #827: MySQL load empty data infile cause the session being in an illegal state
### Questions

Load empty data infile cause can't create Transaction correctly. new TransactionImpl schedule command "BEGIN" but result in fail with exception.

### Version

3.9.4

### Do you have a reproducer?

https://github.com/longdt/vertx-mysql-txn-bug
Please checkout this repo then run TransactionTest

### Steps to reproduce

1. use pool.query("LOAD DATA LOCAL INFILE ")  with empty file
2. on result handler. Try get Transaction from pool (`pool.begin()`) 
3. use Transaction to query data 
4. on result handler. Exception: io.vertx.core.VertxException: Transaction already completed occur

### Extra

OS: Ubuntu 20.04
java: 11
mysql server: 5.7

Sorry for my poor english

BillyYccc:
Motivation:

Currently when the client executes loading an empty file request, it will incorrectly send duplicated empty packets to the server which leads to corrupting the session.

fixes #827

### DIFF
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/SimpleQueryCommandCodec.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/SimpleQueryCommandCodec.java
@@ -26,7 +26,6 @@ import java.io.File;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


vietj:
shall we merge that @BillyYccc ?

BillyYccc:
yes we're good to go

PR #835, opened 2020-12-03
Rewrite the internal state transition of the Transaction 

### LINKED ISSUE #833: Improve transaction rollback of an already completed transaction
The implementation of the `Transaction` does not handle correctly reentrant termination of the transaction, i.e calling `rollback` or `commit` while the transaction state becomes completed. This can lead to releasing twice the connection to the connection pool preventing the state of the application to progress.

Here is a reproducer:

```java
@Test
  public void testReproducer(TestContext ctx) {
    Async async = ctx.async();
    connector.accept(ctx.asyncAssertSuccess(tx -> {
      tx.query("SELECT whatever from DOES_NOT_EXIST").execute(ctx.asyncAssertFailure());
      tx.query("SELECT 1").execute(ctx.asyncAssertFailure(err -> {
        tx.rollback(ctx.asyncAssertFailure(v -> {
          // Already rolled back
          async.complete();
        }));
      }));
    }));
  }
```

This test adds 2 queries to the same transaction and add the second before the result of the first query (thing that is not possible with synchronous API, unless 2 threads are used).

When the first query is executed, the transaction is rolled backed and with the current implementation it will fail all the pending queries (See TransactionImpl#wrap method) before the transaction is rolled back (same method after).

This test will rollback the transaction in the completion handler of the 2nd query and then the transaction will try to roll back the transaction. This will result in 2 callbacks that will both try to return the connection to the pool and the 2nd one will fail with an IllegalStateException.

vietj:
to better handle reentrancy. A few simplification have been made too

fixes #833

### DIFF
--- a/vertx-pg-client/src/test/java/io/vertx/pgclient/PgConnectionTestBase.java
+++ b/vertx-pg-client/src/test/java/io/vertx/pgclient/PgConnectionTestBase.java
@@ -402,29 +402,23 @@ public abstract class PgConnectionTestBase extends PgClientTestBase<SqlConnectio
   @Test
   public void testTransactionAbort(TestContext ctx) {
     Async done = ctx.async();
+    Async done2 = ctx.async();
     connector.accept(ctx.asyncAssertSuccess(conn -> {
       deleteFromTestTable(ctx, conn, () -> {
         Transaction tx = conn.begin();
-        AtomicInteger failures = new AtomicInteger();
-        tx.abortHandler(v -> ctx.assertEquals(0, failures.getAndIncrement()));
-        AtomicReference<AsyncResult<RowSet<Row>>> queryAfterFailed = new AtomicReference<>();
-        AtomicReference<AsyncResult<Void>> commit = new AtomicReference<>();
+        tx.abortHandler(v -> {
+          done2.complete();
+        });
         conn.query("INSERT INTO Test (id, val) VALUES (1, 'val-1')").execute(ar1 -> { });
-        conn.query("INSERT INTO Test (id, val) VALUES (1, 'val-2')").execute(ar2 -> {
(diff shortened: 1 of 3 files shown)


chris-brace:
Holy crap thank you for this. What a turn around. Just from a quick look at the diff it seems like this may resolve the issue, but I would like to test it locally. Do these builds end up on maven somewhere? It's not exactly clear at a cursory glance how i can build the library locally.

vietj:
@chris-brace they shall be available in Sonatype OSS snapshot repository https://oss.sonatype.org/content/repositories/snapshots/

raviprakashshahi:
Hi @vietj Thanks for the fix, could you please confirm this fix is also present in vertx version 4.2.7

vietj:
Vert.x 4 transaction has a slightly different implementation that does not allow this to happen as it does not seek to pro actively rollback the transaction. When a transaction rollbacks implicitly, the transaction is marked as failed and the transaction state progress when the application ends the transaction. Marking as rollback then implies the transaction is completed with a rollback operation.

raviprakashshahi:
Hi @vietj Thanks for the response. We have just migrated from Vertx 3 to 4 and have been facing connection pool issue in some of the environment with high traffic. I am not sure if the issues are related - In my cause the call to get a connection from pool times out, and once it starts happening, over time all the calls timeout.

We looked at certain metrics and saw metric `vertx_sql_queue_pending` goes high when the issue starts to appear, could you please let me know if there is any resource scarcity or some connection leakage.

Tried setting the connection_idle_timeout to a high value. - 3 hours

Now, we are executing queries on the pool and not getting a new connection, but still we are seeing issues, this time I am not able to get the stacktrace but the system performance degrades over time.

Any pointers would be helpful here. Thanks in advance,

PR #843, opened 2020-12-05
Add support for connection retry

### LINKED ISSUE #805: Connect retry
The current options provides a retry that is used by the TCP layer but it would be good to have it working at the protocol level,  when a database is started and not yet ready it can error the connection and close it.

BillyYccc:
Motivation:

Fixes #805 

Conformance:


vietj:
very good, we ship it with 4.0

PR #848, opened 2020-12-09
The close handler should be called when the sql connection is closed from a different context

### LINKED ISSUE #847: [vertx 4.0.0] Connection.close callback may not be called
### Version

Vertx 4.0.0

### Context

Some code working in vertx 3.9.x begin to hang.
After investigation it was comming from `connection.close().await()` (using kotlin-coroutine)
Replacing the kotlin extension `await()` by `onSuccess { ... }.onFailure{ ... }.onComplete{ ... }` showed that none is called.

### Do you have a reproducer?

No, but I think I have an explaination.

The method `close()` is called on a `ConnectionConnectionImpl`. This method create a new `Promise` then call `close(Promise ...)`
If the context is not vertx current context, it dispatch a new `close()` call in the context. But this call has no link to the promise created by the initial `close()` call.

```
@Override
  public Future<Void> close() {
    Promise<Void> promise = promise(); // new promise here
    close(promise);
    return promise.future();
  }

private void close(Promise<Void> promise) {
    if (context == Vertx.currentContext()) {
      if (tx != null) {
        tx.rollback(ar -> conn.close(this, promise));
        tx = null;
      } else {
        conn.close(this, promise);
      }
    } else {
      context.runOnContext(v -> close()); // so `promise` argument of method is never used, a close() will generate a new one
    }
  }
```

BillyYccc:
Signed-off-by: Billy Yuan <billy112487983@gmail.com>

Motivation:

fix #847 

Conformance:

### DIFF
--- a/vertx-pg-client/src/test/java/io/vertx/pgclient/PgConnectionTestBase.java
+++ b/vertx-pg-client/src/test/java/io/vertx/pgclient/PgConnectionTestBase.java
@@ -17,7 +17,6 @@
 
 package io.vertx.pgclient;
 
-import io.vertx.core.impl.VertxInternal;
 import io.vertx.sqlclient.Row;
 import io.vertx.sqlclient.RowSet;
 import io.vertx.sqlclient.SqlConnection;
@@ -471,14 +470,12 @@ public abstract class PgConnectionTestBase extends PgClientTestBase<SqlConnectio
       conn.query("SELECT 1").execute(ctx.asyncAssertSuccess(res -> {
         ctx.assertEquals(1, res.size());
         // schedule from another context
-        VertxInternal vertxInternal = (VertxInternal) vertx;
-        vertxInternal.createEventLoopContext().runOnContext(v -> {
+        new Thread(() -> {
           conn.close(v2 -> {
             done.complete();
           });
-        });
+        }).start();
       }));
     }));
-    done.await(2000);
   }
 }
--- a/vertx-pg-client/src/test/java/io/vertx/pgclient/PgConnectionTestBase.java
+++ b/vertx-pg-client/src/test/java/io/vertx/pgclient/PgConnectionTestBase.java
@@ -17,6 +17,7 @@
 
 package io.vertx.pgclient;
 
+import io.vertx.core.impl.VertxInternal;
 import io.vertx.sqlclient.Row;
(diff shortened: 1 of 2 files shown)


vertx-pg-client/src/test/java/io/vertx/pgclient/PgConnectionTestBase.java
```diff
@@ -462,4 +463,22 @@ public void testTransactionAbort(TestContext ctx) {
...
+        // schedule from another context
+        VertxInternal vertxInternal = (VertxInternal) vertx;
+        vertxInternal.createEventLoopContext().runOnContext(v -> {
+          conn.close(v2 -> {
+            done.complete();
+          });
+        });
+      }));
+    }));
+    done.await(2000);
```
vietj, line 482: await is not needed, vertx-unit will wait for you

you need to use await when there is an action to take after

vertx-pg-client/src/test/java/io/vertx/pgclient/PgConnectionTestBase.java
```diff
@@ -462,4 +463,22 @@ public void testTransactionAbort(TestContext ctx) {
...
     }));
   }
+
+  @Test
+  public void testCloseConnectionFromDifferentContext(TestContext ctx) {
+    Async done = ctx.async(1);
+    connector.accept(ctx.asyncAssertSuccess(conn -> {
+      conn.query("SELECT 1").execute(ctx.asyncAssertSuccess(res -> {
+        ctx.assertEquals(1, res.size());
+        // schedule from another context
```
vietj, line 473: rather use a Java thread it will create another context/event loop

i.e

new Thread(() -> ...) {
}.start();

BillyYccc: fixed in https://github.com/eclipse-vertx/vertx-sql-client/pull/848/commits/ce73aae89f5b511e222bceb2c6d7650fa164d448.

vertx-pg-client/src/test/java/io/vertx/pgclient/PgConnectionTestBase.java
BillyYccc: ah I thought it was going to make the test fail-faster rather than being stuck if the callback is never called.

PR #852, opened 2020-12-15
Use UTC timezone when testing LocalDate

julianladisch:
If Java time zone is +1200 and connection time zone is +0000
then Java will be one day ahead the database connection date
when the test runs before noon.


vietj:
funny

vietj:
can you backport it to 3.9 branch ?

PR #865, opened 2020-12-22
Add tests for checking pooled MySQL connections closed by server due …

BillyYccc:
…to wait_timeout

Signed-off-by: Billy Yuan <billy112487983@gmail.com>

Motivation:

Conformance:

### DIFF
--- a/vertx-sql-client/src/main/asciidoc/connections.adoc
+++ b/vertx-sql-client/src/main/asciidoc/connections.adoc
@@ -3,7 +3,7 @@
 === Getting a connection
 
 When you need to execute sequential queries (without a transaction), you can create a new connection
-or borrow one from the pool:
+or borrow one from the pool, remember once you acquired the connection from the pool until you return it back to the pool, you should take care of the connection because it might be closed by the server for some reasons such as an idle time out.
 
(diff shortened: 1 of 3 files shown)


vertx-sql-client/src/main/asciidoc/connections.adoc
```diff
@@ -3,7 +3,7 @@
 === Getting a connection
 
 When you need to execute sequential queries (without a transaction), you can create a new connection
-or borrow one from the pool:
+or borrow one from the pool, remember once you acquired the connection from the pool until you return it back to the pool, you should take care of the connection because it might be closed by the server for some reasons such as an idle time out.
```
gavinking, line 7: "reason", not "reasons"

vertx-sql-client/src/main/asciidoc/connections.adoc
```diff
@@ -3,7 +3,7 @@
...
```
gavinking, line 6: Should be a new sentence, not a comma.

"Remember that between acquiring the connection from the pool and returning it to the pool, you should...."

gavinking (changes requested):

PR #874, opened 2021-01-18
New connection pool implementation

### LINKED ISSUE #933: New connection pool
A new implementation of the connection pool based on Vert.x Core reusable pool, this pool provides a more robust implementation and does not depend on a Vert.x context. The new pool adds connection timeout supports and idle connection timeout.

In addition a new pooled client is introduced for PostgreSQL and DB2 that allows to obtain a pooled client that perform client operations on the pool that are pipelined on the client. This pooled client does not allow to obtain a connection from the pool on purpose to avoid mixing pipelined pool operations and connection borrowing.

vietj:
Based on https://github.com/eclipse-vertx/vert.x/pull/3731
Fixes #933

### DIFF
--- a/vertx-pg-client/src/main/java/io/vertx/pgclient/PgPool.java
+++ b/vertx-pg-client/src/main/java/io/vertx/pgclient/PgPool.java
@@ -23,7 +23,7 @@ import io.vertx.sqlclient.PoolOptions;
 import io.vertx.sqlclient.Pool;
 import io.vertx.codegen.annotations.VertxGen;
 import io.vertx.core.Vertx;
-import io.vertx.core.VertxOptions;
+import io.vertx.sqlclient.SqlClient;
 
 /**
  * A {@link Pool pool} of {@link PgConnection PostgreSQL connections}.
@@ -89,22 +89,79 @@ public interface PgPool extends Pool {
    * @return the connection pool
    */
   static PgPool pool(PgConnectOptions connectOptions, PoolOptions poolOptions) {
-    if (Vertx.currentContext() != null) {
-      throw new IllegalStateException("Running in a Vertx context => use PgPool#pool(Vertx, PgConnectOptions, PoolOptions) instead");
-    }
-    VertxOptions vertxOptions = new VertxOptions();
-    if (connectOptions.isUsingDomainSocket()) {
-      vertxOptions.setPreferNativeTransport(true);
-    }
-    VertxInternal vertx = (VertxInternal) Vertx.vertx(vertxOptions);
-    return PgPoolImpl.create(vertx.getOrCreateContext(), true, connectOptions, poolOptions);
+    return PgPoolImpl.create(false, connectOptions, poolOptions);
   }
 
   /**
    * Like {@link #pool(PgConnectOptions, PoolOptions)} with a specific {@link Vertx} instance.
    */
   static PgPool pool(Vertx vertx, PgConnectOptions connectOptions, PoolOptions poolOptions) {
-    return PgPoolImpl.create(((VertxInternal)vertx).getOrCreateContext(), false, connectOptions, poolOptions);
+    return PgPoolImpl.create(((VertxInternal)vertx), false, false, connectOptions, poolOptions);
   }
 
+  /**
+   * Like {@link #client(PoolOptions)} with a default {@code poolOptions}.
+   */
+  static SqlClient client() {
+    return client(PgConnectOptions.fromEnv(), new PoolOptions());
+  }
+
+  /**
+   * Like {@link #client(PgConnectOptions, PoolOptions)} with {@code connectOptions} build from the environment variables.
+   */
+  static SqlClient client(PoolOptions poolOptions) {
+    return client(PgConnectOptions.fromEnv(), poolOptions);
+  }
+
+  /**
+   * Like {@link #pool(String, PoolOptions)} with a default {@code poolOptions}.
+   */
+  static SqlClient client(String connectionUri) {
+    return client(connectionUri, new PoolOptions());
+  }
+
+  /**
+   * Like {@link #client(PgConnectOptions, PoolOptions)} with {@code connectOptions} build from {@code connectionUri}.
+   */
+  static SqlClient client(String connectionUri, PoolOptions poolOptions) {
+    return client(PgConnectOptions.fromUri(connectionUri), poolOptions);
+  }
+
+  /**
+   * Like {@link #client(Vertx, String,PoolOptions)} with a default {@code poolOptions}.
+   */
+  static SqlClient client(Vertx vertx, String connectionUri) {
+    return client(vertx, PgConnectOptions.fromUri(connectionUri), new PoolOptions());
+  }
+
+  /**
+   * Like {@link #client(Vertx, PgConnectOptions, PoolOptions)} with {@code connectOptions} build from the environment variables.
+   */
+  static SqlClient client(Vertx vertx, PoolOptions poolOptions) {
+    return client(vertx, PgConnectOptions.fromEnv(), poolOptions);
+  }
+
+  /**
+   * Like {@link #client(Vertx, PgConnectOptions, PoolOptions)} with {@code connectOptions} build from {@code connectionUri}.
+   */
+  static SqlClient client(Vertx vertx, String connectionUri, PoolOptions poolOptions) {
+    return client(vertx, PgConnectOptions.fromUri(connectionUri), poolOptions);
+  }
+
+  /**
+   * Create a client backed by a connection pool to the database configured with the given {@code connectOptions} and {@code poolOptions}.
+   *
+   * @param poolOptions the options for creating the backing pool
+   * @return the client
+   */
+  static SqlClient client(PgConnectOptions connectOptions, PoolOptions poolOptions) {
+    return PgPoolImpl.create(true, connectOptions, poolOptions);
+  }
+
+  /**
+   * Like {@link #client(PgConnectOptions, PoolOptions)} with a specific {@link Vertx} instance.
+   */
+  static SqlClient client(Vertx vertx, PgConnectOptions connectOptions, PoolOptions poolOptions) {
+    return PgPoolImpl.create(((VertxInternal)vertx), false, true, connectOptions, poolOptions);
+  }
 }
--- a/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgPoolImpl.java
+++ b/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgPoolImpl.java
@@ -19,7 +19,6 @@ package io.vertx.pgclient.impl;
 
 import io.vertx.core.impl.CloseFuture;
 import io.vertx.core.impl.ContextInternal;
-import io.vertx.core.impl.EventLoopContext;
 import io.vertx.core.spi.metrics.ClientMetrics;
 import io.vertx.core.spi.metrics.VertxMetrics;
 import io.vertx.pgclient.*;
@@ -45,8 +44,7 @@ public class PgPoolImpl extends PoolBase<PgPoolImpl> implements PgPool {
     QueryTracer tracer = context.tracer() == null ? null : new QueryTracer(context.tracer(), connectOptions);
     VertxMetrics vertxMetrics = context.owner().metricsSPI();
     ClientMetrics metrics = vertxMetrics != null ? vertxMetrics.createClientMetrics(connectOptions.getSocketAddress(), "sql", connectOptions.getMetricsName()) : null;
-    EventLoopContext eventLoopContext = ConnectionFactory.asEventLoopContext(context);
-    PgPoolImpl pool = new PgPoolImpl(eventLoopContext, new PgConnectionFactory(eventLoopContext, connectOptions), tracer, metrics, poolOptions);
+    PgPoolImpl pool = new PgPoolImpl(context, new PgConnectionFactory(context.owner(), connectOptions), tracer, metrics, poolOptions);
     CloseFuture closeFuture = pool.closeFuture();
     if (closeVertx) {
       closeFuture.future().onComplete(ar -> context.owner().close());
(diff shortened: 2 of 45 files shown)


vietj:
@BillyYccc please have a look, this is the proof of concept for the new pool

BillyYccc:
I'll check it later

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/SqlConnectionPool.java
```diff
@@ -0,0 +1,245 @@
...
+  public int available() {
+    return maxSize - pool.size();
+  }
+
+  public int size() {
+    return pool.size();
+  }
+
+  public void acquire(ContextInternal context, Handler<AsyncResult<Connection>> waiter) {
+    pool.acquire((EventLoopContext) context, 1, ar -> {
```
BillyYccc, line 104: This cast will throw a `ClassCastException` under `DuplicatedContext` for example it's called in a HTTP Server request handler.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/PoolBase.java
```diff
@@ -97,21 +101,23 @@ public void getConnection(Handler<AsyncResult<SqlConnection>> handler) {
...
         if (metrics != null) {
           metrics.dequeueRequest(metric);
         }
-        conn.schedule(cmd, promise);
-        promise.future().onComplete(ar -> {
+        conn.schedule(context, cmd)
+          .onComplete(promise)
+          .onComplete(v -> {
           // Use null promise instead
           conn.close(this, Promise.promise());
```
BillyYccc, line 122: I think we need to rework this because it won't work in pipelining mode. I find we have performance regression since https://github.com/eclipse-vertx/vertx-sql-client/pull/744, since all the pending commands will stuck instead of pipelining until the connection is returned back to the pool after the command completes. It can be reproduced by using sleep function and pipelining.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/SqlConnectionPool.java
```diff
@@ -0,0 +1,245 @@
...
+      .collect(Collectors.toList());
+    CompositeFuture
+      .join(futures)
+      .<Void>mapEmpty()
+      .onComplete(promise);
+  }
+
+  private class PooledConnection implements Connection, Connection.Holder  {
+
+    private final Connection conn;
```
BillyYccc, line 145: I'm wondering whether we can remove `Connection` interface and just wrap `SqlConnection` in `PooledConnection` instead as we have to wrap `Connection` in `SqlConnection` each time for connection acquirement.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlClientBase.java
```diff
@@ -53,6 +54,8 @@ public int appendQueryPlaceholder(StringBuilder queryBuilder, int index, int cur
     return current;
   }
 
+  protected abstract ContextInternal context();
```
BillyYccc, line 57: this seems unnecessary here because we use contexual promise interface instead?

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/SqlConnectionPool.java
vietj: alright it should need an integration test to ensure it's fine (and be corrected)

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/PoolBase.java
vietj: oh really ? can we have a test that check pipelining ?

vietj: I was actually thinking at something else for pipelining, to use the connection capacity to allow a connection to be borrowed multiple times fro mthe pool

vietj: I'm first thinking how can we write a test that ensure pipelining works, any idea ?

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/SqlConnectionPool.java
vietj: we can check that later, it's a preliminary integration test :-)

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlClientBase.java
vietj: I need to check

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/PoolBase.java
BillyYccc: we can write a mini proxy server which can check if multiple query requests are sent to the server in the same round trip

vietj: yes,but that requires to parse on the proxy the protocol

vietj: well one thing we could do, is send 2 request with specific identifiers and look for those IDs and check that we have the 2 ids are written before the 2 ids are written back back on the response.

vietj: @BillyYccc here is a tentative test that uses a proxy based that delays the packets sent to the database: https://github.com/eclipse-vertx/vertx-sql-client/pull/878

vietj: I think one way we can make everyone happy is to use the connection capacity.

On the pool we define a pipelining limit (like for HTTP/1.x connection pool) and we set the initial capacity `C` of the connection to the limit, so the connection can be borrowed at most `C` times from the pool.

However this needs to be accommodated if a client wants an exclusive connection (i.e borrow with full capacity).

BillyYccc: > @BillyYccc here is a tentative test that uses a proxy that delays packets sent to the database: #878
> 
> This test fails on a pool but works on an un-pooled connection (not in the PR but checked by myself).

yes it works for unpooled connection since all scheduling commands are directly enqueued in the pending queue of the connection.

BillyYccc: > I think one way we can make everyone happy is to use the connection capacity.
> 
> On the pool we define a pipelining limit (like for HTTP/1.x connection pool) and we set the initial capacity `C` of the connection to the limit, so the connection can be borrowed at most `C` times from the pool.
> 
> However this needs to be accommodated if a client wants an exclusive connection (i.e borrow with full capacity).

yes we need to acquire exclusively for the stateful connection scene such as a transaction execution, we have to make sure the transactional commands are scheduled in the same connection.

vietj: I have mixed feelings about implementing this as it does not seem to work well together, i.e having a pool policy that is both performant and fair does not seem possible, either the pool is fair and a request to an exclusive connection might block pipelining use or the pool might be unfair.

So I think the users should  have a choice to do between a pipelining pool and a pool that grants exclusive access to a connection.

BillyYccc: > So I would be inclined to have a pipelining limit=1 by default on `PoolOptions` and throw an exception if `getConnection` is called when limit>1.

Abandoing the usage of pipelining with `getConnection` is not good, because sometimes you might want to chain some operations in a reusable pooled connection without transactions but with pipelining enabled.
I wonder if we can abstract the fact and maintain two pools in the `Pool` interface, one is dedicated for `getConnection` use case and another one for one-shot executions. If the user calls `getConnection` then he gets a pooled connection and can do whatever he wants with the connection, commands like `transactions` will also be scheduled on the pool.

BillyYccc: > So I would be inclined to have a pipelining limit=1 by default on `PoolOptions` and throw an exception if `getConnection` is called when limit>1.

Abandoing the usage of pipelining with `getConnection` is not good, because sometimes you might want to chain some operations in a reusable pooled connection without transactions but with pipelining enabled, like cursor and batching.
I wonder if we can abstract the fact and maintain two pools in the `Pool` interface, one is dedicated for `getConnection` use case and another one for one-shot executions. If the user calls `getConnection` then he gets a pooled connection and can do whatever he wants with the connection, commands like `transactions` will also be scheduled on the pool.

BillyYccc: This would also introduce no breaking API changes for users but we might need some additional configuration params...

vietj: So, we can support a non pipelined pool that uses pipelined connections, that is not a problem, i.e one can create a pool with pipelining limit = 1 on the pool but still use pipelining when borrowing a connection from the pool

vietj: What we can have is a method that creates an `SqlClient` (so no getConnection method) that uses a pool of connections  that are pipelined, e.g

```java
SqlClient client = PgPool.pipelinedClient(poolOptions);
````

Comment (vietj) [reply vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/PoolBase.java:122]: or we can simply assume that `pool` methods will either throw an exception if pipeliningLimit > 1 or ignore this setting.

Comment (vietj) [reply vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/PoolBase.java]: @BillyYccc I implemented pool pipelining, in the way I removed the need to create a pooled connection when executing a simple command. Currently `PgPool` got a `client` method that will create a pipelined pool with a pipelining limit set to the connection pipelining limit.

File: vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/SqlConnectionPool.java (hunk 3 above)
Comment (vietj) [reply vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/SqlConnectionPool.java:160]: This has been removed for executing simple commands on the pool and is only used when acquiring a acquiring a connection out of the pool. You can see this as a protection layer between the physical connection and the API.

File: vertx-pg-client/src/main/java/io/vertx/pgclient/PgPool.java [hunk 5]
```diff
@@ -97,14 +98,31 @@ static PgPool pool(PgConnectOptions connectOptions, PoolOptions poolOptions) {
...
+    return PgPoolImpl.create(((VertxInternal)vertx).getOrCreateContext(), false, false, connectOptions, poolOptions);
   }
 
+  /**
+   * Create a client backed by a connection pool to the database configured with the given {@code connectOptions} and {@code poolOptions}.
+   *
+   * @param poolOptions the options for creating the backing pool
+   * @return the client
+   */
+  static SqlClient client(PgConnectOptions connectOptions, PoolOptions poolOptions) {
```
BillyYccc, line 117: we need an overload factory method with vertx instance parameter.

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgPoolImpl.java
```diff
@@ -40,11 +40,12 @@
  */
 public class PgPoolImpl extends PoolBase<PgPoolImpl> implements PgPool {
 
-  public static PgPoolImpl create(ContextInternal context, boolean closeVertx, PgConnectOptions connectOptions, PoolOptions poolOptions) {
+  public static PgPoolImpl create(ContextInternal context, boolean closeVertx, boolean pipelined, PgConnectOptions connectOptions, PoolOptions poolOptions) {
     QueryTracer tracer = context.tracer() == null ? null : new QueryTracer(context.tracer(), connectOptions);
     VertxMetrics vertxMetrics = context.owner().metricsSPI();
     ClientMetrics metrics = vertxMetrics != null ? vertxMetrics.createClientMetrics(connectOptions.getSocketAddress(), "sql", connectOptions.getMetricsName()) : null;
-    PgPoolImpl pool = new PgPoolImpl(context, new PgConnectionFactory(ConnectionFactory.asEventLoopContext(context), connectOptions), tracer, metrics, poolOptions);
+    int pipeliningLimit = pipelined ? connectOptions.getPipeliningLimit() : 1;
```
BillyYccc, line 47: If I am understanding it right, for a pool created using `Pool.pool` will behave like this after changes:

For Pool#getConnection operations, the query commands will be executed with pipeliningLimit=1(pipeliningLimit breaking change)
For Pool#query operations, the query commands will be scheduled in different pooled connections but be ing executed with pipeliningLimit=1(this is actually not a breaking change because currently our pool works in a way like this)
Please correct me if I'm wrong.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/SqlConnectionPool.java
```diff
@@ -0,0 +1,260 @@
...
+  public <R> Future<R> execute(ContextInternal context, CommandBase<R> cmd) {
+    Promise<R> promise = context.promise();
+    pool.acquire((EventLoopContext) context, 1, ar -> {
+      if (ar.succeeded()) {
+        Lease<PooledConnection> lease = ar.result();
+        PooledConnection pooled = lease.get();
+        pooled.schedule(context, cmd)
+          .onComplete(promise)
+          .onComplete(v -> {
+            lease.recycle();
```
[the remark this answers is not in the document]
vietj: I don't think so

vertx-pg-client/src/main/java/io/vertx/pgclient/PgPool.java
vietj: yes indeed

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgPoolImpl.java
vietj: yes, the goal is that connections are used exclusively and pipelining will only occur with the `client()` method

PR #875, opened 2021-01-19
Add support for MySQL pipelining execution

### LINKED ISSUE #861: MySQL command pipelining feature support
#### Describe the feature

Even though there is no relative official documentation stating that MySQL supports command pipelining mode which means the client sends multiple requests sequencely without waiting for the repsonse, it seems that it's possible for a client to achieve this.

We need to investigate how the pipelining works in different scenes:
* simple queries
* prepared queries(cache and one-shot)
* different command flow execution(mixing simple query + prepared query)
* utility command
* transaction

etc...

#### Use cases

I believe using pipeling mode in simple queries and prepared queries would improve performance much like what reactive postgres client ever achieved.

#### Contribution

I will investigate how this would work.

### LINKED ISSUE #873: MySQL managed one-shot prepared queries statement auto closing command should be sent before any pending commands in the queue
Motivation:
Currently MySQL auto closing command is fired after the execution is completed, this works fine because in MySQL Client the pipeling limit is only 1 and it will be sent before other pending commands, however it might cause unexpected order issues in pipeling mode which we will support later.

BillyYccc:
Motivation:

fixes #861, fixes #873

### DIFF
--- a/vertx-mysql-client/src/main/asciidoc/index.adoc
+++ b/vertx-mysql-client/src/main/asciidoc/index.adoc
@@ -661,6 +661,7 @@ You can also manage the lifecycle of prepared statements manually by creating a 
 
 There is time when you want to batch insert data into the database, you can use `PreparedQuery#executeBatch` which provides a simple API to handle this.
 Keep in mind that MySQL does not natively support batching protocol so the API is only a sugar by executing the prepared statement one after another, which means more network round trips are required comparing to inserting multiple rows by executing one prepared statement with a list of values.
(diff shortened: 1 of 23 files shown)


BillyYccc:
I will create another two PRs for optional metadata and encoder rework as described in the https://github.com/eclipse-vertx/vertx-sql-client/issues/861#issuecomment-762375574.

vietj:
should this be reviewed ?

BillyYccc:
I will rebase on the master to adapt to the new pool implementation.

vietj:
@BillyYccc can we review this now to be in 4.1.1 ?

BillyYccc:
Closed with a substitution PR.

PR #878, opened 2021-01-25
Added a test to ensure pipelining on a pool

vietj:
There is a performance regression on pool pipelining that prevents queries to be pipelined on the pooled connections. This PR provides only a test example to check that pipelining works.

The test performs a series of N queries and proxies the database to add latency of 1 second. If pipelining works we should have the N queries executed under 2 seconds as each query should be sent independently of the result.


BillyYccc:
this is quite a wise attempt so that we don't need to parse the protocol :smile:

vietj:
thanks!

PR #880, opened 2021-01-31
Fixing incorrect link syntax

gunnarmorling:
This was using MD syntax rather than AsciiDoc.

I'm also having some doubts about the "Getting Started" example. When running this in a simple main method, the program is finished before the (asynchronous) result handler is invoked. Shouldn't there be a countdown latch or other means in place to make sure the result is received before the program terminates?

### DIFF
--- a/vertx-pg-client/src/main/asciidoc/index.adoc
+++ b/vertx-pg-client/src/main/asciidoc/index.adoc
@@ -100,7 +100,7 @@ Make sure you have added the required `netty-transport-native` dependency in you
 {@link examples.PgClientExamples#connecting06}
 ----
 
-More information can be found in the [Vert.x documentation](https://vertx.io/docs/vertx-core/java/#_native_transports).
+More information can be found in the https://vertx.io/docs/vertx-core/java/#_native_transports[Vert.x documentation].
 
 == Connect retries
 


BillyYccc:
Can you do a favor and fix the error Asciidoc syntax in https://github.com/eclipse-vertx/vertx-sql-client/blob/master/vertx-mysql-client/src/main/asciidoc/index.adoc#unix-domain-socket as well?

You need also sign the commit(see https://github.com/eclipse-vertx/vertx-sql-client/blob/master/CONTRIBUTING.md#contributions).

BillyYccc:
> I'm also having some doubts about the "Getting Started" example. When running this in a simple main method, the program is finished before the (asynchronous) result handler is invoked. Shouldn't there be a countdown latch or other means in place to make sure the result is received before the program terminates?

once the vertx eventloop thread is created, it will prevent main thread from exiting until the related vertx instance is closed.(i.e. eventloop thread is shutdown)

tsegismont (approved):
Thank you @gunnarmorling !

tsegismont:
@BillyYccc it seems this commit comes from GH online editor, I am not sure it is possible to sign-off commits with this.

Anyway, when the size of the contribution is very small and not fundamental to the project (fixing a documentation link is a good example) then there's a tolerance for accepting it.

gunnarmorling:
Thanks for merging! Sorry about the sign-off, always forget that it's needed for Eclipse projects.

> once the vertx eventloop thread is created, it will prevent main thread from exiting until the related vertx instance is closed.(i.e. eventloop thread is shutdown)

Hum, it's not what I observed, the program finished without any noticeable output when just copy/pasting the example as is. Weird.

tsegismont:
Le lun. 1 févr. 2021 à 11:06, Gunnar Morling <notifications@github.com> a
écrit :

> Hum, it's not what I observed, the program finished without any noticeable
> output when just copy/pasting the example as is. Weird.
>

Billy is right, the program should not stop. I just tried the example and
it worked for me :/

BillyYccc:
In the example the pool is created without assigning a vertx instance so it
will create one and the pool is closed after execution which will trigger
the closehook to shutdown the associated vertx instance so the program will
exit. But it's strange no output is showing in the console.

Gunnar Morling <notifications@github.com> 于 2021年2月1日周一 18:06写道：

Issue #41, opened 2017-12-10
Pool improvements


vietj:
- connection timeout
- idle timeout

pratikpparikh:
@vietj can this be made part of 0.9.1 as this is very important to make sure that connection timeout is handled properly.

Globegitter:
@vietj what is the status on this? I can see one can already set a `idleTimeout` through `PgConnectOptions`, is that then also respected by the pool? But as far as I can see a connection timeout itself is still missing. Can you also confirm that?

And I agree with @pratikpparikh having a connection timeout would be important to have.

vietj:
we will have this when we reuse the current pool implement from vertx

vietj:
I'm scheduling this for 4.0

Globegitter:
@vietj that is great new, but if now I set:

```
PgPool.pool(
        pgConnectOptionsOf(
            host = "localhost",
            port = 5432,
            database = "postgres",
            user = "postgres",
            password = "postgres",
            connectTimeout = 1000,
            idleTimeout = 1000,
            idleTimeoutUnit = TimeUnit.MILLISECONDS
        ),
        poolOptionsOf()
    )
```

then is `connectTimeout` and `idleTimeout` ignored by the pool?

vietj:
those are connection settings, not pool settings. Those should have effect though, i.e the connection should be closed after 1s of TCP inactivity whehter it's int he pool or not.

vietj:
and for connect timeout there are chances it works correctly although one should check

Globegitter:
Cool thanks for the quick responses, will give it a try and report back here.

PR #881, opened 2021-02-01
Vertx PgPool lost connection unexpectedly

pjgg:
# Scenario:

When a user gets a connection from pool some milliseconds behind IDLE expiration time, then "sometimes" when you are going to use the connection, you get the following error from the server:

`Fail to read any response from the server, the underlying connection might get lost unexpectedly.`

So, looks that the connection it's available when you request it, but then disappears when you use it.

[Original Ref:](https://github.com/quarkusio/quarkus/issues/14608)

### API decisions: 

Currently, Idle timeOut is a `PgConnectOptions` parameter. We believe that pool Idle and Tcp/connection Idle is going to be the same in most of the cases, so in this proposal, we keep hidden pool Idle, that it's going to take his value from `PgConnectOptions` Idle parameter. 

Example:

```
     int poolSize = 5;
     int idle = 1000;
    connectionOptions.setIdleTimeout(idle);
    connectionOptions.setIdleTimeoutUnit(TimeUnit.MILLISECONDS);

    PgPool pool = createPool(new PgConnectOptions(connectionOptions), poolSize);
```

In the above example, connection and pool idle has the same value, 1000 milliseconds

### Implementation design

`ConnectionPool` will handle available connections through a new data structured called `ExpiryQueue`. This data structure is a FIFO queue that only returns not expired connections. If all connections are expired then will remove all expired connections (all) and then return `null`, which will be handler by `ConnectionPool`.

Linked Ticket: #41

### DIFF
--- a/vertx-db2-client/src/main/java/io/vertx/db2client/impl/DB2PoolImpl.java
+++ b/vertx-db2-client/src/main/java/io/vertx/db2client/impl/DB2PoolImpl.java
@@ -19,6 +19,7 @@ import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
 import io.vertx.core.impl.CloseFuture;
 import io.vertx.core.impl.ContextInternal;
+import io.vertx.core.impl.EventLoopContext;
 import io.vertx.core.spi.metrics.ClientMetrics;
 import io.vertx.core.spi.metrics.VertxMetrics;
 import io.vertx.db2client.DB2ConnectOptions;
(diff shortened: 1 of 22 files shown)


tsegismont:
Thank you @pjgg 

PR #874 introduces a new pool implementation that will be integrated in Vert.x 4.1. This new pool implementation is able to give a maximum time to live for pooled objects (e.g we use it in HTTP client w/ keepalive time).

It is better to build the idle timeout feature on top of this new pool implementation for Vert.x 4.

That being said, a contribution for the 3.9 branch would be great. Would you like to do it? For this eventuality, here are few goals and comments to this PR:

- pool idle timeout should be configurable in `PoolOptions`
  - by default, connections should not expire
  - it should be independent of `TCPSSLOptions` idle timeout
- we can avoid creating an `ExpiryQueue` with two maps
  - a timer can be created on the pool context when the pool connection is returned to the pool (timers are relatively cheap to create/cancel)
  - when the timer is triggered, the connection is removed from the pool
  - the timer is canceled if the connection is borrowed

pjgg:
Great!. So I will move on this approach(adding all the feedback that you give me) to branch 3.9
So this work was moved to [this issue](https://github.com/eclipse-vertx/vertx-sql-client/pull/883)

tsegismont:
Closing, will implemented pool timeout based on #41

PR #883, opened 2021-02-03
Vertx pool Idle timeout

pjgg:
Original Ref: #881 

## API decisions:
* pool idle configuration it's handler by 'PoolOptions'.
* By default idle is not enable (connections doesn't expire)

## Implementation design
* Connections expiration time is managed by  `ConnectionPool -> PooledConnection` inner class through Vertx timers.

### DIFF
--- a/vertx-pg-client/src/test/java/io/vertx/pgclient/PgPoolTest.java
+++ b/vertx-pg-client/src/test/java/io/vertx/pgclient/PgPoolTest.java
@@ -20,9 +20,8 @@ package io.vertx.pgclient;
 import io.vertx.ext.unit.Async;
 import io.vertx.ext.unit.TestContext;
 import io.vertx.sqlclient.PoolOptions;
-import org.junit.Test;
-
 import java.util.concurrent.atomic.AtomicReference;
+import org.junit.Test;
 
 /**
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
@@ -30,8 +29,8 @@ import java.util.concurrent.atomic.AtomicReference;
 public class PgPoolTest extends PgPoolTestBase {
 
   @Override
-  protected PgPool createPool(PgConnectOptions options, int size) {
-    return PgPool.pool(vertx, options, new PoolOptions().setMaxSize(size));
+  protected PgPool createPool(PgConnectOptions options, PoolOptions poolOpts) {
+    return PgPool.pool(vertx, options, poolOpts);
   }
 
   @Test
@@ -44,7 +43,7 @@ public class PgPoolTest extends PgPoolTestBase {
       conn.connect();
     });
     proxy.listen(8080, "localhost", ctx.asyncAssertSuccess(v1 -> {
-      PgPool pool = createPool(new PgConnectOptions(options).setPort(8080).setHost("localhost"), 1);
+      PgPool pool = createPool(new PgConnectOptions(options).setPort(8080).setHost("localhost"), poolOptions.setMaxSize(1));
       pool.getConnection(ctx.asyncAssertSuccess(conn -> {
         proxyConn.get().close();
       }));
@@ -59,7 +58,7 @@ public class PgPoolTest extends PgPoolTestBase {
   @Test
   public void testAuthFailure(TestContext ctx) {
     Async async = ctx.async();
-    PgPool pool = createPool(new PgConnectOptions(options).setPassword("wrong"), 1);
+    PgPool pool = createPool(new PgConnectOptions(options).setPassword("wrong"), poolOptions.setMaxSize(1));
     pool.query("SELECT id, randomnumber from WORLD").execute(ctx.asyncAssertFailure(v2 -> {
       async.complete();
     }));
@@ -163,3 +162,4 @@ public class PgPoolTest extends PgPoolTestBase {
     }
   }
 }
+
--- a/vertx-pg-client/src/test/java/io/vertx/pgclient/PgPoolTestBase.java
+++ b/vertx-pg-client/src/test/java/io/vertx/pgclient/PgPoolTestBase.java
@@ -17,16 +17,22 @@
 
 package io.vertx.pgclient;
 
-import io.vertx.sqlclient.SqlResult;
-import io.vertx.sqlclient.Tuple;
 import io.vertx.core.Vertx;
 import io.vertx.ext.unit.Async;
 import io.vertx.ext.unit.TestContext;
+import io.vertx.ext.unit.junit.Repeat;
+import io.vertx.ext.unit.junit.RepeatRule;
+import io.vertx.sqlclient.PoolOptions;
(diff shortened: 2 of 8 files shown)


vertx-pg-client/src/test/java/io/vertx/pgclient/PgPoolTest.java
```diff
@@ -162,4 +179,43 @@ public void testUseAvailableResources(TestContext ctx) {
       pool.close();
     }
   }
+
+  // Scenario:
+  // When a user gets a connection from pgPool some milliseconds behind IDLE expiration time,
+  // sometimes you get the following error:
+  // Fail to read any response from the server, the underlying connection might get lost unexpectedly.
+  @Test(expected = TimeoutException.class)
+  public void checkBorderConditionBetweenIdleAndGetConnection(TestContext ctx) {
```
tsegismont, line 188: This test must be moved to the abstract pool test class because the timeout features applies to all connection pools (not just Pg)

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/ConnectionPool.java
```diff
@@ -95,18 +109,22 @@ public void close() {
 
   private class PooledConnection implements Connection, Connection.Holder  {
 
+    private Vertx vertx;
```
tsegismont, line 112: You could avoid to store this variable, I think

tsegismont (changes requested):
Thank you @pjgg , I made a first pass

gavinking:
Oh, great!

vertx-pg-client/src/test/java/io/vertx/pgclient/PgPoolTest.java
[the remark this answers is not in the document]
pjgg: 500 iterations :), this issue is difficult to reproduce, we need some extra iterations.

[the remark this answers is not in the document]
tsegismont: Big enough to be confident and low enough to make it quick :wink:  This is a unit test.

vertx-pg-client/src/test/java/io/vertx/pgclient/PgPoolTestBase.java
```diff
@@ -190,4 +204,26 @@ public void testCancelRequest(TestContext ctx) {
...
+    poolOptions.setMaxSize(5).setIdleTimeout(idle).setIdleTimeUnit(TimeUnit.MILLISECONDS);
+
+    PgPool pool = createPool(new PgConnectOptions(options), poolOptions);
+
+    System.out.println("###################################################: ");
+    IntStream.range(0, concurrentRequestAmount).forEach(n -> CompletableFuture.runAsync(() ->
+      pool.query("SELECT CURRENT_TIMESTAMP;").execute(ctx.asyncAssertSuccess(rowSet -> {
+        for(Row row: rowSet){
+          async.countDown();
+          System.out.println("Connection #" + async.count() + " " + row.getOffsetDateTime(0));
```
tsegismont, line 225:
```suggestion
```

vertx-pg-client/src/test/java/io/vertx/pgclient/PgPoolTestBase.java
```diff
@@ -190,4 +204,26 @@ public void testCancelRequest(TestContext ctx) {
...
+        }
+      }))));
+  }
```
tsegismont, line 228: After the concurrent query test, we should also make sure no connection was leaked by sending a query to Postgres and inspecting the number of live connections.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/ConnectionPool.java
```diff
@@ -156,6 +172,21 @@ public void handleClosed() {
...
     }
 
+    public void idleStart(){
+      if(idleTimeout > 0) {
+        timerId = context.owner().setTimer(idleTimeout, v -> handleClosed());
+      }
+    }
+
+    public boolean idleStop(){
+      boolean stopped = true;
```
tsegismont, line 182:
```suggestion
```

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/ConnectionPool.java
```diff
@@ -156,6 +172,21 @@ public void handleClosed() {
...
+      if(idleTimeout > 0) {
+        stopped = context.owner().cancelTimer(timerId);
```
tsegismont, line 184:
```suggestion
        return context.owner().cancelTimer(timerId);
```

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/ConnectionPool.java
```diff
@@ -156,6 +172,21 @@ public void handleClosed() {
...
+      }
+
```
tsegismont, line 186:
```suggestion
```

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/ConnectionPool.java
```diff
@@ -156,6 +172,21 @@ public void handleClosed() {
...
+      return stopped;
```
tsegismont, line 187:
```suggestion
      return true;
```

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/ConnectionPool.java
```diff
@@ -97,16 +111,18 @@ public void close() {
 
     private final Connection conn;
     private Holder holder;
+    private long timerId;
 
     PooledConnection(Connection conn) {
       this.conn = conn;
+      if(Objects.isNull(context)) idleTimeout = 0;
```
tsegismont, line 118:
```suggestion
      if(context == null) {
        idleTimeout = 0;
      }
```

tsegismont (changes requested):
Thanks @pjgg for the updates. A few more concerns but we're almost there.

vertx-pg-client/src/test/java/io/vertx/pgclient/PgPoolTestBase.java
pjgg: Do you mean something like this isn't?. If we are not reusing connections, then `pg_stat_activity'` will grow and grow...exceeding the poolSize.

tsegismont: Right, but it is not explicit.

tsegismont (approved):
Thank you @pjgg , great work!

pjgg:
Thank U @tsegismont  !!!

PR #884, opened 2021-02-06
MySQL Client several features overview for 4.1

BillyYccc:
This PR **might not** be directly merged. It tracks some changes for MySQL client `4.1.x`.

* Add support for command pipelining - a PR exists in https://github.com/eclipse-vertx/vertx-sql-client/pull/875 (contains a implementation breaking change in batching so it can' be integrated into `4.0.x`)
* Add support for optional resultset metadata - we might integrate it into `4.0.x`
* `MySQLDecoder` rework - replace `ByteToMessageDecoder` with a `ChannelInboundHandlerAdapter` and simplify the decoding packet logic
* I find it seems unnecessary to rework `MySQLEncoder` to minimize flushes because we can use `FlushConsolidationHandler` instead.
* Some edge minor optimizations

### DIFF
--- a/vertx-mysql-client/src/main/asciidoc/index.adoc
+++ b/vertx-mysql-client/src/main/asciidoc/index.adoc
@@ -129,7 +129,7 @@ Then set the path to the domain socket in {@link io.vertx.mysqlclient.MySQLConne
 {@link examples.MySQLClientExamples#connectWithUnixDomainSocket}
 ----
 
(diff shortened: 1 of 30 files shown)


vietj:
@BillyYccc I converted this PR to draft PR

vietj:
what is the status of this PR @BillyYccc ? does it replaces the other pipelining for MySQL PR ? it is not clear to me

PR #895, opened 2021-02-16
Port pool idle timeout changes from 3.9

tsegismont:
See #883


tsegismont:
In MySQL pool test, I added some code to kill all sessions before verifying idle timeout behavior

PR #898, opened 2021-02-19
Fix MySQL server version parser for whitespaces

TimothyKlim:
Motivation:

Fix "WARNING: Incorrect parsing server version tokens" problem with facebook branch of mysql build from sources.


BillyYccc:
Thanks for the contribution!

PR #928, opened 2021-03-29
Datetimeoffset data type support

tsegismont:


tsegismont:
@vietj @BillyYccc there are 2 tests failures. It happens when testing null values. I have started to investigate and it looks like a bug in our [null row](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/c0da3cc9-4441-476b-aa99-5a518f22dc0a) parser implementation, which is not specific to `datetimeoffset` type nor any other data type.

I will continue investigating tomorrow. In the meantime, any feedback is welcome.

tsegismont:
@vietj @BillyYccc I got back to this and found the issue with NBC row decoding. The PR is ready for review.

vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java
```diff
@@ -518,7 +525,14 @@ default LocalTime getLocalTime(int pos) {
    */
   @GenIgnore(GenIgnore.PERMITTED_TYPE)
   default LocalDateTime getLocalDateTime(int pos) {
-    return (LocalDateTime) getValue(pos);
+    Object val = getValue(pos);
```
vietj, line 528: why was this changed ?

vietj: this introduce a change in the behavior that prevents the user to make mistake, i.e to use a local date time when it is actually an offset date time which does not have the semantic.

vietj:
this PR makes an important change in Tuple that make tests failing for Postgres that requires some change to pass:

e.g

```java
  @Test
  public void testDecodeTimestampTzPlusInfinity(TestContext ctx) {
    OffsetDateTime expected = OffsetDateTime.MAX;
    ColumnChecker checker = ColumnChecker.checkColumn(0, "TimestampTz")
      .returns(Tuple::getValue, Row::getValue, expected)
      .returns(Tuple::getOffsetTime, Row::getOffsetTime, expected.toOffsetTime())
      .returns(Tuple::getOffsetDateTime, Row::getOffsetDateTime, expected)
      .returns(Tuple::getLocalDate, Row::getLocalDate, expected.toLocalDate())
      .returns(Tuple::getLocalTime, Row::getLocalTime, expected.toLocalTime())
      .returns(Tuple::getLocalDateTime, Row::getLocalDateTime, expected.toLocalDateTime())
      .returns(Tuple::getTemporal, Row::getTemporal, expected);
    testDecodeDataTimeGeneric(ctx, "TIMESTAMP WITH TIME ZONE", "TimestampTz", checker, expected);
  }
```

I don't get how this PR could be merged since tests are not passing anymore.

vietj:
clearly the CI indicates a failure 

Screen Shot 2021-04-19 at 10 56 40

vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java
tsegismont: Hey @vietj ! This was changed to align the behavior of `getLocalDateTime()` with that of `getLocalTime()` and `getLocalDate()`. As you said, the purpose is tolerate the use of these methods instead of simply failing in case the type in database also has offset info.

I'm sorry for missing the failures in the pg client module. Thank you for fixing the problem in a follow-up commit.

vietj: Current conversions you are referring to truncate the value, pretty much like you would convert a long to an int.

In this case we are converting an offset value to a local value and loose the zone information which is equivalent of translating to the date to an UTC date and lose contextual information.

vietj: That's why I think that in this particular case, there should be a class cast exception because it is a mis use.

tsegismont: I understand what you mean with truncation of value.

But if the client tolerates using, for example, `getShort()` for a big integer or decimal value, why not tolerate `getLocalTime()` or `getLocalDateTime()` for a zoned/offset datetime?

Our users might be using a legacy DB that has a column defined as zoned/offset datetime but they know it only stores values at UTC (or another fixed offset).

vietj: Calling `getLocalTime()` backed by a `LocalDateTime` will get the `LocalTime` part of the value is similar to getting the  the short value of a big integer (precision lost).

In this case, the zone is used to interpret the actual value and changes the value, e.g when the value is 4PM CEST, then calling `getLocaleTime()` returns the value 2PM.

Users should get a time zoned value from the corresponding time zone accessor.

tsegismont: >Calling getLocalTime() backed by a LocalDateTime will get the LocalTime part of the value is similar to getting the the short value of a big integer (precision lost).

I disagree. If you get the time (or date) part of a datetime, you don't lose precision, you extract a part of the information which has its own meaning.

>In this case, the zone is used to interpret the actual value and changes the value, e.g when the value is 4PM CEST, then calling getLocaleTime() returns the value 2PM.

I think `getLocalTime()` would return 4PM actually. Anyway, my point was the user might want to read data that's stored in a zoned datetime column but they know in advance that all values are on the same timezone (of the country they live in and operate the application from).

If you're not convinced with this use case, think about a user calling `getShort()` on a decimal column because they known in advance all values fit in short type.

If the client tolerates the latter case, why should it fail the former?

vietj: btw did you see that this is an undocumented breaking change ?

vietj: you are right this is not precision loss, but it is similar to calling `getShort` on a `BigInteger`.

That being said, this is a debate of opinion, the main issue is that this change was not discussed before and documented.

tsegismont: > btw did you notice that this is an undocumented breaking change ?

No I didn't. I was not worried about impact as this change leads to code that is more permissive.
How would it break user code?

>That being said, this is a debate of opinion, the main issue is that this change was not discussed before and documented.

OK. I'll wait for you reply to the question above and then I'll send an email to vertx-dev

vietj:
it broke pg-client tests that were relying on the fact this method throws
ClassCastException when this is not a type it can be coerced to. The code
that used to throw CCE is now validly returning an object.

tsegismont:
>it broke pg-client tests that were relying on the fact this method throws
ClassCastException when this is not a type it can be coerced to. The code
that used to throw CCE is now validly returning an object.

Right. Sorry again for missing the failure in pg tests that should have been fixed in the PR originally.

That said, I believe most users don't have expectations on `ClassCastException` being thrown from a user. If they hit already, they probably have adapted their code to avoid it.

I'll go ahead and write to vertx-dev

PR #937, opened 2021-04-16
Connection leak when using pool idle-timeout

### LINKED ISSUE #935: Connection leak when using idle-timeout on PoolOptions
This applies to 3.9.6, 3.9.7 and 4.0.3. It does not apply to 4.1.0 (snapshot with new pool implementation).

When an idle-timeout is set on `PoolOptions`, the pooled connection is not closed properly.
It is removed from the list of available connections but stays open.

On PostgreSQL, this can be observed by inspecting `pg_stat_activity` and finding connections in `IDLE` state.

tsegismont:
Fixes #935

Previously, the connection was removed from the available list but not closed properly.
Consequently, the pool would create new connections and leave idle connections on the backend.


vietj:
I think we should simply remove the pool timeout in 3.9 instead of trying to fix a bug and perhaps introduce new bugs.

PR #946, opened 2021-04-30
Fixes for numeric/decimal support

tsegismont:
The client was able to decode row values but not to encode query parameters.

This commit implements encoding and simplifies decoding.


tsegismont:
@vietj if you have time to take a look at this soon, perhaps we could have it in the beta release?

tsegismont:
Related to #608

tsegismont:
@BillyYccc would you have some time this week to take a look at this PR?

PR #956, opened 2021-05-07
MS SQL numeric/decimal column support with BigDecimal instead of Numeric

tsegismont:
The support for these types was originally implemented using the SQL Client `Numeric` Java type.

The `Numeric` Java type has been created because Postgres `numeric`/`decimal` columns can have the `NaN` value.

MS SQL doesn't have it so we can simplify things and work with `BigDecimal`.


vietj (approved):
LGTM

PR #961, opened 2021-05-20
Add a NullValue API type for internal Tuple storage

tsegismont:
Currently, reactive clients infer Tuple types by inspecting its values. When a value is null, the inferred type is `Object`.
Besides, the MSSQL client will use `varchar` type when building params definition for prepared queries.

This worked so far for the types supported but it won't work in the future for MSSQL types like `binary` or `varbinary`.

This PR adds a NullValue API type that is used for internal `Tuple` storage.

With `NullValue`, it is possible to put a `NULL` value in a `Tuple` and carry along type info.


tsegismont:
@vietj @BillyYccc the PR is still a draft because I would like your input on the design.

When we're all fine with it I'll add some documentation to MSSQL Client.

tsegismont:
@vietj @BillyYccc I updated my branch for MS SQL binary / varbinary support https://github.com/tsegismont/vertx-sql-client/commit/0eafa789a0b68822a072a44caf37403aaaed9dad

You can see how NullValue would be used in this commit

vertx-sql-client/src/main/java/io/vertx/sqlclient/data/NullValue.java
```diff
@@ -0,0 +1,83 @@
...
+import java.time.*;
+import java.time.temporal.Temporal;
+import java.util.UUID;
+
+/**
+ * Instances of this class indicate the type associated with a {@code NULL} column value.
+ * This is useful when infering database types from {@link io.vertx.sqlclient.Tuple} items.
+ */
+@VertxGen
+public interface NullValue {
```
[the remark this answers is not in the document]
tsegismont: Why?

vietj:
I think we actually must, otherwise it will not work for Vert.x RX (and
similar approaches).

With Vert.x RX (or any other similar wrapper approach), then NullValue will
not be unwrapped by the internal class and it will use the RX version of
NullValue instead of the original value.

vertx-sql-client/src/main/java/io/vertx/sqlclient/data/NullValue.java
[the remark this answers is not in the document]
tsegismont: >I think we actually must, otherwise it will not work for Vert.x RX (and
similar approaches).

>With Vert.x RX (or any other similar wrapper approach), then NullValue will
not be unwrapped by the internal class and it will use the RX version of
NullValue instead of the original value.

Right. Let me try this approach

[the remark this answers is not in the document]
tsegismont: See new commit

tsegismont:
@vietj I've updated the PR.

Also, I ran a `vertx-jdbc-client` build locally. It was successful. JDBC Client has a `Tuple` implementation, indeed. But it's a row so it never messes with `NullValue`. Anyway it extends `ArrayTuple` so it's fine.

vertx-sql-client/src/main/java/io/vertx/sqlclient/data/NullValue.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.UUID;
+
+/**
+ * Instances of this class indicate the type associated with a {@code NULL} column value.
+ * This is useful when infering database types from {@link io.vertx.sqlclient.Tuple} items.
+ */
+@DataObject
+public class NullValue {
+
+  public static NullValue Boolean = new NullValue(Boolean.class);
```
vietj, line 33: those constant declarations needs to be final

vertx-sql-client/src/main/java/io/vertx/sqlclient/data/NullValue.java
```diff
@@ -0,0 +1,115 @@
...
+  @GenIgnore
+  public NullValue(Class type) {
+    this.type = type;
+  }
+
+  public String getType() {
+    return type.getName();
+  }
+
+  public void setType(String type) {
```
vietj, line 100: I think we need to use `Class` here and avoid messing with classloading.

vertx-sql-client/src/main/java/io/vertx/sqlclient/data/NullValue.java
```diff
@@ -0,0 +1,115 @@
...
+  private Class<?> type;
+
+  public NullValue() {
+  }
+
+  public NullValue(NullValue other) {
+    type = other.type;
+  }
+
+  public NullValue(JsonObject json) {
```
vietj, line 79: I don't think we need here a JSON constructor

vertx-sql-client/src/main/java/io/vertx/sqlclient/data/NullValue.java
```diff
@@ -0,0 +1,115 @@
...
+import java.time.*;
+import java.time.temporal.Temporal;
+import java.util.UUID;
+
+/**
+ * Instances of this class indicate the type associated with a {@code NULL} column value.
+ * This is useful when infering database types from {@link io.vertx.sqlclient.Tuple} items.
+ */
+@DataObject
+public class NullValue {
```
vietj, line 31: I think we actually should simply drop `@DataObject` and make it a plain class to avoid restriction of data objects, like we did for `Numeric`

vertx-sql-client/src/main/java/io/vertx/sqlclient/data/NullValue.java
```diff
@@ -0,0 +1,115 @@
...
+        this.type = Class.forName(type);
+      } catch (ClassNotFoundException e) {
+        throw new VertxException("Cannot create NullValue instance", e);
+      }
+    }
+  }
+
+  @SuppressWarnings("rawtypes")
+  @GenIgnore
+  public NullValue(Class type) {
```
vietj, line 92: instead we should use a static `of` method to have `NullValue.of(String.class)`

tsegismont:
@vietj PTAL

vertx-sql-client/src/main/java/io/vertx/sqlclient/data/NullValue.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.Objects;
+import java.util.UUID;
+
+/**
+ * Instances of this class indicate the type associated with a {@code NULL} column value.
+ * This is useful when infering database types from {@link io.vertx.sqlclient.Tuple} items.
+ */
+public class NullValue {
+
+  private static final Map<Class<?>, NullValue> cached = new HashMap<>();
```
vietj, line 32: we need to avoid caching such values, users should cache themselves the null value instance (we can document that)

vietj (commented):
We should avoid caching user null values other than the usual types, e.g we can cache JVM types and populate this cache with a static initializer in the class for such JVM types (String, etc...)

vertx-sql-client/src/main/java/io/vertx/sqlclient/data/NullValue.java
tsegismont: We don't cache other `NullValue` instances than those defined in this class (except the `Buffer`/`Temporal` and their array equivalents). Perhaps the name `cache` is confusing and you'd prefer `lookupTable`?

vietj: right sorry for that, then I think we could just have a comment above the cache saying that we only cache JVM types (we can also cache Buffer I think).

besides and that's minor we should avoid caching in the constructor and instead it should be done after the object has been fully initialized (so using a static private method).

tsegismont:
@vietj PTAL

PR #964, opened 2021-05-26
MSSQL binary/varbinary data type support

tsegismont:
MSSQL `binary`/`varbinary` data type support using Vert.x `Buffer`.

### DIFF
--- a/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedQueryCommandCodec.java
+++ b/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedQueryCommandCodec.java
@@ -13,6 +13,7 @@ package io.vertx.mssqlclient.impl.codec;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.ChannelHandlerContext;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.mssqlclient.impl.protocol.MessageStatus;
 import io.vertx.mssqlclient.impl.protocol.MessageType;
 import io.vertx.mssqlclient.impl.protocol.TdsMessage;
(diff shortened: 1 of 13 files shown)


vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedQueryCommandCodec.java
```diff
@@ -476,4 +479,15 @@ private void encodeDecimalParameter(ByteBuf payload, BigDecimal value) {
       payload.writeByte(bytes[i]);
     }
   }
+
+  private void encodeBufferParameter(ByteBuf payload, Buffer value) {
+    payload.writeByte(0x00);
```
vietj, line 484: we should consider minimizing ByteBuf write operations, each write has a cost with respect to bound checks and also often ByteBuf have nested buffers

vietj: so here instead of 3 write we could have a single write with `0x00_00_00 & MSSQLDataTypeId.BIGBINARYTYPE_ID`

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/MSSQLDataTypeCodec.java
```diff
@@ -168,6 +176,13 @@ private static CharSequence decodeNVarchar(ByteBuf in) {
     return in.readCharSequence(length, StandardCharsets.UTF_16LE);
   }
 
+  private static Buffer decodeBinary(ByteBuf in) {
+    int length = in.readUnsignedShortLE();
+    byte[] bytes = new byte[length];
```
vietj, line 181: I think we could directly here create `Unpooled.buffer(length)` and then copy to it with readBytes and then wrap it with `Buffer.buffer(ByteBuf)`

vietj (changes requested):
code looks correct, a few remarks

- there is no actual documentation about how to handle null binaries with `NullValue`
- we should consider minimizing the number of ByteBuf write operations because they are not free (you can trace Netty write code to see the actual work)

so

- request change : documentation
- whished change : optimize writes, but I know there are other places in the MSSQL client so it can be defered

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedQueryCommandCodec.java
tsegismont: Filed https://github.com/eclipse-vertx/vertx-sql-client/issues/965

tsegismont:
@vietj PTAL

vietj (approved):
LGTM

PR #966, opened 2021-05-26
Release TdsMessage after it's been decoded

### LINKED ISSUE #926: Check MSSQL decoder for leaks
See https://github.com/eclipse-vertx/vertx-sql-client/commit/8e577af6238cb5538b9dea52164645106ccdb13a

tsegismont:
Closes #926

Previously, TdsMessageDecoder was a MessageToMessageDecoder which releases the message after decode method.

We never noticed any issues because no test in our test suite leads to a server response that is big enough to be split into different packets.
So the message was always created from newTdsMessageFromSinglePacket and released automatically by Netty.

Now TdsMessageDecoder is a simple ChannelInboundHandlerAdapter and we release the TdsMessage manually after the messages is effectively decoded or the channel handler is removed.


tsegismont:
@vietj I ran the test suite with Netty's leak detector set to PARANOID. No leak was detected.

vietj (approved):

Issue #967, opened 2021-05-27
Cursor cannot be used in a different context.


rasmusfaber:
### Version

3.9.6, 3.9.8-SNAPSHOT ~~and master (and probably others)~~.

### Context

If you try using the RowStream created by PreparedStatement.createStream() in another context than the one it was created in, it fails when it reaches the end.

It was a bit difficult to reproduce it using only vertx, but the problem appears immediately when trying to return data using Reactive Resteasy with Mutiny.

### Do you have a reproducer?

The test case here: https://github.com/rasmusfaber/vertx-sql-client/blob/3.9/vertx-pg-client/src/test/java/io/vertx/pgclient/PreparedStatementTestBase.java#L257

fails if you run it without the fix here
https://github.com/rasmusfaber/vertx-sql-client/blob/3.9/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/PreparedStatementImpl.java#L201

rasmusfaber:
(Pull-requests incoming)

vietj:
out of curiosity how does this happen with resteasy reactive ? is it because of mutiny ?

vietj:
as far as I can see, the cursor could be used from different HTTP requests that would not have the same context ?

rasmusfaber:
I believe what happens is this (there are many layers involved here, so it can be hard to track it all):

Mutiny allows you to convert a ReadStream<T> into a Multi<T>. When it does that, it pauses the stream and then calls fetch on it when a subscriber requests data.

RestEasy Reactive allows you to return a Multi<String> from your rest methods. It then has the http-handler subscribe to the output. But the http-handler is running in a different context than the one that called the rest method, so it is run in a different context, which then resulted in the exception (which incidentally was just swallowed, so you just see your rest method not returning).

If you are interested, I can probably put a small project together which reproduces it in the Mutiny/Resteasy context.

But it boils down to something like this (this is the code, I would like to simplify a bit by boiling it into the framework - I will get back to you with the sample code as soon as I get a bit more time, my first ideas are not quite working out):

```
    private io.vertx.mutiny.pgclient.PgPool client;

    @GET
    @Path("test")    
    public Multi<String> restMethod() {
		return SqlClientHelper.usingConnectionMulti(client,
				conn -> {
					Transaction transaction = conn.begin();
					return Multi.createBy().concatenating().streams(
							conn.prepare("SELECT foo FROM bar").onItem().transformToMulti(ps ->
[...]

PR #968, opened 2021-05-27
Check context when closing cursor

rasmusfaber:
Motivation:

Fix for https://github.com/eclipse-vertx/vertx-sql-client/issues/967

Conformance:

Check.


vietj:
for master this is likely not the right fix because of the way we handle communication between the client and the internal socket

vietj:
do you have a test for this issue ?

rasmusfaber:
Yes, sorry. I can't reproduce it on master after all. It looks like the issues were fixed with https://github.com/eclipse-vertx/vertx-sql-client/commit/9d33256cc1b49c5c9a0cb794eebf1a3cb8991fe6#diff-0de8b326c2028289e736255bc6868132071c24d1df94694cd700f63bde274ab2R128

vietj:
@rasmusfaber did you check whether the issue happens with master ? normally master handles correctly different contexts

Issue #963, opened 2021-05-26
Table created with non-nullable column even without `NOT NULL` in column definition


tsegismont:
When using the MS SQL Client for `CREATE TABLE` statements, the resulting columns are non-`nullable` even when `NOT NULL` is not specified on the column definition.

This does not happen when using the JDBC client on the same database.

```java
PreparedStatement create = conn.prepareStatement("create table Basic (id int, dessimal numeric(19,2), primary key (id))");
create.executeUpdate();
PreparedStatement insert = conn.prepareStatement("INSERT INTO Basic (id, dessimal) values (3, ?)");
insert.setBigDecimal(1, new BigDecimal("3.2"));
insert.executeUpdate();
```

The snippet above works fine but this one fails:

```java
connnection.preparedQuery("create table Basic2 (id int, dessimal numeric(19,2), primary key (id))")
  .execute(ctx.asyncAssertSuccess(create -> {
    connnection.preparedQuery("INSERT INTO Basic2 (id, dessimal) values (3, @p1)")
      .execute(Tuple.of(NullValue.BigDecimal), ctx.asyncAssertSuccess());
  }));
```

The error message is:

```
io.vertx.mssqlclient.MSSQLException: {number=515, state=2, severity=16, message='Cannot insert the value NULL into column 'dessimal', table 'master.dbo.Basic2'; column does not allow nulls. INSERT fails.', serverName='0ac9600ef0bb', lineNumber=1}
```

As a workaround, the column definition must explicitly allow `NULL` values:

```sql
create table Basic2 (id int, dessimal numeric(19,2) null, primary key (id))
```

PR #970, opened 2021-06-01
Prepare MS SQL Server doc before 4.1 release

tsegismont:


gavinking:
@tsegismont does this mean it's near done?

tsegismont:
@gavinking it depends on what you mean by done :-)

It's ready for a tech preview but there are important [limitations](https://github.com/eclipse-vertx/vertx-sql-client/blame/master/vertx-mssql-client/src/main/asciidoc/index.adoc#L16)

gavinking:
Well I mean "done" as in "can we include it as a feature of HR yet?" :-)

tsegismont:
@gavinking when I last tried my HR branch for Vert.x 4 + MS SQL Client, I got 77% tests passing (the rest are failed or ignored).

We can reach a better coverage by:

- updating HR to handle `NULL` parameters correctly https://vertx.io/docs/vertx-mssql-client/java/#_handling_null
- implementing batch queries in the client
- fixing #963 that breaks HR's schema setup feature

I would like to do this before Vert.x 4.1.1 is released.

This is not a direct answer to your question obviously but I hope it helps ;-)

gavinking:
> updating HR to handle NULL parameters correctly https://vertx.io/docs/vertx-mssql-client/java/#_handling_null

Well, wouldn't it be better to fix the SQL Server client so that it behaves like the three other Vert.x drivers and doesn't have this requirement?

I mean, it's something I always *hated* in JDBC, and I was very happy to see that the Vert.x drivers did not suffer from this.

tsegismont:
@gavinking we added it on purpose, and I can assure you we'd have avoided it if we could.

In short, SQL Server requires parameter type definitions when preparing queries. When a parameter is `null`, we can default to `nvarchar` and that works for many types... but not all (e.g. `varbinary`).

gavinking:
Well that sucks, but OK, fine. Let me know if this does the trick: https://github.com/hibernate/hibernate-reactive/pull/845.

PR #975, opened 2021-06-09
MS SQL Connection Options: same as JDBC

### LINKED ISSUE #963: Table created with non-nullable column even without `NOT NULL` in column definition

tsegismont:
Fixes #963

In particular, ODBC_ON implies ANSI_DEFAULTS which activates the following ISO behaviors:

- ANSI_NULLS
- CURSOR_CLOSE_ON_COMMIT
- ANSI_NULL_DFLT_ON
- IMPLICIT_TRANSACTIONS
- ANSI_PADDING
- QUOTED_IDENTIFIER
- ANSI_WARNINGS

### DIFF
--- a/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/InitCommandCodec.java
+++ b/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/InitCommandCodec.java
@@ -86,9 +86,17 @@ class InitCommandCodec extends MSSQLCommandCodec<Connection, InitCommand> {
     packet.writeIntLE(0x00); // ClientProgVer
     packet.writeIntLE(0x00); // ClientPID
     packet.writeIntLE(0x00); // ConnectionID
-    packet.writeByte(LoginPacket.DEFAULT_OPTION_FLAGS1
-      | LoginPacket.OPTION_FLAGS1_DUMPLOAD_OFF); // OptionFlags1
-    packet.writeByte(LoginPacket.DEFAULT_OPTION_FLAGS2); // OptionFlags2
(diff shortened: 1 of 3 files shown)


gavinking:
Excellent.

PR #982, opened 2021-06-11
MSSQL: Fix datetime2 encoding

### LINKED ISSUE #977: [mssql] Incorrect encoding of datatime2 for LocalDateTime values to the ms
With MS Sql Server.

Testcase to add to `MSSQLPreparedQueryNotNullableDataTypeTest`
```
  @Test
  public void testEncodeDateTimeWithMillis(TestContext ctx) {
    final LocalDateTime start = LocalDateTime.now().truncatedTo(ChronoUnit.MILLIS);
    testPreparedQueryEncodeGeneric(ctx, "not_nullable_datatype", "test_datetime2", start, row -> {
      ColumnChecker.checkColumn(0, "test_datetime2")
        .returns(Tuple::getValue, Row::getValue, start)
        .returns(Tuple::getLocalDateTime, Row::getLocalDateTime, start)
        .returns(Tuple::getLocalDate, Row::getLocalDate, start.toLocalDate())
        .returns(Tuple::getLocalTime, Row::getLocalTime, start.toLocalTime())
        .returns(LocalDateTime.class, start)
        .forRow(row);
    });
  }
```

Result:
```
junit.framework.AssertionFailedError: Expected that public abstract java.lang.Object io.vertx.sqlclient.Tuple.getValue(int) would not fail: 
Expected :2021-06-09T17:20:40.643
Actual   :2021-06-09T17:31:23
```

tsegismont:
Fixes #977

LocalTime, LocalDateTime and OffsetDateTime have nanosecond precision.
Also, the maximum scale in MSSQL is 7 (hundreds of nanos).

So when encoding time-related values, we will always send a value with maximum scale.
The server will do round/truncate as necessary.

### DIFF
--- a/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedQueryCommandCodec.java
+++ b/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedQueryCommandCodec.java
@@ -285,11 +285,11 @@ class ExtendedQueryCommandCodec<T> extends QueryCommandBaseCodec<T, ExtendedQuer
     } else if (value instanceof LocalDate) {
       encodeDateNParameter(payload, (LocalDate) value);
     } else if (value instanceof LocalTime) {
-      encodeTimeNParameter(payload, (LocalTime) value, (byte) 6);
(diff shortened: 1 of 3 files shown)


DavideD:
Awesome! Thanks

Issue #242, opened 2019-02-03
Provide a method for the connection pool to define an initial handler for every new created connection


BillyYccc:
like `PgPool#connectionInitHandler` so that users can set an handler and for instance execute some sqls for every new created connection.
Motivation from https://github.com/reactiverse/reactive-pg-client/issues/234.

vietj:
Hi,

in this case I would rather provide a factory for connection, i.e

static PgPool pool(PgConnectOptions options, Function<PgConnectOptions, Future<PgConnection> factory) { ... }

where factory is called is called to create a connection, so user can perform this.

pilzm:
Hi there,
We have a scenario where we need to regenerate passwords periodically (tokens with expiration time). Am I right in assuming this will be possible with this proposed change?

vietj:
@pilzm no it will not be possible, this PR aims to perform some work after a connection has been established. I think what you are looking for is rather creating a pool from a `Supplier`, e.g `PgPool.pool(() -> new ConnectOptions().setUsername("foo").setPassword("bar"))`

pilzm:
@vietj sounds exactly like what I need :) Should I just create a new `enhancement` issue?

vietj:
yes please

tsegismont:
Closed in #986

PR #986, opened 2021-06-16
Connection pool improvements

vietj:
Pool enhancements.

See #242


vietj:
@Rattenkrieg ping

vietj:
this is almost ready for review

vertx-sql-client/src/main/java/io/vertx/sqlclient/PoolConfig.java
```diff
@@ -0,0 +1,129 @@
...
+   * <p>
+   *
+   * When the {@code serverProvider} fails to return a value, the pool falls back to the {@code base} options.
+   *
+   * @param base the base options for all servers
+   * @param serverProvider the list of servers
+   * @return a reference to this, so the API can be used fluently
+   */
+  @GenIgnore
+  <C extends SqlConnectOptions> PoolConfig connectingTo(C base, Supplier<C> serverProvider);
```
robotmrv, line 102: In our case getting connection options could involve io, is it possible to make it async friendly, i.e. provide `Supplier<Future<C>>` or something similar?

vietj: I think that would be too complicated, instead the supplier should communicate asynchronously with something else that retrieves the options, e.g the options are set in a volatile field and the provider reads that field

robotmrv: yes, it would be more complicated  
but without it user have to wrap Pool and all its methods that could create a new connection and add its async logic around that would refresh `SqlConnectOptions` and retry. (at least this would happen in my case since credentials storage does not provide notifications mechanism...)  

As alternative it could be added a hook that would be invoked before an attempt of creating a new connection. Would it be  possible? 

Anyway `Supplier<C>` is better than nothing, at least now it is possible to not recreate entire Pool.

vietj: I will try to use a `Supplier<Future<>>` as it seems legit

vietj: now it uses a future, can you have a look ?

robotmrv: looks great, thanks

vietj:
The API has been changed to get away of the `PoolConfig` that was not really necessary and instead use the `Pool`, the `Supplier<SqlConnectOptions>` is replaced by `Function<Context, SqlConnection>`.

PR #988, opened 2021-06-16
Allow to configure database container with a fixed port

tsegismont:
This simplifies debugging with tools like wireshark.


tsegismont:
@vietj @BillyYccc tell me if you'd like this for Pg and MySQL too.

vietj:
yes please @tsegismont it's a blessing to have such feature

tsegismont:
I didn't do it for MySQL because some tests don't pass when the flag is present. Not sure why

PR #991, opened 2021-06-18
bump pg to 13 in tests

### LINKED ISSUE #877: PostgreSQL 12 and 13 support
Any plans on supporting PostgreSQL 12 and 13?

Rattenkrieg:
Seems there is nothing new in pg 13 to break vertx-pg-client. Only a validation for line type added. Also pg13 refuses to start with 1024 bit key making `TLSTest` failing, so I generated a 2048 bit key locally to verify this tests. I guess maintainers would like to keep their own key in this repository so it's up to them to generate a longer one.
Fixes #877

### DIFF
--- a/vertx-pg-client/src/main/java/io/vertx/pgclient/data/Line.java
+++ b/vertx-pg-client/src/main/java/io/vertx/pgclient/data/Line.java
@@ -11,11 +11,8 @@ public class Line {
   private double b;
   private double c;
 
-  public Line() {
-    this(0.0, 0.0, 0.0);
-  }
-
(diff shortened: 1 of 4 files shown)


vietj:
can you break this PR in two because it makes two different things and have for each a proper description of the change

Rattenkrieg:
@vietj you mean one PR with changes to `Line` and another one with pg version bump, right? Sure then.

vietj:
yes, and describe the changes, that's better even those seems to be one
liners

the line change should be tested also

PR #994, opened 2021-06-21
TDS packet splitting support

### LINKED ISSUE #758: MSSQL support packet size configuration
The server sends an info token with the desired packet size.

Currently we send messages in a single packet, regardless of the desired packet size.

tsegismont:
Fixes #758
    
When the client initializes the connection (LOGIN7), it sends a desired size for TDS packets.
    
The server may or may not accept it. In the response, it will give the packet size it wants and we must enforce it. This value may changed with an ENVCHANGE token later.


tsegismont:
@vietj can you please review? The PR is divided in two commits that can be reviewed independently:

1. refactoring of TDS codec
2. TDS packet splitting support

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLConnectOptions.java
```diff
@@ -119,6 +117,30 @@ public MSSQLConnectOptions addProperty(String key, String value) {
...
+   * Get the desired size for TDS packets.
+   *
+   * @return the desired packet size
+   */
+  public int getPacketSize() {
+    return packetSize;
+  }
+
+  /**
+   * Set the desired size for TDS packets.
```
vietj, line 130: in bytes

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedQueryCommandBaseCodec.java
```diff
@@ -44,158 +36,76 @@ void encode(TdsMessageEncoder encoder) {
...
+    super.handleResultSetDone();
     rowCount = 0;
   }
 
-  protected abstract void handleMessageDecoded();
+  @Override
+  protected void handleReturnValue(ByteBuf payload) {
+    MSSQLPreparedStatement ps = (MSSQLPreparedStatement) cmd.preparedStatement();
+    if (ps.handle == 0) {
+      payload.skipBytes(2); // skip ordinal position
```
vietj, line 55: consider calling this a single time, it looks cute but in practice it costs (it falls in the same area as `write`)

vietj: e.g you could do something like

```java
int bytesToSkip = 2 // sky ordinal position
                           + 2 * payload.readUnsignedByte() // skip param name
                           ...;
```

vietj:
are you able to consistently test this behavior ?

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLConnectOptions.java
tsegismont:
```suggestion
   * Set the desired size (in bytes) for TDS packets.
```

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLConnectOptions.java
```diff
@@ -119,6 +117,30 @@ public MSSQLConnectOptions addProperty(String key, String value) {
     return (MSSQLConnectOptions) super.addProperty(key, value);
   }
 
+  /**
+   * Get the desired size for TDS packets.
```
tsegismont, line 121:
```suggestion
   * Get the desired size (in bytes) for TDS packets.
```

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLConnectOptions.java
```diff
@@ -119,6 +117,30 @@ public MSSQLConnectOptions addProperty(String key, String value) {
...
+    return packetSize;
+  }
+
+  /**
+   * Set the desired size for TDS packets.
+   * <p>
+   * The client will use the value as a parameter in the LOGIN7 packet.
+   * The server may or may not accept it.
+   *
+   * @param packetSize the desired packet size
```
tsegismont, line 135:
```suggestion
   * @param packetSize the desired packet size (in bytes)
```

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedQueryCommandBaseCodec.java
tsegismont: Done

tsegismont:
@vietj PTAL

> are you able to consistently test this behavior ?

In `MSSQLMultipleTdsPacketsTest` there's a test that executes a large batch query. The payload is too large to fit into a single packet of default size. Same for the result.

vietj:
that is what I thought but I was not sure.

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLConnectOptions.java
```diff
@@ -130,6 +128,30 @@ public MSSQLConnectOptions addProperty(String key, String value) {
...
+   * Set the desired size (in bytes) for TDS packets.
+   * <p>
+   * The client will use the value as a parameter in the LOGIN7 packet.
+   * The server may or may not accept it.
+   *
+   * @param packetSize the desired packet size (in bytes)
+   * @return a reference to this, so the API can be used fluently
+   * @throws IllegalArgumentException if {@code packetSize} is smaller than {@link #MIN_PACKET_SIZE} or bigger than {@link #MAX_PACKET_SIZE}
+   */
+  public MSSQLConnectOptions setPacketSize(int packetSize) {
```
vietj, line 150: missing bound check advocated by the `@throws` clause

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/MSSQLConnectionFactory.java
```diff
@@ -28,13 +30,16 @@
 
 public class MSSQLConnectionFactory extends ConnectionFactoryBase {
 
+  private int packetSize;
```
vietj, line 33: final

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedBatchQueryCommandCodec.java
```diff
@@ -26,22 +26,22 @@
...
       return;
     }
-    super.encode(encoder);
+    super.encode();
   }
 
   @Override
-  protected void handleMessageDecoded() {
+  protected void handleDecodingComplete() {
     if (paramsList.size() == 1 || ++messageDecoded == 2) {
```
vietj, line 45: ++ will only be applied when `paramsList.size() != -1` is that the correct thing that is to do here ? i.e is it equivalent to

```java
if (paramsList.size() == 1) {
   ...
} else {
  messageDecoded++;
  if (messageDecoded == 2) {
    ...
  } else {
    ...
  }
}
```

vietj: it is not part of this review but a general remark when seing this

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/QueryCommandBaseCodec.java
```diff
@@ -18,53 +18,40 @@
...
 
+  @Override
   protected void handleNbcRow(ByteBuf payload) {
     rowResultDecoder.handleNbcRow(rowResultDecoder.desc.columnDatas.length, payload);
   }
 
-  protected void handleResultSetDone(int affectedRows) {
+  @Override
+  protected void handleAffectedRows(long count) {
+    rowCount += count;
```
vietj, line 51: we could check `count >= 0` to catch programming error at this point

vietj (changes requested):
A few comments

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedBatchQueryCommandCodec.java
tsegismont: Yes, it is on purpose and equivalent to the snippet above.

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/QueryCommandBaseCodec.java
tsegismont: There are a lot of other assertions we could add. As an example, when decoding parameters, we could check that scale/length are consistent with the `INT` type. I thought about it but couldn't see the value add.

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLConnectOptions.java
tsegismont: Done

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/MSSQLConnectionFactory.java
tsegismont: Done

tsegismont:
@vietj PTAL

vietj (approved):

PR #1017, opened 2021-07-30
MSSQL Client Encryption support

tsegismont:
See individual commits for easier review

- [x] encryption level negotiation
- [x] encryption implementation
- [x] manual testing with Docker
- [x] manual testing with Azure database (see [comment](https://github.com/eclipse-vertx/vertx-sql-client/pull/1017#issuecomment-896565891))
- [x] automated testing with Docker
- [x] documentation


vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/TdsSslHandshakeCodec.java
```diff
@@ -0,0 +1,106 @@
...
+    @Override
+    protected Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception {
+      ByteBuf byteBuf = (ByteBuf) super.decode(ctx, in);
+      if (byteBuf == null) {
+        return null;
+      }
+
+      int length = in.getUnsignedShort(2);
+
+      return byteBuf.slice(PACKET_HEADER_SIZE, length - PACKET_HEADER_SIZE);
```
vietj, line 52: as a slice is returned, it is not clear when the original `ByteBuf` will be released as we don't keep a reference on it.

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/TdsSslHandshakeCodec.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  private static class Encoder extends ChannelOutboundHandlerAdapter {
+
+    private ByteBufAllocator alloc;
+    private CompositeByteBuf accumulator;
```
vietj, line 59: it does not seem that the accumulator is released on a failure of the pipeline

vietj (commented):
a few concerns about potential buffer leaks

tsegismont:
Thanks @vietj , I will address the concerns asap

tsegismont:
### Manual testing with Azure results

Connecting from the outside world works but not from Azure Web App Service.

Did some debugging and it seems this happens because our client doesn't support the ROUTING token of a login response, unlike the JDBC client.

So this PR only partially fix the issue with connecting to Azure SQL servers.

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/TdsSslHandshakeCodec.java
tsegismont: The slice instance `release` method is a delegate to the `release` method of the original `ByteBuf`. So when the slice is released (by the `SslHandler`, which comes next in the pipeline) the original `ByteBuf` will be released too.

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/TdsSslHandshakeCodec.java
tsegismont: I thought this case was handled by the `handlerRemoved` method. Do you have another case in mind?

vietj: right I missed that

vietj (approved):

tsegismont:
@vietj PTAL, I added a final commit with doc + examples.

tsegismont:
@vietj do you want to review the last commit with documentation or should I go ahead and merge?

vietj (approved):
LGTM, however I think we should summarize the SSL setup interactions somewhere (javadoc is fine) so anyone debugging can understand how the piece interact together

tsegismont:
good point, I will do that and merge

PR #1029, opened 2021-08-24
Enhance Sql Client SPI

### LINKED ISSUE #1028: Enhance SQL Client SPI
Enhance SQL Client Driver SPI

Hi,

I think the current way SQL Client Drivers are exposed through Service Provider Interfaces presents some problems and should be improved.

I see actually several issues with current Driver interface which prevents the SPI to fullfil its purpose:

1. There is no way to set product specific connection options since these must be set on a `SqlConnectOptions` implementation specific to the driver which is usually not accessible when using the SPI. Basically, if we need to have vertx-pg-client on the classpath, the SPI becomes less interesting and we can directly manipulate Pg* types to create a SQL client.

2. There is no way to pick the right driver corresponding to a static configuration (eg. uri, properties, json...). Using a SPI we expect to be able to pick a driver that would understand a particular abstract configuration. Currently the Driver defines the `acceptsOptions(SqlConnectOptions connectOptions)` which is of little interest since all implementations returns true when a SqlConnectOptions instance is provided. So for this to be useful we must supply a specific `SqlConnectOptions` instance, in which case we could also directly create the specific SQL client and ignore the SPI.

The `Pool#pool(SqlConnectOptions database, PoolOptions options)` method provides a way to pick the right implementation based on the type of `SqlConnectOptions` but this actually doesn't solve the issue.

3. `Driver#createPool()`, `Driver#createConnectionFactory()` and `Driver#acceptsOptions()` implementations are not consistent. All implementations actually wraps options in `createPool()` and `createConnectionFactory()` without checking whether the options is accepted by the driver. We can for instance invoke `PgDriver#createPool()` with a `DB2ConnectOptions` instance that will eventually be wrapped in a `PgConnectOptions` whereas `PgDriver#acceptsOptions()` returns false since the driver only accept `PgConnectOptions` and `SqlConnectOptions` instances.

The purpose of the SPI is to be able to create a SQL client instance from an abstract static configuration, we basically don't know which data store we want to connect to. The client code should never be aware of the actual implementation, the Driver implementation to use should be given by configuration in a URI scheme for example.

Issue #767 describe a similar issue, I propose to solve these issues without dispersing the SPI by exposing methods based on a connection URI in the sole `Driver` interface in order to abstract `SqlConnectOptions` implementation as well as `Pool` and `ConnectionFactory`.

We could expose the following methods in Driver:
```
// Determines whether the driver support the connection URI based on the scheme for instance
boolean acceptsUri(String connectionUri);

// Returns a specific SqlConnectOptions instance created from the connection URI
<T extends SqlConnectOptions> T getConnectOptions(String connectionUri);

// Returns a specific SqlConnectOptions instance created from the connection URI and populated with the json
<T extends SqlConnectOptions> T getConnectOptions(String connectionUri, JsonObject json);
```

A `DriveLoader` class in the spi package could also be created to implement the `ServiceLoader` logic and expose drivers. I'm not sure about the `Pool#pool(SqlConnectOptions database, PoolOptions options)` which imho is part of the SPI but not in an obvious way, I'll prefer to say that the `Driver` interface and more generally the SPI package are the main entry points rather than dispersing the SPI logic in the SQL client API

This represents some changes in the API and shall surely be discussed. I'm making a corresponding PR right now to make things more concrete, please let me know what you think.

jkuhn1:
Motivation:

Fix #1028

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/spi/Driver.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/spi/Driver.java
@@ -16,14 +16,15 @@
  */
 package io.vertx.sqlclient.spi;
 
+import java.util.List;
+
 import io.vertx.core.Vertx;
+import io.vertx.core.json.JsonObject;
 import io.vertx.sqlclient.Pool;
 import io.vertx.sqlclient.PoolOptions;
 import io.vertx.sqlclient.SqlConnectOptions;
 
-import java.util.List;
-
-/**
+/**<
  * An entry point to the Vertx Reactive SQL Client
  * Every driver must implement this interface.
  */
@@ -52,5 +53,32 @@ public interface Driver {
    * @return true if the driver accepts the {@code connectOptions}, false otherwise
    */
   boolean acceptsOptions(SqlConnectOptions connectOptions);
-
+  
+  /**
+   * @param connectionUri a connection URI
+   * @return true if the driver accepts the {@code uri}, false otherwise
+   */
+  boolean acceptsUri(String connectionUri);
+  
+  /**
+   * Returns a specific {@link SqlConnectOptions} instance corresponding to the {@code uri}.
+   * 
+   * @param <T> the expected type of {@code SqlConnectOptions}
+   * @param connectionUri a connection URI
+   * @return connect options specific to the driver
+   * @throws IllegalArgumentException if the driver does not accept the {@code uri}
+   */
+  <T extends SqlConnectOptions> T getConnectOptions(String connectionUri);
+  
+  /**
+   * Returns a specific {@link SqlConnectOptions} instance corresponding to the {@code uri} and 
+   * initialized with {@code json}.
+   * 
+   * @param <T> the expected type of {@code SqlConnectOptions}
+   * @param connectionUri a connection URI
+   * @param json a json
+   * @return connect options specific to the driver
+   * @throws IllegalArgumentException if the driver does not accept the {@code uri}
+   */
+  <T extends SqlConnectOptions> T getConnectOptions(String connectionUri, JsonObject json);
 }
--- a/vertx-db2-client/src/main/java/io/vertx/db2client/spi/DB2Driver.java
+++ b/vertx-db2-client/src/main/java/io/vertx/db2client/spi/DB2Driver.java
@@ -17,6 +17,7 @@ package io.vertx.db2client.spi;
 
 import io.vertx.core.Vertx;
 import io.vertx.core.impl.VertxInternal;
+import io.vertx.core.json.JsonObject;
 import io.vertx.db2client.DB2ConnectOptions;
 import io.vertx.db2client.DB2Pool;
 import io.vertx.db2client.impl.DB2ConnectionFactory;
@@ -29,6 +30,8 @@ import io.vertx.sqlclient.spi.ConnectionFactory;
 import java.util.List;
 
 public class DB2Driver implements Driver {
+
+  private static final String ACCEPT_URI_REGEX = "(db2)://.*";
 
   @Override
   public DB2Pool createPool(Vertx vertx, List<? extends SqlConnectOptions> databases, PoolOptions options) {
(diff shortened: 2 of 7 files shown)


vertx-sql-client/src/main/java/io/vertx/sqlclient/spi/Driver.java
```diff
@@ -52,5 +53,32 @@
...
+  
+  /**
+   * Returns a specific {@link SqlConnectOptions} instance corresponding to the {@code uri}.
+   * 
+   * @param <T> the expected type of {@code SqlConnectOptions}
+   * @param connectionUri a connection URI
+   * @return connect options specific to the driver
+   * @throws IllegalArgumentException if the driver does not accept the {@code uri}
+   */
+  <T extends SqlConnectOptions> T getConnectOptions(String connectionUri);
```
vietj, line 71: I can't see where this is used in this PR

vertx-sql-client/src/main/java/io/vertx/sqlclient/spi/Driver.java
```diff
@@ -52,5 +53,32 @@
...
+   * Returns a specific {@link SqlConnectOptions} instance corresponding to the {@code uri} and 
+   * initialized with {@code json}.
+   * 
+   * @param <T> the expected type of {@code SqlConnectOptions}
+   * @param connectionUri a connection URI
+   * @param json a json
+   * @return connect options specific to the driver
+   * @throws IllegalArgumentException if the driver does not accept the {@code uri}
+   */
+  <T extends SqlConnectOptions> T getConnectOptions(String connectionUri, JsonObject json);
```
vietj, line 83: we should try to remove this as it always do the same thing

vertx-sql-client/src/main/java/io/vertx/sqlclient/spi/Driver.java
jkuhn1: Hi,

In order to create a `Pool` or a `ConnectionFactory` using `createPool()` or `createConnectionFactory()`, we basically need to be able to create specific `SqlConnectOptions` (eg. `PgConnectOptions`). So the normal flow to use a `Driver` would be:

```java
Vertx vertx = Vertx.vertx();

// From configuration file
String connectionUri = ...
JsonObject dbConnectOptionsJson = ...

// Get instance corresponding to the connection URI using DriverLoader
Driver driver = DriverLoader.getDriver(connectionUri);

// Load specific Sql connect options
SqlConnectOptions dbConnectOptions = optionsdriver.getConnectOptions(connectionUri, dbConnectOptions);

// Get specific connection factory
ConnectionFactory connectionFactory = driver.createConnectionFactory(vertx, dbConnectOptions);
```
That way we can create a code that only depends on vertx-sql-client and which can create a `Pool`or a `ConnectionFactory` for a specific database assuming the proper Driver is on the classpath at runtime (eg. vertx-pg-client).

I hope this clarifies, this should go to unit tests by the way :-)

vertx-sql-client/src/main/java/io/vertx/sqlclient/spi/Driver.java
jkuhn1: Ok now that that I got back into it, I get it. Yes indeed `connectionUri` and `JsonObject` does more or less the same thing so either we decide to use them together (merge) for convenience as I suggested or we can use one or the other defining two methods. You're probably right it seems more consistent to do that way.

vertx-sql-client/src/main/java/io/vertx/sqlclient/spi/Driver.java
vietj: users should not interact with the driver class directly

vietj: only frameworks are intended to be

vietj:
@jkuhn1 can you look at https://github.com/eclipse-vertx/vertx-sql-client/tree/generic-connect-options-parsing

I think it solves the generic fashion.

jkuhn1:
@vietj sorry, the link seems dead (404).

Following your comments, I understand that you expect a regular user to use `Pool.pool()` instead of the SPI. Although I don't totally agree that only frameworks should interact with the `Driver` class, I understand and respect your position. As you have understood, my main issue is related to the `SqlConnectOptions` which are typed for each DB providers, this makes it difficult to have a proper abstraction and be able to switch DB providers by configuration.

What I proposed in this PR was to delegate the creation of the specific `SqlConnectOptions` instance to the `Driver` using untyped URI or JSON. This can then be used within `Pool` (hidden from regular users) to create a specific pool instance while being able to apply provider specific configuration, we can for instance create methods `Pool#pool(String connectionUri)` and `Pool#pool(JsonObject json)` and use the SPI behind the scene (this is already the case).

To give you some background, I'm currenty developing the [Inverno framework](https://inverno.io/), Vertx sql client is non-blocking and by far the most performant (good job by the way), so I normally choosed to integrate it behind a Reactor 3 API for DB access. Here's what I needed to do to be able to initialize a pool in an *abstract* way using untyped configuration (JSON): https://github.com/inverno-io/inverno-mods/blob/master/inverno-sql-vertx/src/main/java/io/inverno/mod/sql/vertx/internal/PoolSqlClientWrapper.java. It was not possible to actually use the SPI and the service loader because of the typed options. Apart from the fact that I now have optional dependencies on all implementations, supporting a new implementation (eg. Oracle) requires code update. This is framework, I want to be able to use the SPI and `ServiceLoader` or at least obtained a pool instance from an untyped configuration.

Please let me know if there's something I can use.

vietj:
here is the PR actually @jkuhn1 https://github.com/eclipse-vertx/vertx-sql-client/pull/1101

jkuhn1:
That would definitely do. Thanks!

Issue #1040, opened 2021-09-13
java.lang.UnsupportedOperationException: Unable to decode metadata for DATETIMN


rasmushaglund:
Hello, I'm trying out the MSSQL client but encountered an error while trying to decode a nullable datetime field for MSSQL 2017 CU21.

https://github.com/eclipse-vertx/vertx-sql-client/blob/560041bb849dd6f80de9e50dfef8704e6b20c87b/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/DataType.java#L395

I debugged the code and saw that there's no metadata handler for the type. I don't fully understand the system here, but when I read an unsigned byte from the payload for the metadata decoding, ByteBufUtils.readUnsignedByteLengthString(payload) gave the correct columnName.

Can I help somehow? If I were about to add support for DATETIMN, what parts of the code would I look into?

vietj:
@tsegismont

tsegismont:
Hi @rasmushaglund , thanks for offering your help.

You can start by implementing the methods in the `DataType` class. Then you need to update the [data types test cases](https://github.com/eclipse-vertx/vertx-sql-client/tree/master/vertx-mssql-client/src/test/java/io/vertx/mssqlclient/data).

If you haven't contributed to Vert.x before, I recommend reading https://github.com/vert-x3/wiki/wiki/Development-Process and https://github.com/vert-x3/wiki/wiki/Vert.x-code-style-guidelines.

Looking forward to your contribution!

rasmushaglund:
Thanks @tsegismont. 

I assume this is the type documentation? https://docs.microsoft.com/en-us/sql/t-sql/data-types/datetime-transact-sql?view=sql-server-ver15

tsegismont:
Looks like it

>

rasmushaglund:
I think I've managed to implemented the type, but I don't know if and how to implement encodeParam. Is there any tests involving encodeParam which I can take a look at? Maybe just try to encode and decode back again and compare the results?

https://github.com/rasmushaglund/vertx-sql-client/commit/31278ffbf3fea1282eb9d2d77615ad0d0b7b2b9a

tsegismont:
There's the encodeParam implementation for TIMEN that you can inspire from:
[...]

PR #1041, opened 2021-09-16
MSSQL: Implement support for type datetime 

rasmushaglund:
Add support for type datetime (DATETIMN)

Related to https://github.com/eclipse-vertx/vertx-sql-client/issues/1040 and https://github.com/eclipse-vertx/vertx-sql-client/issues/608


tsegismont (approved):
Thank you @rasmushaglund , great job!

PR #1046, opened 2021-09-18
MS SQL: Add support for smalldatetime

rasmushaglund:
Add support for smalldatetime.

Apparently MS SQL returns values for columns with smalldatetime as datetime of length 4, so I had to modify the decoding of DATETIMN so it supports data with length 4 to get this to work.


rasmushaglund:
Sorry, I failed to merge in master

PR #1049, opened 2021-09-20
MSSQL: Add support for money and smallmoney datatypes

rasmushaglund:
Add support for money and smallmoney (MONEYN, MONEY, MONEY4).

Related #608

### DIFF
--- a/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/DataType.java
+++ b/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/DataType.java
@@ -20,8 +20,6 @@ import io.vertx.core.buffer.Buffer;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.math.RoundingMode;
-import java.nio.ByteBuffer;
-import java.nio.ByteOrder;
 import java.nio.charset.StandardCharsets;
 import java.sql.JDBCType;
 import java.time.*;


rasmushaglund:
I'm not entirely sure that money and smallmoney should be typed as BigDecimal, but the problem I ran into was that when I used double/float as type they were displayed in scientific notation when converted to string, but maybe that's ok?

rasmushaglund:
When looking at the mssql-jdbc-driver it also represents money and smallmoney as BigDecimal. https://github.com/microsoft/mssql-jdbc/blob/0f3e0a10f2f85b5b357811907fdb487d63e49830/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerParameterMetaData.java#L486-L488

rasmushaglund:
Looking at the mssql-jdbc-driver I see that they also represent money/smallmoney as BigDecimal. https://github.com/microsoft/mssql-jdbc/blob/89bb744675941113a0ca9bcd7dc3c8f1310bb23c/src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java#L129-L130

rasmushaglund:
Looking at mssql-jdbc-driver I see that they also use BigDecimal as type for money/smallmoney. https://github.com/microsoft/mssql-jdbc/blob/89bb744675941113a0ca9bcd7dc3c8f1310bb23c/src/main/java/com/microsoft/sqlserver/jdbc/DataTypes.java#L158-L161

tsegismont:
@rasmushaglund thanks, I will review the PR asap.

Yes, `BigDecimal` is the safe choice for representing money (you can find on the web tons of article about it).

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/DataType.java
```diff
@@ -152,7 +155,27 @@ public Object decodeValue(ByteBuf byteBuf, Metadata metadata) {
...
+    }
+
+    @Override
+    public Object decodeValue(ByteBuf byteBuf, Metadata metadata) {
+      ByteBuffer bb = ByteBuffer.allocate(8);
+      bb.order(ByteOrder.BIG_ENDIAN);
+      bb.putInt(byteBuf.readIntLE());
+      bb.putInt(byteBuf.readIntLE());
+
+      return new BigDecimal(new BigInteger(bb.array())).divide(new BigDecimal("10000"), 4, RoundingMode.UP);
```
tsegismont, line 171: This could be simplified to avoid the allocation:

```suggestion
      long highBits = (long) byteBuf.readIntLE() << 32;
      long lowBits = byteBuf.readIntLE() & 0xFFFFFFFFL;
      BigInteger bigInteger = BigInteger.valueOf(highBits | lowBits);
      return new BigDecimal(bigInteger).divide(new BigDecimal("10000"), 4, RoundingMode.UP);
```

tsegismont (changes requested):
LGTM, with a small change request.

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/DataType.java
rasmushaglund: Looks super!

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/DataType.java
```diff
@@ -19,6 +19,9 @@
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.math.RoundingMode;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
```
tsegismont, line 24:
```suggestion
```

PR #1051, opened 2021-09-24
Close stream in createStream example

julianladisch:
Each stream occupies memory in the database and needs to be
closed to avoid a database out-of-memory error.

### DIFF
--- a/vertx-pg-client/src/main/java/examples/SqlClientExamples.java
+++ b/vertx-pg-client/src/main/java/examples/SqlClientExamples.java
@@ -328,8 +328,12 @@ public class SqlClientExamples {
               System.out.println("Error: " + err.getMessage());
             });
             stream.endHandler(v -> {
-              tx.commit();
-              System.out.println("End of stream");
+              // Close the stream to release the resources in the database
+              stream.close(closed -> {
+                tx.commit(committed -> {
+                  System.out.println("End of stream");
+                });
+              });
             });
             stream.handler(row -> {
               System.out.println("User: " + row.getString("last_name"));


tsegismont (commented):
Thank you @julianladisch . I thought Postgres would destroy the cursor after transaction is complete.

Have you noticed an issue in practice>

julianladisch:
Yes.

How to reproduce:

```
  void usingCursors03(SqlConnection connection) {
    connection.prepare("SELECT $1 AS last_name", ar0 -> {
      if (ar0.succeeded()) {
        PreparedStatement pq = ar0.result();

        // Streams require to run within a transaction
        System.out.println("begin");
        connection.begin(ar1 -> {
          if (ar1.succeeded()) {
            Transaction tx = ar1.result();

            // Fetch 50 rows at a time
            RowStream<Row> stream = pq.createStream(50, Tuple.of("julien"));

            // Use the stream
            stream.exceptionHandler(err -> {
              System.out.println("Error: " + err.getMessage());
            });
            stream.endHandler(v -> {
              tx.commit(x -> usingCursors03(connection));
              System.out.println("End of stream");
            });
            stream.handler(row -> {
              System.out.println("User: " + row.getString("last_name"));
            });
          }
        });
      }
    });
  }

  @Test
  void usingCursors03(VertxTestContext vtc) {
    PgPool.pool().getConnection()
    .onComplete(vtc.succeeding(this::usingCursors03));
  }
```

Use `ps -F ax` or `pg_top` to watch how the database memory used by the connection increases by 0.5 MB per second.

It affects the FOLIO project in practice: https://issues.folio.org/browse/RMB-863

PR #1062, opened 2021-11-02
Fix ClassCastException when using `Tuple#getBuffer` with a `String`

### LINKED ISSUE #1061: Tuple#getBuffer doesn't work if value is a String
I can send a fix.
Maybe this is not really a bug but a new feature

DavideD:
Fixes #1061


DavideD:
This PR should be mergeble now if you want to apply ir.

PR #1075, opened 2021-11-24
Configure Oracle client from URI

### LINKED ISSUE #1071: Missing method OracleConnectOptions.fromURI()
All other `*ConnectionOptions` have this method.

see: [#797](https://github.com/eclipse-vertx/vertx-sql-client/issues/797)
and [PG version]: (https://github.com/eclipse-vertx/vertx-sql-client/blob/master/vertx-pg-client/src/main/java/io/vertx/pgclient/PgConnectOptions.java#L53)

Would it make sense to add a couple methods like this one to an interface?

tsegismont:
Closes #1071

This PR adds methods to:

- configure a pool from connection uri
- initialize connection options from uri

In order to avoid confusion, the `jdbc:` prefix is omitted.

### DIFF
--- a/vertx-oracle-client/src/main/asciidoc/index.adoc
+++ b/vertx-oracle-client/src/main/asciidoc/index.adoc
@@ -86,11 +86,20 @@ There are several alternatives for you to configure the client.
 
 === Data Object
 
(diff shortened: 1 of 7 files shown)


tsegismont:
cc @blafond

PR #1083, opened 2021-11-29
Implement Oracle generated keys retrieval

### LINKED ISSUE #1074: [Oracle]  Return a generated identity after an insert
In Oracle, when a column is defined as:
```
 create table EntityWithIdentity (id number(19,0) generated as identity, name varchar2(255 char), position number(10,0), primary key (id))
```

It's possible to get the value after an insert as:
```
insert into EntityWithIdentity (id, name, position) values (:1, :2 , :3 ) RETURNING id INTO :1
```

I don't think that client support this syntax at the moment.

tsegismont:
Fixes #1074


gavinking:
Is there some way we can get this API unified with the one for MySQL?

tsegismont:
@gavinking see https://github.com/eclipse-vertx/vertx-sql-client/pull/1083/commits/f4c323539a77bd1a95c80811381c5fd9f23421c1, the generated key values are stored as propery of the `SqlResult`. This is similar to the behavior of the MySQL client.

cescoffier (approved):

gavinking:
Right, I understand, but it's two different (very similar) APIs. I'm asking if they could be unified in some way.

tsegismont:
@gavinking MySQL can only return a big unsigned integer, Oracle can return a set of key values.
In Reactive SQL Clients we try to propose API which represent the corresponding database behavior.

vertx-oracle-client/src/test/java/io/vertx/oracleclient/test/OracleGeneratedKeysTest.java
```diff
@@ -0,0 +1,109 @@
...
+    doTest(ctx, () -> {
+      return null;
+    }, generated -> verifyGenerated(generated, "ROWID", byte[].class));
+  }
+
+  @Test
+  public void shouldRetrieveGeneratedKeyByName(TestContext ctx) {
+    doTest(ctx, () -> {
+      return new OraclePrepareOptions()
+        .setAutoGeneratedKeysIndexes(new JsonArray().add("id"));
```
DavideD, line 79: Wait, are you saying that I need to know in advance the name of the column that's going to be generated if I want to get the `id`?

DavideD: Well, no. In Postgres I can specify the return value as part of the query, for example `insert into EntityWithIdentity (name, position) values ($1, $2) returning id`. Here you are passing the return value using `OraclePrepareOptions`.
Maybe I'm missing something?

gavinking: I had understood that there was a way to do this with a named out parameter.

DavideD: I'm asking because our queries are currently like this:
```
insert into EntityWithIdentity (id, name, position) values (default, ?, ?)
```

But the result only contains the column with the `ROWID`

tsegismont: @DavideD right, the name of the column to return is not provided in the query itself, like with PostgreSQL, but in `PrepareOptions`. What I meant is that, in both cases, you need to know the name of that column.

tsegismont: @gavinking using `RETURNING INTO` only works when invoking a callable statement afaik

gavinking: Oh OK.

DavideD: > I had understood that there was a way to do this with a named out parameter.

@gavinking  Didn't we decide to not use output parameters because this was easier? I don't mind using output parameters if it's the way to go, but I didn't see an example in the tests in this class.

Having to use `OraclePrepareOptions` makes things more complicated for us and this is the only database that requires it (if I've understood correctly).

DavideD: > What I meant is that, in both cases, you need to know the name of that column.

In Postgres I can do `getValue(0)` and it will return the id even if I don't know the name of the column. Here's it seems that if I don't specify the name in the options the value doesn't appear at all

gavinking: I mean, I don't think we decided anything in particular.

I still advocate that the client should abstract this, just like the JDBC driver does. I really don't see what's so hard about that.

tsegismont: @DavideD this client is different from the others. In particular, it's based on the Oracle JDBC driver async extensions. In other words, we can't implement something that's not allowed by the Oracle JDBC implementation.

DavideD: Yes, an abstraction would be great.

tsegismont: >I still advocate that the client should abstract this, just like the JDBC driver does.

Can you show some code with Oracle JDBC API that allows to retrieve a generated key value without providing its name or index when preparing the statement?

If this is possible I'd be glad to change the implementation. But to my knowledge, it is not possible.

gavinking:
Look, currently the MySQL client forces us to have the following awful code in Hibernate Reactive:

```java
	static {
		try {
			mySqlDriverClass = Class.forName( "io.vertx.mysqlclient.MySQLClient" );
		}
		catch (ClassNotFoundException cnfe) {
			mySqlDriverClass = null;
		}
		if ( mySqlDriverClass != null ) {
			try {
				mySqlLastInsertedId = (PropertyKind<Long>)
						mySqlDriverClass.getField("LAST_INSERTED_ID").get(null);
			}
			catch (NoSuchFieldException | IllegalArgumentException | IllegalAccessException | SecurityException e) {
				throw new RuntimeException("Unable to obtain MySQLClient.LAST_INSERTED_ID field", e);
			}
		}
	}
```

It appears to me that the current solution contained here makes this issue even worse.

vertx-oracle-client/src/test/java/io/vertx/oracleclient/test/OracleGeneratedKeysTest.java
gavinking: > Can you show some code with Oracle JDBC API that allows to retrieve a generated key value without providing its name or index when preparing the statement?

I really don't understand this question. Why can't you just use an arbitrary name that *you* decide on? Why do we care what name it is?

gavinking: I mean, perhaps I've misunderstood your test code, but it doesn't look like there's anything in your SQL `insert` statement that contains the name you're passing with the `OraclePrepareOptions`. It looks like an arbitrary key. Have I misread the code?

tsegismont: @gavinking you can't just use an arbitrary name because Oracle needs to know which column you want, see https://github.com/eclipse-vertx/vertx-sql-client/blob/master/vertx-oracle-client/src/main/java/io/vertx/oracleclient/impl/commands/QueryCommand.java#L78-L120

Indeed that's not related to the query itself, but to how the JDBC API requires you to prepare the statement: https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/Connection.html#prepareStatement(java.lang.String,java.lang.String%5B%5D)

gavinking: > because Oracle needs to know which column you want

Um. OK. But I'm confused. Because that doesn't really seem to be the case in your tests.

You have written:

```java
    public void retrieveGeneratedKeyByIndex(SqlClient client) {
     String sql = "INSERT INTO EntityWithIdentity (name, position) VALUES (?, ?)";

     // Retrieve generated key column value by index
     OraclePrepareOptions options = new OraclePrepareOptions()
       .setAutoGeneratedKeysIndexes(new JsonArray().add("1"));

     client.preparedQuery(sql, options).execute(Tuple.of("john", 3), ar -> {
       if (ar.succeeded()) {
         RowSet<Row> result = ar.result();

         Row generated = result.property(OracleClient.GENERATED_KEYS);
         Long id = generated.getLong("ID");
       }
     });
   }
```

From the code above, it really doesn't look like Oracle knows ahead of time that the column name is `ID`. It looks like you could have written:


```java
    public void retrieveGeneratedKeyByIndex(SqlClient client) {
     String sql = "INSERT INTO EntityWithIdentity (name, position) VALUES (?, ?)";

     // Retrieve generated key column value by index
     OraclePrepareOptions options = new OraclePrepareOptions()
       .setAutoGeneratedKeysIndexes(new JsonArray().add("1"));

     client.preparedQuery(sql, options).execute(Tuple.of("john", 3), ar -> {
       if (ar.succeeded()) {
         RowSet<Row> result = ar.result();

         Row generated = result.property(OracleClient.GENERATED_KEYS);
         Long id = generated.getLong(1); // <---- look up by column index
       }
     });
   }
```

What happens if you do that?

DavideD: I've tested it and it seems to work.
One can specify:
```
.setAutoGeneratedKeysIndexes(new JsonArray().add(1))
```
and then read the value with:
```
Long id = generated.getLong(0);
```
because now RowId is not returned anymore, so the right index to get the value is `0`

DavideD: @tsegismont but that's not the only preparedStatement method, you also have [this one](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/Connection.html#prepareStatement(java.lang.String,int)):
```
PreparedStatement prepareStatement(String sql, int autoGeneratedKeys) throws SQLException

autoGeneratedKeys - a flag indicating whether auto-generated keys should be returned; one of Statement.RETURN_GENERATED_KEYS or Statement.NO_GENERATED_KEYS
```

tsegismont: @gavinking the snippet you posted is the test `retrieveGeneratedKeyByIndex`. So you tell Oracle which is the key index when preparing the statement and, after execution, the value and the column name are given by `statement.getGeneratedKeys()`.

@DavideD with Oracle, this prepareStatement method gives you back the `ROWID`, not the key value. There is a test case as well in this PR.

DavideD: Ok, I think I get it now.
But this option to specify the name of the ids, couldn't it be in a common place? It doesn't seem specific to oracle

DavideD: Couldn't it be in `PrepareOptions`?

gavinking: I mean, that `"1"` there:

- isn't a reference to anything in the SQL `insert` statement (if anything in the SQL has index 1, it's `name`, not `id`), and
- it sure isn't the name of a column of the table.

So I don't see what it can possibly be other than just an arbitrary index.

gavinking: I mean look, it seems to me that the assertion that this can't be implemented just has to be wrong, since the Oracle JDBC driver implements it. Surely there's some way to make it work. 

The API you're proposing isn't user-friendly at all, and in particular it makes it very difficult to write code that works across different platforms. So it really is worth spending some time digging into this.

gavinking: > I've tested it and it seems to work.

@DavideD I mean, if you say it works, it looks just perfect to me. I really think the case of retrieving exactly one generated key is much more common than retrieving multiple generated keys, and so I think it makes sense that the Vert.x drivers provide an optimizes API for that usage.

tsegismont: @gavinking can you show me in Hibernate (classic, not reactive) how is implemented generated key retrieval? Perhaps we've misunderstood how the JDBC API works.

gavinking: I mean, in Hibernate classic it's all wrapped up in a couple of layers of abstraction, so that's not a good example. But a quick google search turned up this page, which shows three allowed ways to do it:

<https://docs.oracle.com/javadb/10.6.2.1/ref/crefjavstateautogen.html>

That's with a `Statement` or course, whereas Hibernate uses a `PreparedStatement`.

FTR Hibernate uses the first approach, with `PreparedStatement.RETURN_GENERATED_KEYS`.

gavinking: > FTR Hibernate uses the first approach, with `PreparedStatement.RETURN_GENERATED_KEYS`.

Well, actually let me walk that back. I'm now unsure what Hibernate does because it's buried under too many layers of abstraction and I don't currently have a way to debug it on Ora.

That documentation page certainly does seem to indicate that it's possible, however, and, furthermore, I still don't see why the approach of passing an index can't work in general. Perhaps there's something I'm missing.

DavideD: Hibernate ORM passes the id column names.

it's not much work to get the id column names and I already have a proof of concept. The issue we have is that passing them in vertx requires the `OraclePrepareOptions` class that i's only in the oracle client driver. Making it complicated to 
reuse the same code when using different clients. This property doesn't seem specific to Oracle though.

DavideD: This is the line in Hibernate ORM: https://github.com/hibernate/hibernate-orm/blob/67e09b12908de27686169ade8d590b5b7ea1f9ef/hibernate-core/src/main/java/org/hibernate/engine/jdbc/internal/StatementPreparerImpl.java#L120

gavinking: Right, to be clear: my main problem with this is not that we might have to pass the column name. It's that the current API forces us to write some super-nasty code using Java reflection.

tsegismont: >But a quick google search turned up this page, which shows three allowed ways to do it:
>https://docs.oracle.com/javadb/10.6.2.1/ref/crefjavstateautogen.html

@gavinking this is a page for JavaDB a.k.a Derby

> The issue we have is that passing them in vertx requires the OraclePrepareOptions class that's only in the oracle client driver.

@DavideD got you. I'll share a PR today for a solution that does not require you to depend on Oracle Client specific code.

gavinking: > this is a page for JavaDB a.k.a Derby

Ah haha. OK. My bad. 

(Though I _still_ don't see why the index-based `.setAutoGeneratedKeysIndexes()` approach doesn't work.)

Anyway, as long as we can get a single unified API that works with both MySQL and Oracle, then we can figure it out. That's the main issue here.

DavideD: @tsegismont Would it be possible to move `setAutoGeneratedKeysIndexes` in `PrepareOptions`?

DavideD: Thanks @tsegismont

PR #1099, opened 2021-12-08
Create PreparedQuery with options in JSON format

tsegismont:
This new method allow to create a `PreparedQuery` with Oracle specific options without creating `OraclePreparedOptions`.

cc @DavideD


tsegismont:
@DavideD so with this change you could do:

```java
    PropertyKind<Row> generatedKeys = PropertyKind.create("generated-keys", Row.class);

    String sql = "INSERT INTO EntityWithIdentity (name, position) VALUES (?, ?)";

    // Retrieve generated key column value by name
    JsonObject options = new JsonObject()
      .put("autoGeneratedKeysIndexes", new JsonArray().add("ID"));

    client.preparedQuery(sql, options).execute(Tuple.of("john", 3), ar -> {
      if (ar.succeeded()) {
        RowSet<Row> result = ar.result();

        Row generated = result.property(generatedKeys);
        Long id = generated.getLong("ID");
      }
    });
```

And there would be no type specific to the Oracle Client.

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlClient.java
```diff
@@ -54,6 +54,13 @@
    */
   PreparedQuery<RowSet<Row>> preparedQuery(String sql, PrepareOptions options);
 
+  /**
+   * Like {@link #preparedQuery(String, PrepareOptions)} with options in JSON format.
+   */
+  default PreparedQuery<RowSet<Row>> preparedQuery(String sql, JsonObject options) {
```
vietj, line 60: I prefer to avoid this, if users need they can create a PrepareOptions from a JsonObject

tsegismont: If users create generic `PrepareOptions` from `JsonObject`, the Oracle specific options will be lost, won't they?

tsegismont: What we could do instead in Oracle Client is what the various `MetricFactory` do with `MetricsOptions`, for example: https://github.com/vert-x3/vertx-micrometer-metrics/blob/103c5f024d3f36e40a704254e3eb8d28dfec0bd5/src/main/java/io/vertx/micrometer/impl/VertxMetricsFactoryImpl.java#L40-L45

vietj: yes that would work nicely I think

tsegismont: >also sometimes what @pmlopes did is that the data object extends the JsonObject directly

@vietj @pmlopes any specific example in mind?

tsegismont:
@vietj I've updated the PR according to your feedback

@DavideD here's how it would work:

```java
    PropertyKind<Row> generatedKeys = PropertyKind.create("generated-keys", Row.class);

    String sql = "INSERT INTO EntityWithIdentity (name, position) VALUES (?, ?)";

    // Retrieve generated key column value by name
    JsonObject options = new JsonObject()
      .put("autoGeneratedKeysIndexes", new JsonArray().add("ID"));

    client.preparedQuery(sql, new PrepareOptions(options)).execute(Tuple.of("john", 3), ar -> {
      if (ar.succeeded()) {
        RowSet<Row> result = ar.result();

        Row generated = result.property(generatedKeys);
        Long id = generated.getLong("ID");
      }
    });
```

Issue #965, opened 2021-05-26
Consider minimizing ByteBuf write operations


tsegismont:
we should consider minimizing ByteBuf write operations, each write has a cost with respect to bound checks and also often ByteBuf have nested buffers

_Originally posted by @vietj in https://github.com/eclipse-vertx/vertx-sql-client/pull/964#r639733371_

PR #1109, opened 2021-12-18
Support for UUID/uniqueidentifier data types in MSSQL.

Yaytay:
Motivation:

Support for UUID/uniqueidentifier data types in MSSQL.
Our databases make a lot of use of uniqueidentifier columns - without support for them we cannot use vertx-sql-client.

### DIFF
--- a/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/DataType.java
+++ b/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/DataType.java
@@ -26,6 +26,7 @@ import java.time.*;
 import java.time.temporal.ChronoUnit;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.UUID;
 
 import static io.vertx.mssqlclient.impl.utils.ByteBufUtils.*;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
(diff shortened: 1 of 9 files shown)


Yaytay:
Just noticed this: https://github.com/eclipse-vertx/vertx-sql-client/issues/965
I can reduce from four ByteBuf:write* calls to two writeLong calls or even one writeBytes call if I do the byte-twiddling myself.
Let me know if you'd like me to make that change.

vietj:
thanks for the contribution @Yaytay

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/DataType.java
```diff
@@ -240,7 +241,56 @@ public Object decodeValue(ByteBuf byteBuf, Metadata metadata) {
...
+    @Override
+    public Metadata decodeMetadata(ByteBuf byteBuf) {
+      Metadata metadata = new Metadata();
+      metadata.length = byteBuf.readByte();
+      return metadata;
+    }
+
+    @Override
+    public JDBCType jdbcType(Metadata metadata) {
+      if (metadata.length == 16) return JDBCType.CHAR;
```
tsegismont, line 254:
```suggestion
      if (metadata.length == 16) return JDBCType.GUID;
```

tsegismont (changes requested):
Thank you @Yaytay , great job!

Just a couple changes needed. Do you think you can do it before tomorrow morning CEST ? We'll release a new 4.2 version and, in this case, the change will be part of it.

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/DataType.java
Yaytay: There is no JDBCType.GUID in https://docs.oracle.com/javase/8/docs/api/java/sql/JDBCType.html.
Any other idea?
JDBCType.OTHER?
JDBCType.OBJECT?
JDBCType.BINARY?

tsegismont (approved):
Thanks @Yaytay 

I'll go ahead and merge so that the PR is included in 4.2.3

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/DataType.java
tsegismont: Changed it to `OTHER`

Issue #1111, opened 2021-12-20
Support XML Type on Postgres


mauravan:
#### Describe the feature

Postgres supports the storage of valid xml documents with the XMLType. https://www.postgresql.org/docs/9.3/datatype-xml.html
I think it would make sense to Map it to a Java String

#### Use cases

From the link above: 

> The xml data type can be used to store XML data. Its advantage over storing XML data in a text field is that it checks the input values for well-formedness, and there are support functions to perform type-safe operations on it

#### Contribution

I would like to implement this but would probably need a few pointers. 

I already looked at these classes (DataType.java, DataTypeCodec.java, PreparedStatementTestBase.java, RowImpl.java) and added a class to wrap the xml data which i think might be a bit redundant in the end as we could just use the String class i think. Anyway a quick talk about how a satisfactory solution could look like and what you guys think would be needed to get this going would be appreciated.

Kind regards

vietj:
please provide a PR, I think a wrapper will be necessary for XML

mauravan:
Hi @vietj, thanks for your response. Here is the PR: https://github.com/eclipse-vertx/vertx-sql-client/pull/1112

PR #1112, opened 2021-12-21
add  xmltype in postgres

mauravan:
Motivation:

see #1111 

I'm looking for a little guidance on this. 

1. where can i find an integration test i could use to test against while implementing?
2. maybe some pointers on where to start

thanks
R


vietj:
@mauravan you can look at the existing tests of other PG specific data type and add tests in their location

vietj:
I'm curious, did you try using a string ?

mauravan:
that's why im looking for an integration test - i suspect we _might_ be able to use a String as long as we pass it as UTF8. Then we could just remove the wrapper class. 

i dod not have a lot of time to look into it, but i will have some time soon and will dig a bit deeper

mauravan:
Hi @vietj,

So turns out we might as well use String. Or do you think there is value in an extra class like the PgSQLXML i added? otherwise i will happily remove it.

Also what needs to happen next to get this merged?

Thanks for your help,
R

vietj:
this needs to be reviewed to be merged

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
```diff
@@ -229,5 +220,8 @@ static DataType lookup(Class<?> type) {
     encodingTypeToDataType.put(Polygon[].class, POLYGON_ARRAY);
     encodingTypeToDataType.put(Circle.class, CIRCLE);
     encodingTypeToDataType.put(Circle[].class, CIRCLE_ARRAY);
+
+//    encodingTypeToDataType.put(String.class, XML);
```
vietj, line 224: why is this commented ?

vietj:
this PR changes a lot of things that should not be changed (like reformatting), can you keep only to the essential changes ?

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
mauravan: See the Comment from the 22. of Dec:

` So i did it using String now and it works just fine i think - one problem i encountered was the mapping in the DataType.java because there is already a mapping for String to Varchar

encodingTypeToDataType.put(String.class, VARCHAR); encodingTypeToDataType.put(String[].class, VARCHAR_ARRAY); encodingTypeToDataType.put(String.class, XML); encodingTypeToDataType.put(String[].class, XML_ARRAY);

Which seems to be only important in case of a PrepareStatementCommand with parameterTypes which the test i'm executing are apparently not using. Maybe you have some more insight into that?`

mauravan:
> this PR changes a lot of things that should not be changed (like reformatting), can you keep only to the essential changes ?

Hi @vietj, thanks for the feedback - i reverted all the changes.

The one big Problem with the `private static final IntObjectMap<DataType> oidToDataType = new IntObjectHashMap<>();` remains. this structure does not allow for a duplicate mapping of String. 

One solution i could think of is to use a wrapper class for XML which would then just hold the String. Do you have any other ideas?

vietj:
why is it an issue ?

mauravan:
There is these 2 cases for the Map: 

`  static DataType valueOf(int oid) {
    DataType value = oidToDataType.get(oid);
    if (value == null) {
      logger.debug("Postgres type OID=" + oid + " not handled - using unknown type instead");
      return UNKNOWN;
    } else {
      return value;
    }
  }

  static DataType lookup(Class<?> type) {
    DataType dataType = encodingTypeToDataType.get(type);
    if (dataType == null) {
      if (Buffer.class.isAssignableFrom(type)) {
        return BYTEA;
      }
      dataType = DataType.UNKNOWN;
    }
    return dataType;
  }`

Where for the valueOf Method we have a mapping for each oid for the lookup i think there can only be a mapping per type. So either the Mapping is

`    encodingTypeToDataType.put(String.class, XML);
    encodingTypeToDataType.put(String[].class, XML_ARRAY); `

Or it is
`    encodingTypeToDataType.put(String.class, VARCHAR);
    encodingTypeToDataType.put(String[].class, VARCHAR_ARRAY);
`
The lookup seems to be called during encoding and the valueOf during decoding. So it seems to me, that encoding of String -> varchar would be disabled?

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
```diff
@@ -186,8 +186,11 @@ static DataType lookup(Class<?> type) {
     for (DataType dataType : values()) {
       oidToDataType.put(dataType.id, dataType);
     }
-    encodingTypeToDataType.put(String.class, VARCHAR);
-    encodingTypeToDataType.put(String[].class, VARCHAR_ARRAY);
+//    encodingTypeToDataType.put(String.class, VARCHAR);
```
vietj, line 189: why is this commented ?

mauravan: As far as I understand it, there can only be one mapping per class to a Type. This is the Problem i was talking about in the other comments which we need to solve here - so either there will be a need for a wrapper class or someone can come up with a better solution for this

vietj: you are totally right, this is because we cannot specify which database type to use in a tuple for a given java type

mauravan: So there will be something like a new Class like `PgXMLType` which just holds a String. Would you agree to that?

vietj:
I need to think more about it and discuss the ability to  specify
types on a tuple

PR #1128, opened 2022-02-01
Oracle pool

### LINKED ISSUE #1127: Connection pooling for Oracle client
As of 4.2.4, the Reactive Oracle client API has an `OraclePool` but it is implemented by creating/closing a JDBC connection without caching them.

tsegismont:
Fixes #1127


tsegismont:
@Sanne would you mind having a look at this PR? In particular, could you confirm that nothing else is needed beyond the content of the last commit in order to comply with the JDBC API requirements regarding connection pooling?

vertx-oracle-client/src/main/java/io/vertx/oracleclient/OracleConnection.java
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * Create a connection to Oracle with the given {@code connectOptions}.
+   *
+   * @param vertx the vertx instance
+   * @param connectOptions the options for the connection
+   * @param handler the handler called with the connection or the failure
+   */
+  static void connect(Vertx vertx, OracleConnectOptions connectOptions, Handler<AsyncResult<OracleConnection>> handler) {
+    Future<OracleConnection> fut = OracleConnectionImpl.connect(vertx, connectOptions);
+    if (handler != null) {
```
Sanne, line 41: so what happens when the handler is null? My guess is a connection is eventually opened, and it's leaked?

In that case, wouldn't it be better for this code to throw an exception before attempting to connect?

tsegismont:
>Before I do, could you clarify why you mention JDBC API ? Why does it matter in this context?

The Reactive Oracle Client is different than the other Vert.x SQL clients because it is built atop the Oracle JDBC driver.
And, since JDBC 4.3, connections pools shall invoke `beginRequest`/`endRequest` (see https://dba.stackexchange.com/a/208215 and https://groups.google.com/g/r2dbc/c/zZ09hFMR9N8/m/aTKoFU8pBwAJ)

vietj:
I would like the before/after oracle specific connection event to be moved from the `Connection` interface to a subclass of `SqlConnectionPool` for oracle that would handle this with a hook so all happens in a single place and does not affect other implementation that do not need this.

vietj (changes requested):
I would like the before/after oracle specific connection event to be moved from the `Connection` interface to a subclass of `SqlConnectionPool` for oracle that would handle this with a hook so all happens in a single place and does not affect other implementation that do not need this.

vietj:
ping @tsegismont

tsegismont:
@vietj please take a look at the last commit.

It adds a flag to the connection object indicating whether pool hooks should be used.
In order to avoid too much complexity, the flag is only used in the fast track where the potential performance penalty can be important.

vietj:
after reconsidering, I think it is more clean to add 2 functions (`Function<Connection, Future<Void>`) for the lifecycle in the SqlConnectionPool class for this matter instead of adding this to the `Connection` interface

tsegismont:
@vietj PTAL

Sanne:
awesome :100:

PR #1138, opened 2022-02-11
Wrong JDBC type for timestampz

### LINKED ISSUE #1124: [ORACLE] Wrong JDBC type returned for `current_timestamp`
Reproducer for `OracleColumnDescriptorTest`:
```
  @Test
  public void testMetadata(TestContext ctx) {
    pool.withConnection(conn -> conn.query("SELECT id, val, current_timestamp FROM mutable").execute(), ctx.asyncAssertSuccess(rows -> {
      ...
      // It will return JDBCType.OTHER instead of the expected one
      ctx.assertEquals( JDBCType.TIMESTAMP_WITH_TIMEZONE, rows.columnDescriptors().get(2).jdbcType());
    }));
  }
```

Note that the `typeName()` is actually correct.

I've updated the existing testcase here: https://github.com/DavideD/vertx-sql-client/commit/8ddd42bb87c4befb3a9867de934f392259472df7

tsegismont:
Fixes #1124

Oracle returns a int type which does not match `JDBCType.TIMESTAMP_WITH_TIMEZONE` when the column's DB type is `TIMESTAMPZ`.

### DIFF
--- a/vertx-oracle-client/src/main/java/io/vertx/oracleclient/impl/OracleColumnDesc.java
+++ b/vertx-oracle-client/src/main/java/io/vertx/oracleclient/impl/OracleColumnDesc.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2021 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2022 Contributors to the Eclipse Foundation
  *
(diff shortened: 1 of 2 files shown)


tsegismont:
@DavideD what do you think about this?

DavideD:
I think it will work, thanks

PR #1170, opened 2022-03-31
Oracle Transaction demarcation broken when using OraclePool

### LINKED ISSUE #1169: Broken transaction demarcation with OraclePool
When queries are executed between two transactions, the second transaction fail to start with:

```
io.vertx.oracleclient.OracleException: Error : 1453, Position : 0, Sql = SET TRANSACTION ISOLATION LEVEL READ COMMITTED, OriginalSql = SET TRANSACTION ISOLATION LEVEL READ COMMITTED, Error Msg = ORA-01453: SET TRANSACTION must be first statement of transaction
```

Here is a reproducer:

```java
  @Test
  public void testTransactionsInConsecutiveConnectionAcquisitions(TestContext ctx) {
    Pool pool = getPool();
    pool.withTransaction(client -> client.query("INSERT INTO mutable (id,val) VALUES (1,'bim')").execute().<Void>mapEmpty())
      .compose(v -> pool.withConnection(client -> client.query("DELETE FROM mutable WHERE id = 1").execute().<Void>mapEmpty()))
      .compose(v -> pool.withTransaction(client -> client.query("SELECT 1 FROM DUAL").execute().<Void>mapEmpty()))
      .onComplete(ctx.asyncAssertSuccess());
  }
```

The problem is that the client does not restore the auto-commit mode after a transaction is committed or rolled-back.
So when we think we're starting another transaction, in fact another one is already active and statements have been executed inside.

This problem was not visible before 4.2.5 because the Oracle client pool was a fake one: connections were "physically" closed when returned to the pool.

tsegismont:
Fixes #1169

The autocommit mode must be re-activated after the transaction completes or fails.

### DIFF
--- a/vertx-oracle-client/src/main/java/io/vertx/oracleclient/impl/Helper.java
+++ b/vertx-oracle-client/src/main/java/io/vertx/oracleclient/impl/Helper.java
@@ -11,6 +11,7 @@
 package io.vertx.oracleclient.impl;
 
 import io.vertx.core.Future;
+import io.vertx.core.Handler;
 import io.vertx.core.Promise;
 import io.vertx.core.VertxException;
 import io.vertx.core.impl.ContextInternal;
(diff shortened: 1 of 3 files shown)


vertx-oracle-client/src/main/java/io/vertx/oracleclient/impl/Helper.java
```diff
@@ -332,4 +344,34 @@ default void run() throws VertxException {
      */
     JdbcRow copy();
   }
+
+  @FunctionalInterface
+  public interface SQLBlockingCodeHandler<T> extends Handler<Promise<T>> {
```
vietj, line 349: as this is blocker I think it would be better to have a synchronous return from doHandler, i.e 

```
T doHandle();
```

and let the `handle` implementation complete the promise.

tsegismont: Good point. I've updated the PR.

vietj (approved):

tsegismont:
@DavideD would you mind trying this patch and confirm it fixes the issue?

DavideD:
Seems to work fine. Thanks a lot

vietj:
thanks for the confirmation @DavideD

PR #1175, opened 2022-04-08
MSSQL Cursor and Row streaming implementation

### LINKED ISSUE #606: [MSSQL]Cursor support
add support for using cursor/row streaming with sql server client

tsegismont:
Closes #606 

We use static cursors so that we get information about the total number of rows to fetch.

Cursor data is stored in a map at the connection level.
This allows to manage state using the cursor ids generated by the client as keys.

### DIFF
--- a/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/CursorData.java
+++ b/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/CursorData.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2011-2022 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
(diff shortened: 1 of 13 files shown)


tsegismont:
@vietj any comment about this one?

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/CursorData.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.mssqlclient.impl.codec;
+
+/**
+ * Data related to a server cursor.
+ */
+public class CursorData {
```
vietj, line 17: consider making it package private if the class is not shared outside

vietj (approved):
minor comment about CursorData visibility

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/CursorData.java
tsegismont: Done

PR #1190, opened 2022-05-23
Fix ClassCastException when get null enum value

longdt:
I found something typo mistake in document of method io.vertx.sqlclient.Tuple.get(Class<T> type, int position):

  /**
   * Get the the at the specified {@code position} and the specified {@code type}.
   *
   * <p>The type can be one of the types returned by the row (e.g {@code String.class}) or an array
   * of the type (e.g {@code String[].class})).
   *
   * @param type the expected value type
   * @param position the value position
   * @return the value if the value is not found or null.
   */
  default <T> T get(Class<T> type, int position) {
 
The return statement: @return the value if the value is not found or null. should be: @return the value if the value is found or null. 
And according to document of method  io.vertx.sqlclient.Tuple.get(Class<T> type, int position) the implementation: io.vertx.pgclient.impl.RowImpl.get(Class<T> type, int position) should return null But I get ClassCastException when try get value of enum (db store value = null)
main reason here is RowImpl.getEnum(Class enumType, int pos) doesn't allow null value

### DIFF
--- a/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/RowImpl.java
+++ b/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/RowImpl.java
@@ -227,8 +227,10 @@ public class RowImpl extends ArrayTuple implements Row {
           return constants[ordinal];
         }
       }
+    } else if (val == null) {
+      return null;
     }
-    return null;
+    throw new ClassCastException();
   }
 
   /**
--- a/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/RowImpl.java
+++ b/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/RowImpl.java
@@ -228,7 +228,7 @@ public class RowImpl extends ArrayTuple implements Row {
         }
       }
     }
-    throw new ClassCastException();
+    return null;
   }
 
   /**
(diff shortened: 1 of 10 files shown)


vertx-pg-client/src/main/java/io/vertx/pgclient/impl/RowImpl.java
```diff
@@ -228,7 +228,7 @@ private Object getEnum(Class enumType, int pos) {
         }
       }
     }
-    throw new ClassCastException();
+    return null;
```
vietj, line 231: this CCE is here for a reason, e.g the class type is incorrect, so it should return null only if vall is null otheriwse it must throw CCE. It seems this case is not tested and it should be tested as well

longdt: At first glance, I think you want to throw exception if incorrect class type, but I checked other Row implementation and see all of them return null. So not sure  what do you want to. 
It's ok, let me add check null before return null

longdt: Done. Anyway, do you want to add this check for other Row implementations? @vietj

vietj: yes the impl should be consistent in similar methods I think

longdt: yes, I updated code for other implementations before

PR #1192, opened 2022-05-24
Implement Oracle binary types encoding and decoding

### LINKED ISSUE #1085: Filtering on column of type raw throws an exception in Oracle
This is the exception:
```
ORA-00932: inconsistent datatypes: expected - got BLOB
```
It happens with the following query (for example) passing a parameter of type `Buffer`:
```
select * from UUIDEntity where id in (?,?)
```
where the table `UUIDEntity` is created with:
```
create table UUIDEntity (id raw(255) not null, name varchar2(255 char), primary key (id))
```

Note that the error happens even if the field is not a key of the table. Everything works fine if it's an insert query.

I've created a testcase here:  https://github.com/DavideD/vertx-sql-client/commit/b72733af9507a5f197cd1bfed6b11e95afd6310e

tsegismont:
Fixes #1085

### DIFF
--- a/vertx-oracle-client/src/main/asciidoc/index.adoc
+++ b/vertx-oracle-client/src/main/asciidoc/index.adoc
@@ -164,8 +164,17 @@ Currently, the client supports the following Oracle data types:
 * FLOAT(`java.lang.Double`)
 * DATE(`java.time.LocalDate`)
 * TIMESTAMP(`java.time.LocalDateTime`)
+* RAW(`io.vertx.core.buffer.Buffer`)
 
 Tuple decoding uses the above types when storing values.
+
+`BLOB` data type is also supported with one caveat: it must be represented by an instance of {@link io.vertx.oracleclient.data.Blob} when writing or filtering.
+However, when reading `BLOB` data, the client returns a {@link io.vertx.core.buffer.Buffer}.
+
+[source,$lang]
+----
+{@link examples.OracleClientExamples#blobUsage}
+----
 
 == Tracing queries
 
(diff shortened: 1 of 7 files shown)


tsegismont:
@DavideD would you mind reviewing this one please?

DavideD:
Testing...

DavideD:
That specific issue is fixed, but something else fails when I switch from `4.3.0` to `4.3.1-SNAPSHOT`. 
I'm trying to figure out what's going on

DavideD:
This solution doesn't really work for us.

The problem is that a value of type `blob` in the db is returned as `io.vertx.oracleclient.data.Blob` (in `4.3.0` the value is returned as `byte[]`).
That class only exists in the `vertx-oracle-client` (we might not have it on the classpath.
Can't you return the value as `byte[]` like in `4.3.0`?

DavideD:
I don't think so.

We don't know in advance what value will be returned by a query, just that it's something that can be converted to a `java.sql.Blob`. This means that other dbs might return a `byte[]` or  a `String` and if we call `getBuffer`on one of those fields we have an exception (I think).

I would need to run more tests to be more precise.
But this is our implementation at the moment:
```
       private io.vertx.sqlclient.Row row;

	@Override
	public Blob getBlob(String columnLabel) {
		final Object value = row.getValue( columnLabel );
		wasNull = value == null;
		if ( wasNull ) {
			return null;
		}

		if ( value instanceof Buffer ) {
			return BlobProxy.generateProxy( ( (Buffer) value ).getBytes() );
		}
		if ( value instanceof String ) {
			return BlobProxy.generateProxy( ( (String) value ).getBytes() );
		}
		if ( value instanceof byte[] ) {
			return BlobProxy.generateProxy( (byte[]) value );
		}

		throw new IllegalArgumentException( "Unexpected type: " + value.getClass() + " for column " + columnLabel );
	}
```

This is the same code for all dbs.

DavideD:
Atually, it might work, let me try

tsegismont:
@DavideD I changed the implementation, `io.vertx.oracleclient.data.Blob` will be required only when inserting/filtering.

I verified that HR test suite passes with these changes.

DavideD:
OK, thanks

PR #1224, opened 2022-08-08
Log MSSQL INFO token content or let users handle it

### LINKED ISSUE #1223: Log MSSQL INFO token content or let users handle it
When executing commands, the server response may contain an INFO token which holds message/severity data.

Instead of simply ignoring it, we should log the data or let users decide how to handle it.

tsegismont:
Closes #1223


vietj (commented):
LGTM

Issue #1214, opened 2022-07-14
Propagate transactions in context


technical-debt-collector:
#### Read me

Read this first before creating an issue:

- do not use this issue tracker to ask questions, instead use one of these [channels](https://vertx.io/community/). Questions will likely be closed without notice.
- you shall create a feature request only when it is general purpose enough.
- make sure that the feature is not already

#### Describe the feature

While the simplified transaction API works great with singular operations or statements that can be grouped together, there doesn't seem to be any practical way of encapsulating multiple separately implemented DB statements within a transaction, such as in the case of controller layers.

See short discussion in https://github.com/quarkusio/quarkus/issues/26705 where Gavin suggested that propagating the transaction via the Vert.x context is the way to address this.

#### Use cases

- Transaction handling that is easier to manage

#### Contribution
@vietj @tsegismont 

I'd be happy to try looking further into this myself if you think this could be worth exploring, but I am fairly unexperienced with the project so it'd probably require some time and assistance.

vietj:
can you describe more precisely what you would expect ? i.e give a code example of how you think it would / should work. It would help a lot.

technical-debt-collector:
I am currently using Mutiny and not Vert.x directly so I might be a little out of touch here. But could something like the following work?

1. Create a new method on interface Pool that:
- Looks for an active connection via the local context and returns it if found
- Otherwise create another, place that in the local context and return it

2. Add the transaction accessor to Connection as requested in #646 

Then third party code could easily retrieve both the active connection and the associated transaction without having to hold onto either of them.

Alternatively, instead of creating a new method, the existing getConnection-methods functionality could just place the connection they acquire in local context while also returning it, but I'm not sure if that could have unfortunate side effects?

gavinking:
@vietj you could look at how HR does it for inspiration: essentially, we store the connection in the vert.x local context from the outermost call to `withTransaction()`, and pull it from there in nested calls.

vietj:
@technical-debt-collector so something like `getOrCreateContext()` but with a connection

technical-debt-collector:
@vietj Yes, essentially.

gavinking:
> @technical-debt-collector so something like `getOrCreateContext()` but with a connection

But I would make it accept a function reference.

technical-debt-collector:
@gavinking, do you mean like with Pool.withTransaction() where the transaction is committed/rolled back and the connection is released after the function execution?
[...]

PR #1231, opened 2022-09-08
Create propagatable connection, add Transaction accessor

technical-debt-collector:
Motivation:

Proposed solution to https://github.com/eclipse-vertx/vertx-sql-client/issues/1214 and https://github.com/eclipse-vertx/vertx-sql-client/issues/646

This could probably be improved a great deal, but I've reached a stage with this one where I'd need some more eyes on it.
I'm also struggling with running a lot of the tests, so I am somewhat in the dark on whether or not this has broken anything.

Conformance:

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Pool.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Pool.java
@@ -168,7 +168,7 @@ public interface Pool extends SqlClient {
    * Like {@link #withTransaction(Function, Handler)} but allows for setting the mode, defining how the acquired
    * connection is managed during the execution of the function.
    */
-  <T> Future<@Nullable T> withTransaction(TransactionMode mode, Function<SqlConnection, Future<@Nullable T>> function);
+  <T> Future<@Nullable T> withTransaction(TransactionPropagation txPropagation, Function<SqlConnection, Future<@Nullable T>> function);
 
   /**
    * Get a connection from the pool and execute the given {@code function}.
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Pool.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Pool.java
@@ -140,7 +140,7 @@ public interface Pool extends SqlClient {
   }
 
   /**
-   * Like {@link #withTransaction(Function, Handler)} but returns a {@code Future} of the asynchronous result
+   * Like {@link #withTransaction(Function, Handler)} but returns a {@code Future} of the asynchronous result.
    */
   default <T> Future<@Nullable T> withTransaction(Function<SqlConnection, Future<@Nullable T>> function) {
     return getConnection()
@@ -165,10 +165,10 @@ public interface Pool extends SqlClient {
   }
 
   /**
-   * Like {@link #withTransaction(Function, Handler)} but keeps the connection accessible via the context
-   * for the duration of the given {@code function}
+   * Like {@link #withTransaction(Function, Handler)} but allows for setting the mode, defining how the acquired
+   * connection is managed during the execution of the function.
    */
-  <T> Future<@Nullable T> withPropagatedTransaction(Function<SqlConnection, Future<@Nullable T>> function);
+  <T> Future<@Nullable T> withTransaction(TransactionMode mode, Function<SqlConnection, Future<@Nullable T>> function);
 
   /**
    * Get a connection from the pool and execute the given {@code function}.
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Pool.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Pool.java
@@ -98,12 +98,6 @@ public interface Pool extends SqlClient {
    */
   Future<SqlConnection> getConnection();
 
-  Future<SqlConnection> getPropagatableConnection();
-
-  boolean propagatableConnectionIsActive();
-
-  Future<Void> setPropagatableConnection(SqlConnection propagatableConnection);
-
   /**
    * {@inheritDoc}
    *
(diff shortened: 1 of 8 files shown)


vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/PoolImpl.java
```diff
@@ -155,6 +157,24 @@ public Future<SqlConnection> getConnection() {
     });
   }
 
+  @Override
+  public Future<SqlConnection> getPropagatableConnection() {
+    if (propagatableConnection == null) {
```
vietj, line 162: the idea is that such transaction is stored in the Vert.x current context and not a field on the pool, allowing multiple concurrent transactions.

technical-debt-collector: I somehow managed to forget that completely. 
There weren't really that many examples showing how items should be managed in the context, so please let me know if this is not the way you prefer doing this.

I've also cleaned up the code a bit, avoiding unneccesary additional methods here and there.

vietj: it should use Vert.x `Context` with getLocal/putLocal methods

technical-debt-collector: Changed it now. Is it safe to assume there always being a current context available?

vietj: yes, it is lazy created when it does not exist

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/PoolImpl.java
```diff
@@ -155,6 +158,44 @@ public Future<SqlConnection> getConnection() {
     });
   }
 
+  public <T> Future<@Nullable T> withPropagatedTransaction(Function<SqlConnection, Future<@Nullable T>> function) {
+    Context context = Vertx.currentContext();
+    SqlConnection sqlConnection = context.getLocal(PROPAGATABLE_CONNECTION);
+    if (sqlConnection == null) {
+      return initializePropagatedConnectionAndTransaction(function);
+    }
+    return Future.succeededFuture(sqlConnection)
```
vietj, line 167: instead here we need to use `context.succeededFuture` (context must be casted to `ContextInternal`)

technical-debt-collector: Done. However, the authorship got messed up as I did this from another machine (both authorships should refer to the same person, me). Do you need me to sign the ECA once again?

technical-debt-collector: I tried to register an Eclipse user for my work email and signing the ECA there as well. Something changed on the verification page and it says that it's signed, but it still shows as not valid

vertx-sql-client/src/main/java/io/vertx/sqlclient/Pool.java
```diff
@@ -164,6 +164,12 @@ default <T> void withTransaction(Function<SqlConnection, Future<@Nullable T>> fu
         .onComplete(ar -> conn.close()));
   }
 
+  /**
+   * Like {@link #withTransaction(Function, Handler)} but keeps the connection accessible via the context
+   * for the duration of the given {@code function}
+   */
+  <T> Future<@Nullable T> withPropagatedTransaction(Function<SqlConnection, Future<@Nullable T>> function);
```
vietj, line 171: I think we should name this `withTransaction`.

Have you tried running the test case for this, with the existing withTransaction to see what is happening ?

technical-debt-collector: Agree, withPropagatedTransaction is not really pretty. However, we can't really name it withTransaction as the method signature collides with the existing one. Also, imho having distinguishable signatures seems like a good idea too to avoid confusion since the behaviour between the two differs.

I haven't done that, but I'll see if I get some time this weekend to look at it. Just mentioning it again that I have trouble running most of the tests. For now I've only been able to run the PG-tests as my dev environment doesn't seem to like the other ones.

technical-debt-collector: @vietj, I've tested a few different cases now by replacing withTransaction and withPropagatableTransaction with eachother.
Seems to be working fine;
- running withPropagatableTransaction inside the block of withTransaction results in another acquired connection and transaction. Inserting one entry using each of the two connection objects while also querying out the number of rows yields the expected result set size for each transaction inside their respective blocks.
- running withPropagatableTransaction, inserting something using its connection, querying out the result with the Pool.query (i.e. not using the aforementioned connection object) results in 0 rows, same goes withTransaction inside the same block.

The TransactionTestBase obviously needs to run with a larger pool size than 1 which it's currently set to. If you're fine with me increasing it then I'll gladly push a few more tests to verify the stuff mentioned above.

vietj: I think this cas discussed a while ago and we could have a property for with transaction (e.g an enum TransactionMode) that would govern how nested `withTransaction` would behave, e.g `TransactionMode.PROPAGATABLE`.

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlConnection.java
```diff
@@ -92,6 +92,8 @@ public interface SqlConnection extends SqlClient {
    */
   Future<Transaction> begin();
 
+  Transaction getTransaction();
```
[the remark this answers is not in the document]
vietj: need also javadoc like "return the current transaction if any otherwise return null"

vertx-sql-client/src/main/java/io/vertx/sqlclient/TransactionMode.java
```diff
@@ -0,0 +1,25 @@
+package io.vertx.sqlclient;
```
vietj, line 1: missing header

vertx-sql-client/src/main/java/io/vertx/sqlclient/TransactionMode.java
```diff
@@ -0,0 +1,25 @@
...
+ * Defines how the acquired connection will be managed during the execution of the function provided in
+ * {@link Pool#withTransaction(TransactionMode, Function)}.
+ */
+public enum TransactionMode {
+
+  /**
+   * The acquired connection is not stored anywhere, making it local to the provided function execution and to
+   * whereever it is passed.
+   */
+  DEFAULT,
```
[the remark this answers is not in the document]
technical-debt-collector: We could rename the enum to TransactionPropagationMode and the values to NONE and CONTEXT?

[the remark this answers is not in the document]
technical-debt-collector: Any thoughts on this?

technical-debt-collector:
@vietj is this still of any interest? If not then I'll just close it.

vietj:
let me check, this PR was just forgotten

vietj:
I believe it remains something we want, we just don't agreed on the naming. I'm not a fan of TransactionMode because it is too generic, I think instead we should rename `TransactionPropagation` which is more explicit

technical-debt-collector:
I took the liberty to rename the constants too, let me know if you don't like the new ones.

vietj:
thanks for the contribution

dozsa:
This is a helpful capability. Thank you for this enhancement. I saw the associated doc was not updated with this capability and how it would be used. Any plan to add this to the doc? Thanks.

vietj:
can you contribute the documentation ? that would help us

technical-debt-collector:
I'll see if I can find some time to look at this in the coming days

PR #1236, opened 2022-09-16
Fixes to MySQL pipelining

### LINKED ISSUE #1234: MySQL Client sends close statement commend before all rows are decoded
This was introduced in the PR for [pipelining](https://github.com/eclipse-vertx/vertx-sql-client/pull/1168) in order to avoid [unexpected order issues](https://github.com/eclipse-vertx/vertx-sql-client/issues/873).

However, it [breaks compatibility with ProxySQL](https://github.com/quarkusio/quarkus/issues/27386).

### LINKED ISSUE #1235: MySQLConnectOptions copy constructors ignore pipeliningLimit
The copy constructors ignore `pipeliningLimit`. In practice, this prevents pooled connection to support pipelining.

tsegismont:
Fixes #1234, fixes #1235

### DIFF
--- a/vertx-mysql-client/src/main/asciidoc/index.adoc
+++ b/vertx-mysql-client/src/main/asciidoc/index.adoc
@@ -106,6 +106,12 @@ You can configure the client to use pipelining
 
 The default pipelining limit is `1` which disables pipelining.
 
+[CAUTION]
+====
+Do not enable pipelining if you use a proxy that does not support it.
+Otherwise, the proxy might close client connections abruptly.
+====
+
 == Pool versus pooled client
 
 The {@link io.vertx.mysqlclient.MySQLPool} allows you to create a pool or a pooled client
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLConnectOptions.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLConnectOptions.java
@@ -103,6 +103,7 @@ public class MySQLConnectOptions extends SqlConnectOptions {
       this.serverRsaPublicKeyPath = opts.serverRsaPublicKeyPath;
       this.serverRsaPublicKeyValue = opts.serverRsaPublicKeyValue != null ? opts.serverRsaPublicKeyValue.copy() : null;
       this.characterEncoding = opts.characterEncoding;
+      this.pipeliningLimit = opts.pipeliningLimit;
       this.authenticationPlugin = opts.authenticationPlugin;
     }
   }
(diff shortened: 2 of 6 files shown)


tsegismont:
@BillyYccc in #873 you made a comment about ordering issues, can you elaborate? I've early closing in this PR and enabled pipeling for testing, it seems the suite passes. Did you notice issues in cases not tested here?

BillyYccc:
The client will automatically send prepared statement close command after executing query only if the client uses a one-shot prepared query  (i.e. using API `client#preparedQuery` with statement cache disabled) 

Before supporting pipelining we have an ordering issue in the close statement implementation, since the network protocol states the statement close command is a fire & forget command(see https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_stmt_close.html)
when we send a close statement command we immediately pop it from the request queue, this works fine when the pipelining limit is 1 however it will break the request queue order in pipelining mode

pipelining limit = 1 : prep | exec | close | prep | exec | close

pipelining limit > 1 encoding order: prep exec close prep exec close
pipelining limit > 1 decoding order: prep close(* decoding immediately after sending close, the exec resp is not ready) exec prep exec close

 (you can check the PR changes in https://github.com/eclipse-vertx/vertx-sql-client/pull/1168/files#diff-4c41cc49d53e95241af08ca5af42809a1c616304d4971c84ae3bf58f1123ef61L32 and https://github.com/eclipse-vertx/vertx-sql-client/pull/1168/files#diff-4c41cc49d53e95241af08ca5af42809a1c616304d4971c84ae3bf58f1123ef61L32)


I think we cannot simply remove the auto closing statement code in this PR otherwise it will leave the statement leaked, I wonder if the issue happens when using the client with a plain MySQL server or only with the SQLProxy, can you share a reproducer?

tsegismont:
> I think we cannot simply remove the auto closing statement code in this PR otherwise it will leave the statement leaked, 

I believe it doesn't, the statement is closed when all results have been decoded:

https://github.com/eclipse-vertx/vertx-sql-client/blob/f71739f5415aa66194788f80f640e7a0b42e6a2d/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/ExtendedQueryCommandBaseCodec.java#L48-L59

>I wonder if the issue happens when using the client with a plain MySQL server or only with the SQLProxy, can you share a reproducer?

It happens only with ProxySQL. It was originally reported in https://github.com/quarkusio/quarkus/issues/27386. I created a tool to setup Percona cluster with ProxySQL in front https://github.com/tsegismont/pxc-with-proxysql


Note that in this PR there is a change that enables pipelining in `MySQLPipeliningTest` and `MySQLPipeliningQueryTest`. Before that change it wasn't working because the connect options copy constructor ignored `pipeliningLimit`.

I have verified with Wireshark that pipelining is actually working with this PR.

BillyYccc (approved):
Thanks!

vietj (approved):
Need a test if we can afford it.

tsegismont:
@vietj yes, there's a test in the making ;-)

PR #1237, opened 2022-09-19
SQL Clients memory usage and performance improvements

tsegismont:
A series of improvements to memory usage of SQL clients in general and MySQL in particular.

Commits can be reviewed individually.

### DIFF
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
@@ -18,31 +18,88 @@ package io.vertx.mysqlclient.impl;
 
 import io.vertx.mysqlclient.impl.datatype.DataFormat;
 import io.vertx.mysqlclient.impl.protocol.ColumnDefinition;
+import io.vertx.sqlclient.desc.ColumnDescriptor;
 import io.vertx.sqlclient.impl.RowDesc;
 
-import java.util.Arrays;
+import java.util.AbstractList;
 import java.util.Collections;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
+import java.util.List;
 
 public class MySQLRowDesc extends RowDesc {
 
-  private final ColumnDefinition[] columnDefinitions;
+  public final ColumnDefinition[] columnDefinitions;
   private final DataFormat dataFormat;
 
-  public MySQLRowDesc(ColumnDefinition[] columnDefinitions, DataFormat dataFormat) {
-    super(Collections.unmodifiableList(Stream.of(columnDefinitions)
-      .map(ColumnDefinition::name)
-      .collect(Collectors.toList())), Collections.unmodifiableList(Arrays.asList(columnDefinitions)));
+  private MySQLRowDesc(List<String> columnNames, List<ColumnDescriptor> columnDescriptors, ColumnDefinition[] columnDefinitions, DataFormat dataFormat) {
+    super(columnNames, columnDescriptors);
     this.columnDefinitions = columnDefinitions;
     this.dataFormat = dataFormat;
   }
 
-  public ColumnDefinition[] columnDefinitions() {
-    return columnDefinitions;
+
+  public static MySQLRowDesc create(ColumnDefinition[] columnDefinitions, DataFormat dataFormat) {
+    if (columnDefinitions.length == 0) {
+      return new MySQLRowDesc(Collections.emptyList(), Collections.emptyList(), columnDefinitions, dataFormat);
+    }
+    List<String> columnNames = new ColumnNames(columnDefinitions);
+    List<ColumnDescriptor> columnDescriptors = new ColumnDescriptors(columnDefinitions);
+    return new MySQLRowDesc(columnNames, columnDescriptors, columnDefinitions, dataFormat);
+  }
+
+  public int size() {
+    return columnDefinitions.length;
+  }
+
+  public ColumnDefinition get(int index) {
+    if (index < 0 || index >= size()) {
+      throw new IndexOutOfBoundsException();
+    }
+    return columnDefinitions[index];
   }
 
   public DataFormat dataFormat() {
     return dataFormat;
+  }
+
+  private static class ColumnNames extends AbstractList<String> {
+    private final ColumnDefinition[] columnDefinitions;
+
+    public ColumnNames(ColumnDefinition[] columnDefinitions) {
+      this.columnDefinitions = columnDefinitions;
+    }
+
+    @Override
+    public String get(int index) {
+      if (index < 0 || index >= columnDefinitions.length) {
+        throw new IndexOutOfBoundsException();
+      }
+      return columnDefinitions[index].name();
+    }
+
+    @Override
+    public int size() {
+      return columnDefinitions.length;
+    }
+  }
+
+  private static class ColumnDescriptors extends AbstractList<ColumnDescriptor> {
+    private final ColumnDefinition[] columnDefinitions;
+
+    public ColumnDescriptors(ColumnDefinition[] columnDefinitions) {
+      this.columnDefinitions = columnDefinitions;
+    }
+
+    @Override
+    public ColumnDescriptor get(int index) {
+      if (index < 0 || index >= columnDefinitions.length) {
+        throw new IndexOutOfBoundsException();
+      }
+      return columnDefinitions[index];
+    }
+
+    @Override
+    public int size() {
+      return columnDefinitions.length;
+    }
   }
 }
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
@@ -24,6 +24,7 @@ import io.vertx.sqlclient.impl.RowDesc;
 import java.util.AbstractList;
 import java.util.Collections;
 import java.util.List;
+import java.util.RandomAccess;
 
 public class MySQLRowDesc extends RowDesc {
 
@@ -58,7 +59,7 @@ public class MySQLRowDesc extends RowDesc {
     return dataFormat;
   }
 
-  private static class ColumnNames extends AbstractList<String> {
+  private static class ColumnNames extends AbstractList<String> implements RandomAccess {
     private final ColumnDefinition[] columnDefinitions;
 
     public ColumnNames(ColumnDefinition[] columnDefinitions) {
@@ -73,6 +74,18 @@ public class MySQLRowDesc extends RowDesc {
     @Override
     public int size() {
       return columnDefinitions.length;
+    }
+
+    @Override
+    public int indexOf(Object o) {
+      if (o != null) {
+        for (int i = 0; i < columnDefinitions.length; i++) {
+          if (o.equals(columnDefinitions[i].name())) {
+            return i;
+          }
+        }
+      }
+      return -1;
     }
   }
 
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
@@ -80,7 +80,8 @@ public class MySQLRowDesc extends RowDesc {
     public int indexOf(Object o) {
       if (o != null) {
         for (int i = 0; i < columnDefinitions.length; i++) {
-          if (o.equals(columnDefinitions[i].name())) {
+          String name = columnDefinitions[i].name();
+          if (o.hashCode() == name.hashCode() && o.equals(name)) {
             return i;
           }
         }
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
@@ -18,37 +18,21 @@ package io.vertx.mysqlclient.impl;
 
 import io.vertx.mysqlclient.impl.datatype.DataFormat;
 import io.vertx.mysqlclient.impl.protocol.ColumnDefinition;
-import io.vertx.sqlclient.desc.ColumnDescriptor;
 import io.vertx.sqlclient.impl.RowDesc;
-
-import java.util.AbstractList;
-import java.util.Collections;
-import java.util.List;
-import java.util.RandomAccess;
 
 public class MySQLRowDesc extends RowDesc {
 
   private final ColumnDefinition[] columnDefinitions;
   private final DataFormat dataFormat;
 
-  private MySQLRowDesc(List<String> columnNames, List<ColumnDescriptor> columnDescriptors, ColumnDefinition[] columnDefinitions, DataFormat dataFormat) {
-    super(columnNames, columnDescriptors);
+  private MySQLRowDesc(ColumnDefinition[] columnDefinitions, DataFormat dataFormat) {
+    super(columnDefinitions);
     this.columnDefinitions = columnDefinitions;
     this.dataFormat = dataFormat;
   }
 
-
   public static MySQLRowDesc create(ColumnDefinition[] columnDefinitions, DataFormat dataFormat) {
-    if (columnDefinitions.length == 0) {
-      return new MySQLRowDesc(Collections.emptyList(), Collections.emptyList(), columnDefinitions, dataFormat);
-    }
-    List<String> columnNames = new ColumnNames(columnDefinitions);
-    List<ColumnDescriptor> columnDescriptors = new ColumnDescriptors(columnDefinitions);
-    return new MySQLRowDesc(columnNames, columnDescriptors, columnDefinitions, dataFormat);
-  }
-
-  public int size() {
-    return columnDefinitions.length;
+    return new MySQLRowDesc(columnDefinitions, dataFormat);
   }
 
   public ColumnDefinition[] columnDefinitions() {
@@ -61,54 +45,5 @@ public class MySQLRowDesc extends RowDesc {
 
   public DataFormat dataFormat() {
     return dataFormat;
-  }
-
-  private static class ColumnNames extends AbstractList<String> implements RandomAccess {
-    private final ColumnDefinition[] columnDefinitions;
-
-    public ColumnNames(ColumnDefinition[] columnDefinitions) {
-      this.columnDefinitions = columnDefinitions;
-    }
-
-    @Override
-    public String get(int index) {
-      return columnDefinitions[index].name();
-    }
-
-    @Override
-    public int size() {
-      return columnDefinitions.length;
-    }
-
-    @Override
-    public int indexOf(Object o) {
-      if (o != null) {
-        for (int i = 0; i < columnDefinitions.length; i++) {
-          String name = columnDefinitions[i].name();
-          if (o.hashCode() == name.hashCode() && o.equals(name)) {
-            return i;
-          }
-        }
-      }
-      return -1;
-    }
-  }
-
-  private static class ColumnDescriptors extends AbstractList<ColumnDescriptor> {
-    private final ColumnDefinition[] columnDefinitions;
-
-    public ColumnDescriptors(ColumnDefinition[] columnDefinitions) {
-      this.columnDefinitions = columnDefinitions;
-    }
-
-    @Override
-    public ColumnDescriptor get(int index) {
-      return columnDefinitions[index];
-    }
-
-    @Override
-    public int size() {
-      return columnDefinitions.length;
-    }
   }
 }
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
@@ -51,9 +51,6 @@ public class MySQLRowDesc extends RowDesc {
   }
 
   public ColumnDefinition get(int index) {
-    if (index < 0 || index >= size()) {
-      throw new IndexOutOfBoundsException();
-    }
     return columnDefinitions[index];
   }
 
@@ -70,9 +67,6 @@ public class MySQLRowDesc extends RowDesc {
 
     @Override
     public String get(int index) {
-      if (index < 0 || index >= columnDefinitions.length) {
-        throw new IndexOutOfBoundsException();
-      }
       return columnDefinitions[index].name();
     }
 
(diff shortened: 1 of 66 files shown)


tsegismont:
cc @franz1981 (I think you've reviewed my working branch already but here's the proposed change to Vert.x 4.3)

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/RowSetImpl.java
```diff
@@ -42,43 +40,90 @@ static <U> Collector<Row, RowSetImpl<U>, RowSet<U>> collector(Function<Row, U> m
...
+    }
+  }
+
+  @SuppressWarnings("unchecked")
   @Override
   public RowIterator<R> iterator() {
-    Iterator<R> i = list.iterator();
-    return new RowIterator<R>() {
+    if (rowAccumulator != null) {
+      return rowAccumulator.iterator();
```
franz1981, line 81: Three different types here is bad for the JIT: better 2 and `null` if empty or just 2  (choosing the most common ones).

vertx-sql-client/src/test/java/io/vertx/sqlclient/benchmarks/AccumulatorReadBenchmark.java
```diff
@@ -0,0 +1,62 @@
...
+import static io.vertx.sqlclient.benchmarks.Utils.generateStrings;
+
+@Threads(1)
+@State(Scope.Thread)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Warmup(iterations = 20, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 10, time = 2, timeUnit = TimeUnit.SECONDS)
+@Fork(value = 1, jvmArgs = {"-Xms8g", "-Xmx8g", "-Xmn7g"})
+public class AccumulatorReadBenchmark {
```
franz1981, line 30: Will run today I promise :) I'm fighting with microbenchs as well recently

franz1981 (changes requested):
Just minor changes, great work!
Feel free to past some of the results after applying this, just to make easier for the future ourselves remember how good the results are :)

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
```diff
@@ -18,31 +18,88 @@
...
+    List<ColumnDescriptor> columnDescriptors = new ColumnDescriptors(columnDefinitions);
+    return new MySQLRowDesc(columnNames, columnDescriptors, columnDefinitions, dataFormat);
+  }
+
+  public int size() {
+    return columnDefinitions.length;
+  }
+
+  public ColumnDefinition get(int index) {
+    if (index < 0 || index >= size()) {
```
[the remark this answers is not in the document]
tsegismont: No, it is not. I guess `ArrayList` has it because the index could be within the range of capacity but not of the size.

I'll remove it.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
```diff
@@ -18,31 +18,88 @@
...
+      throw new IndexOutOfBoundsException();
+    }
+    return columnDefinitions[index];
   }
 
   public DataFormat dataFormat() {
     return dataFormat;
   }
+
+  private static class ColumnNames extends AbstractList<String> {
```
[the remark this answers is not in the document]
tsegismont: Not sure what you mean, it extends `AbstractList` which throws `UnsupportedOperationException` on modifiers.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/CommandCodec.java
```diff
@@ -169,19 +155,19 @@ String readRestOfPacketString(ByteBuf payload, Charset charset) {
   }
 
   ColumnDefinition decodeColumnDefinitionPacketPayload(ByteBuf payload) {
-    String catalog = BufferUtils.readLengthEncodedString(payload, StandardCharsets.UTF_8);
-    String schema = BufferUtils.readLengthEncodedString(payload, StandardCharsets.UTF_8);
-    String table = BufferUtils.readLengthEncodedString(payload, StandardCharsets.UTF_8);
-    String orgTable = BufferUtils.readLengthEncodedString(payload, StandardCharsets.UTF_8);
+    payload.skipBytes((int) BufferUtils.readLengthEncodedInteger(payload)); // catalogq
```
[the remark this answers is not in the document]
tsegismont: I'll reduce the number of invocations.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/MySQLDecoder.java
```diff
@@ -12,70 +12,74 @@
...
+import static io.vertx.mysqlclient.impl.protocol.Packets.PACKET_PAYLOAD_LENGTH_LIMIT;
 
-class MySQLDecoder extends ByteToMessageDecoder {
+class MySQLDecoder extends ChannelInboundHandlerAdapter {
 
   private final ArrayDeque<CommandCodec<?, ?>> inflight;
-  private final MySQLSocketConnection socketConnection;
 
-  private CompositeByteBuf aggregatedPacketPayload = null;
+  private ByteBufAllocator alloc;
```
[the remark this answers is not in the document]
tsegismont: ok

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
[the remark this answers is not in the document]
franz1981: If outside it's used as `ColumnNames` and never as `AbstractList` (if not for few/one single trivial method) you can save extending it.
It will make the JIT live a bit easier (not another list implementor bla bla bla) - but I'm ok whatever choice you made here, really

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/RowSetImpl.java
tsegismont: ok

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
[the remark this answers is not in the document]
tsegismont: Outside it's never used as `ColumnNames`, only as `List`:

https://github.com/eclipse-vertx/vertx-sql-client/blob/cf1da8a1652f5d1cefaec7c17cd892bfbb4d1448/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java#L44-L46

https://github.com/eclipse-vertx/vertx-sql-client/blob/cf1da8a1652f5d1cefaec7c17cd892bfbb4d1448/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java#L33-L34

https://github.com/eclipse-vertx/vertx-sql-client/blob/3b303d98f2410455eb47f01ac54d03f1c3dfccdf/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/RowDesc.java#L62-L65

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/MySQLDecoder.java
```diff
@@ -12,70 +12,74 @@
...
-      int sequenceId = in.readUnsignedByte();
+  public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
+    alloc = ctx.alloc();
+  }
 
-      if (payloadLength >= PACKET_PAYLOAD_LENGTH_LIMIT && aggregatedPacketPayload == null) {
-        aggregatedPacketPayload = ctx.alloc().compositeBuffer();
-      }
+  @Override
+  public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
```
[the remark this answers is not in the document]
vietj: it is not useful because that works mainly with servers, on a client you write data first and then read it.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
```diff
@@ -18,31 +18,93 @@
...
+
+  public ColumnDefinition get(int index) {
+    return columnDefinitions[index];
   }
 
   public DataFormat dataFormat() {
     return dataFormat;
   }
+
+  private static class ColumnNames extends AbstractList<String> implements RandomAccess {
```
vietj, line 62: eventually this should be moved to `RowDesc` is a follow up PR

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
```diff
@@ -18,31 +18,93 @@
...
+          if (o.hashCode() == name.hashCode() && o.equals(name)) {
+            return i;
+          }
+        }
+      }
+      return -1;
+    }
+  }
+
+  private static class ColumnDescriptors extends AbstractList<ColumnDescriptor> {
```
vietj, line 93: eventually this should be moved to `RowDesc` is a follow up PR

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
```diff
@@ -18,31 +18,93 @@
...
 import java.util.Collections;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
+import java.util.List;
+import java.util.RandomAccess;
 
 public class MySQLRowDesc extends RowDesc {
 
-  private final ColumnDefinition[] columnDefinitions;
+  public final ColumnDefinition[] columnDefinitions;
```
vietj, line 31: it does not seem worth to be public instead of using the accessor

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/MySQLDecoder.java
```diff
@@ -14,68 +14,65 @@
...
+import io.netty.channel.ChannelInboundHandlerAdapter;
 
 import java.util.ArrayDeque;
-import java.util.List;
 
-import static io.vertx.mysqlclient.impl.protocol.Packets.*;
+import static io.vertx.mysqlclient.impl.protocol.Packets.PACKET_PAYLOAD_LENGTH_LIMIT;
 
-class MySQLDecoder extends ByteToMessageDecoder {
+class MySQLDecoder extends ChannelInboundHandlerAdapter {
```
vietj, line 23: you should look at the PgDecoder that is very similar, it is written in a way that it will slice packet itself without using `LengthFieldBasedFrameDecoder` which saves an inbound handler in the channel pipeline.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/protocol/ColumnDefinition.java
```diff
@@ -17,77 +17,30 @@
...
 public final class ColumnDefinition implements ColumnDescriptor {
+
   /*
     https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_query_response_text_resultset_column_definition.html
    */
-  private final String catalog;
-  private final String schema;
-  private final String table;
-  private final String orgTable;
   private final String name;
```
vietj, line 24: add a comment saying that this is a subset of the data that we actually need in practice

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
tsegismont: In fact, I may do this when applying changes to other clients.

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/MySQLRowDesc.java
tsegismont: In fact, I may do this when applying changes to other clients.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/accumulator/Accumulator.java
```diff
@@ -0,0 +1,24 @@
...
+package io.vertx.sqlclient.impl.accumulator;
+
+import io.vertx.sqlclient.RowIterator;
+
+import java.util.function.Consumer;
+
+/**
+ * A data structure similar to a {@link java.util.Collection} but which is append-only.
+ */
+public interface Accumulator<T> extends Consumer<T>, Iterable<T> {
```
vietj, line 21: should be named `RowAccumulator`

vietj (changes requested):
a few comments inline

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/CommandCodec.java
```diff
@@ -169,19 +155,36 @@ String readRestOfPacketString(ByteBuf payload, Charset charset) {
...
-    String orgTable = BufferUtils.readLengthEncodedString(payload, StandardCharsets.UTF_8);
+    int start = payload.readerIndex();
+    int bytesToSkip = 0;
+
+    bytesToSkip += BufferUtils.countBytesOfLengthEncodedString(payload, start + bytesToSkip); // catalog
+    bytesToSkip += BufferUtils.countBytesOfLengthEncodedString(payload, start + bytesToSkip); // schema
+    bytesToSkip += BufferUtils.countBytesOfLengthEncodedString(payload, start + bytesToSkip); // table
+    bytesToSkip += BufferUtils.countBytesOfLengthEncodedString(payload, start + bytesToSkip); // orgTable
+
+    payload.skipBytes(bytesToSkip);
```
franz1981, line 166: lovely, you saved many bound checks here by having just 2 skipBytes

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/PropertyKindMap.java
```diff
@@ -0,0 +1,63 @@
...
+
+  private Object[] elements;
+  private int count;
+
+  public PropertyKindMap() {
+    elements = new Object[2];
+    count = 0;
+  }
+
+  public Object get(PropertyKind<?> property) {
```
franz1981, line 32:
```suggestion
  public Object get(PropertyKind<?> property) {
      final Object[] elements = this.elements;
      for (int i = 0, size = elements.length; i < size; i+=2) {
         final Object key = elements[i];
         assert key != null : "key cannot be null"; 
         if (key == property || key.equals(property)) {
             return elements[i + 1];
         }  
      }
      return null;
  }
```
This will help JIT to correlate the loop variables with the length of the array, saving bound checks and inlining the equality check would help in case key is effectively the same (if it's not common, you can drop the ref equality check)

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/PropertyKindMap.java
```diff
@@ -0,0 +1,63 @@
...
+      int idx = 2 * i;
+      if (property.equals(elements[idx])) {
+        return elements[idx + 1];
+      }
+    }
+    return null;
+  }
+
+  public void put(PropertyKind<?> property, Object value) {
+    for (int i = 0; i < count; i++) {
```
franz1981, line 43: I would perform the same loop as in the previous comment, that help reduce bound checks

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/PropertyKindMap.java
```diff
@@ -0,0 +1,63 @@
...
+        return;
+      }
+    }
+    int idx = 2 * count;
+    if (idx == elements.length) {
+      elements = Arrays.copyOf(elements, 2 * (count + 1));
+    }
+    elements[idx] = property;
+    elements[idx + 1] = value;
+    count++;
```
franz1981, line 56: count is that necessary?
you can just use elements.length / 2

vertx-sql-client/src/test/java/io/vertx/sqlclient/benchmarks/AccumulatorReadBenchmark.java
```diff
@@ -0,0 +1,62 @@
...
+    accumulator = accumulatorType.newInstance();
+    for (String s : arr) {
+      accumulator.accept(s);
+    }
+  }
+
+  @Benchmark
+  public void iterate(Blackhole bh) {
+    for (String s : accumulator) {
+      bh.consume(s.hashCode());
```
franz1981, line 59: you can just consume `s`?

franz1981: or you can do something better here: `BlackHole::consume` preven some JIT loop transformations that are key to compare between versions:

```java
  @Benchmark
  public int iterate() {
    int dummy = 0;
    for (String s : accumulator) {
      dummy += s.length();
    }
    return dummy;
}      
```
This will make the loop as fast as it can as in the real world and will prevent it to be dead-code-eliminated

I suggest to re-run the bench after changing this

vertx-sql-client/src/test/java/io/vertx/sqlclient/benchmarks/AccumulatorWriteBenchmark.java
```diff
@@ -0,0 +1,60 @@
...
+
+import static io.vertx.sqlclient.benchmarks.Utils.generateStrings;
+
+@Threads(1)
+@State(Scope.Thread)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Warmup(iterations = 20, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 10, time = 2, timeUnit = TimeUnit.SECONDS)
+@Fork(value = 1, jvmArgs = {"-Xms8g", "-Xmx8g", "-Xmn7g"})
```
franz1981, line 29: value >= 2

franz1981 (changes requested):
LGTM minor changes

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/PropertyKindMap.java
tsegismont: right, I'll remove it

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/PropertyKindMap.java
tsegismont: I just pushed 8620798d34eb9f84942f2a13f78bfa33d3d4831d to remove the computed `count` field. At the same time, I've changed the loop to use the `elements.length` in condition check. Can you verify it's enough or is still better to make a local copy of the reference to the array?

vertx-sql-client/src/test/java/io/vertx/sqlclient/benchmarks/AccumulatorWriteBenchmark.java
tsegismont: Sorry, can you explain why it is needed?

tsegismont: I found this https://stackoverflow.com/a/35147232/2133695

In 97039b66 I've improved the benchmarks. I'll run them tonight because the ETA with 3 forks is more than 7 hours ahead :sweat_smile:

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/MySQLDecoder.java
tsegismont: Let's address that in a separate PR. I got rid of the last `slice()` calls in this decoder in 36574ee1.

vertx-sql-client/src/test/java/io/vertx/sqlclient/benchmarks/AccumulatorWriteBenchmark.java
tsegismont: I've rolled-back to `ArrayListRowAccumulator` in edb479f

Latest test results do not demonstrate good results for `ChunkedRowAccumulator`.
Until we sort this out, let's continue with `ArrayListRowAccumulator`.

The biggest improvement here anyway is to avoid creating an accumulator if the database returns no rows or a single (which happens very often).

tsegismont:
@vietj @franz1981 PTAL

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/accumulator/ChunkedRowAccumulator.java
```diff
@@ -0,0 +1,116 @@
...
+ */
+
+package io.vertx.sqlclient.impl.accumulator;
+
+import io.vertx.sqlclient.RowIterator;
+
+import java.util.NoSuchElementException;
+import java.util.function.IntUnaryOperator;
+
+public class ChunkedRowAccumulator<T> implements RowAccumulator<T> {
```
vietj, line 19: do we actually use this in our code ?

tsegismont: Not yet. I've met with @franz1981 this morning to talk about the last benchmarks results I got. I will run another round of tests this afternoon.

So we can keep it until we make a final decision. But let's not block merging of this PR because of this.

vietj: sure, but I'm not convinced that the complexity it brings is worth it

tsegismont: We'll see what the new results say. It won't be a game changer regarding latency but allocations may be reduced significantly.

Any other comments about the recent commits?

vietj: all seem fine, we'll rediscuss th chunked accumulaor later then

vietj (approved):

PR #1240, opened 2022-09-29
Use chunked row accumulator for RowSet implementation

tsegismont:
Benchmarks have been updated after another round of review. Also, a new Write/Read benchmark has been added which gives a better idea of what happens when we accumulate rows from DBs then iterate the result set.

The results below are taken from the new Write/Read benchmark

# Regarding latency

- chunked accumulator performs 37% / 26% / 16% / 7% faster with 5 / 20 / 65 / 605 rows
- chunked accumulator performs 25% / 28% slower with 1820 / 5465 rows

The fact that ArrayList grows by 50% when changing capacity may explain that it catches up for large result sets because it needs less iterations before reaching the required total.

The Write only and Read only benchmarks seem to confirm this: ChunkedAccumulator is always faster when adding rows but, in pure reading, ArrayList is faster when the result set is bigger than 10 rows.

Note that when the result set is smaller than 10 rows, ChunkedAccumulator is also 59% faster for reading. This may be explained by the extra complexity in ArrayList (iterator and modification checking).

# Regarding allocations

- chunked accumulator allocates 14% / 46% / 51% / 61% / 48% / 53% less bytes per operation for 5 / 20 / 65 / 605 / 1820 / 5465 rows

The improvement is not as important with result sets smaller than 10 rows because ArrayList does not have to copy its internal array.

# Conclusion

The chunked accumulator performs faster with fewer allocations when dealing with result set sizes which are common in OLTP (few to a thousand rows).

For large result sets, it performs slower but allocates much less.


tsegismont:
@franz1981 in this PR you will find the benchmarks updated as discussed yesterday.

I ran the test using `gc` and `shuffle` params both set to `true`.

tsegismont:
Here are the JSON results: [jmh-result.zip](https://github.com/eclipse-vertx/vertx-sql-client/files/9672424/jmh-result.zip)

Pasted them in https://jmh.morethan.io/ for visualization.

For throughput.: [throughput.pdf](https://github.com/eclipse-vertx/vertx-sql-client/files/9672451/throughput.pdf)

For allocations: [alloc.pdf](https://github.com/eclipse-vertx/vertx-sql-client/files/9672450/alloc.pdf)

Note that in the captures, scales are logarithmic.

vietj:
I am still not convinced by this benchmark, before going further I want to see the impact on an actual benchmark that interacts with a database and not a microbenchmark.

vietj (commented):
I am still not convinced by this benchmark, before going further I want to see the impact on an actual benchmark that interacts with a database and not use a microbenchmark as decision factor.

PR #1244, opened 2022-10-14
Revert Buffer slice() call replacement in SQL Clients …

tsegismont:
… where it makes sense 

There are still legitimate usages of slice() that we can make in SQL Clients, which should be easily replaceable with split() when we upgrade to Netty 5


tsegismont:
Follows-up on 2c619c672a0ae5c601b55dff1a40cad0990baca6

PR #1248, opened 2022-10-28
Bugfix/fix array accessor in tuple

### LINKED ISSUE #1247: Tuple is not working if access primitive array
### Version

4.x.x

### Context

With this check, the result will be `false` if an object is a primitive array, then the accessor will throw ClassCastException
```java
val.getClass() == Object[].class
```

Affect to all `getArrayOfXXX` where XXX is type can be primitive type (Boolean/boolean, Integer/int...)

E.g: 
https://github.com/eclipse-vertx/vertx-sql-client/blob/75ee2e4ad73d0fcf736efd80f5ea4569b72f898a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java#L802-L808

zero88:
Motivation:

Fix  https://github.com/vert-x3/vertx-jdbc-client/issues/290
Fix #1247

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java
@@ -35,8 +35,11 @@ import java.time.*;
 import java.time.temporal.Temporal;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
+import java.util.stream.IntStream;
 
 /**
  * A general purpose tuple.
@@ -428,7 +431,16 @@ public interface Tuple {
    * @return the value
    */
   default JsonArray getJsonArray(int pos) {
-    return (JsonArray) getValue(pos);
+    Object val = getValue(pos);
+    if (val == null) return null;
+    if (val instanceof JsonArray) return (JsonArray) val;
+    if (val instanceof Collection) {
+      return new JsonArray(new ArrayList<>(((Collection<?>) val)));
+    }
+    if (val.getClass().isArray()) {
+      return IntStream.range(0, Array.getLength(val)).mapToObj(i -> Array.get(val, i)).collect(JsonArray::new, JsonArray::add, (o1, o2) -> {});
+    }
+    throw new ClassCastException("Invalid JsonArray value type " + val.getClass());
   }
 
   /**
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java
@@ -29,17 +29,14 @@ import io.vertx.sqlclient.data.Numeric;
 import io.vertx.sqlclient.impl.ArrayTuple;
 import io.vertx.sqlclient.impl.ListTuple;
 
-import java.lang.reflect.Array;
 import java.math.BigDecimal;
 import java.time.*;
 import java.time.temporal.Temporal;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
-import java.util.stream.IntStream;
 
 /**
  * A general purpose tuple.
@@ -74,7 +71,6 @@ public interface Tuple {
    *
    * @return the list wrapped as a tuple
    */
-  @SuppressWarnings("unchecked")
   static <T> Tuple from(List<T> list) {
     return wrap(new ArrayList<>(list));
   }
@@ -88,10 +84,8 @@ public interface Tuple {
    */
   @GenIgnore(GenIgnore.PERMITTED_TYPE)
   static Tuple from(Object[] array) {
-    ArrayList<Object> list = new ArrayList<>(array.length);
-    for (Object o : array) {
-      list.add(o);
-    }
+    List<Object> list = new ArrayList<>(array.length);
+    Collections.addAll(list, array);
     return wrap(list);
   }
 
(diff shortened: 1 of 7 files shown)


zero88:
Hi @vietj , could you pls approve to run CI test

zero88:
Ah, actually some integration test fails related to `ColumnChecker`
Let me re-check, have not yet understood this assertion

zero88:
Ah, actually some integration test fails related to `ColumnChecker`
Let me re-check, have not yet understood this assertion

vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java
```diff
@@ -85,10 +87,8 @@ static <T> Tuple from(List<T> list) {
    */
   @GenIgnore(GenIgnore.PERMITTED_TYPE)
   static Tuple from(Object[] array) {
-    ArrayList<Object> list = new ArrayList<>(array.length);
-    for (Object o : array) {
-      list.add(o);
-    }
+    List<Object> list = new ArrayList<>(array.length);
```
tsegismont, line 90: This does not bring anything, please revert it.

vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java
```diff
@@ -202,7 +202,7 @@ static Tuple of(Object elt1, Object elt2, Object elt3, Object elt4, Object elt5)
    * Create a tuple of six elements.
    *
    * @param elt1 the first value
-   * @param elt2 the second valueg
+   * @param elt2 the second value
```
tsegismont, line 205: nice catch

vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java
```diff
@@ -428,7 +428,16 @@ default JsonObject getJsonObject(int pos) {
...
   default JsonArray getJsonArray(int pos) {
-    return (JsonArray) getValue(pos);
+    Object val = getValue(pos);
+    if (val == null) return null;
+    if (val instanceof JsonArray) return (JsonArray) val;
+    if (val instanceof Collection) {
+      return new JsonArray(new ArrayList<>(((Collection<?>) val)));
+    }
+    if (val.getClass().isArray()) {
+      return IntStream.range(0, Array.getLength(val)).mapToObj(i -> Array.get(val, i)).collect(JsonArray::new, JsonArray::add, (o1, o2) -> {});
```
tsegismont, line 438: It would be better to avoid the stream API on this part. It is not really useful and allocates much more memory.

tsegismont: This applies to the other,  similar changes

vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java
```diff
@@ -645,20 +654,12 @@ default BigDecimal getBigDecimal(int pos) {
...
-      Boolean[] booleanArray = new Boolean[array.length];
-      for (int i = 0;i < array.length;i++) {
-        booleanArray[i] = (Boolean) array[i];
-      }
-      return booleanArray;
-    } else {
-      return (Boolean[]) val; // Throw CCE
+    if (val == null) return null;
+    if (val instanceof Boolean[]) return (Boolean[]) val;
+    if (val.getClass().isArray()) {
```
tsegismont, line 659: we could have a top level `if` that checks if `val` is an array, and then other tests inside the block which inspect the `componentType`

tsegismont: This applies to the other, similar changes

tsegismont (commented):
Thank you @zero88 , a first round of comments

zero88:
After reviewing `Tuple` implementation, I think many inconsistent between `getXX` and `getArrayOfXX`, and it is very cumbersome. 
For example, `getXX` has the coercion converter from another type to the target type but `getArrayOfXX` is missing.

Then, I propose `ChainConverter` to make the implementation simpler. 

Could you have a quick look before I fix the integration test? @vietj @tsegismont

pmlopes (approved):

vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java
```diff
@@ -274,18 +271,10 @@ default Boolean getBoolean(int pos) {
...
-    } else if (val instanceof Short) {
-      return (Short) val;
-    } else if (val instanceof Number) {
-      return ((Number) val).shortValue();
-    } else if (val instanceof Enum<?>) {
-      return (short)((Enum<?>) val).ordinal();
-    } else {
-      return (Short) val; // Throw CCE
-    }
+    return ChainConverter.allowCast(Short.class)
```
vietj, line 274: this seems to not be GC free

vietj:
we should avoid using primitive arrays when tuple are constructed instead of trying to deal with them.

zero88:
Hi @vietj 
These are 2 different issues.

The first one is a primitive array

Actually, I can do 
```java
Tuple t = Tuple.of(new int[]{1, 2, 3})
t.getArrayOfIntegers(0)
```
But actually, no "reality" use case for this on SQL. The prepared parameters can be array, but most SQL client (e.g JDBC) uses an Object array instead of a primitive array.

So, I agree about 
> we should avoid using primitive arrays with Tuple

zero88:
The second one is https://github.com/vert-x3/vertx-jdbc-client/issues/290
Should Tuple store a real object and delegate to each method can convert it to the expected type?

For example:
In the query for the integer array field, the actual value is `Integer{}[1, 2, 3]`, then 
- `getObject` can return `Integer{}[1, 2, 3]`
- `getJsonArray` can return `JsonArray.of(1, 2, 3)`
- `getArrayOfIntegers` can return `Integer{}[1, 2, 3]`, and the same for coercion methods such as `getArrayOfShorts`, etc..

PR #1252, opened 2022-11-07
Type cache pollution: consistently check for TupleInternal instead of Tuple type compatibility

### LINKED ISSUE #1251: Type pollution agent report
### Version

Using version `4.3.4`.

### Context

I've been running the Techempower benchmark with Quarkus + Hibernate Reactive + Vert.x `vertx-pg-client` and attaching the [agent to detect the type pollution cases](https://github.com/RedHatPerf/type-pollution-agent).

This agent is able to report cases of [JDK-8180450](https://bugs.openjdk.org/browse/JDK-8180450) in our libraries, which are very hard to spot otherwise but result in extreme performance penalties, so I'd recommend treating alerts from it at high priority - especially as fixes might be trivial, while returning very high performance impact.

This is the report output using today's Quarkus snapshot:

```
4:	io.vertx.sqlclient.impl.ListTuple
Count:	14491599
Types:
	io.vertx.sqlclient.impl.TupleInternal
	io.vertx.sqlclient.Tuple
Traces:
	io.vertx.sqlclient.impl.command.ExtendedQueryCommand.prepare(ExtendedQueryCommand.java:122)
		class: io.vertx.sqlclient.impl.TupleInternal
		count: 7070022
	io.vertx.sqlclient.impl.command.ExtendedQueryCommand.params(ExtendedQueryCommand.java:160)
		class: io.vertx.sqlclient.Tuple
		count: 7029308
	io.vertx.sqlclient.impl.command.ExtendedQueryCommand.prepare(ExtendedQueryCommand.java:115)
		class: io.vertx.sqlclient.impl.TupleInternal
		count: 193981
	io.vertx.sqlclient.impl.command.ExtendedQueryCommand.prepare(ExtendedQueryCommand.java:114)
		class: io.vertx.sqlclient.Tuple
		count: 179589
	io.vertx.pgclient.impl.codec.ExtendedQueryCommandCodec.encode(ExtendedQueryCommandCodec.java:51)
		class: io.vertx.sqlclient.Tuple
		count: 18699
```

Interpreting this, the problem is trivial: the same concrete type `io.vertx.sqlclient.impl.ListTuple` is occasionally being tested for being of type `io.vertx.sqlclient.impl.TupleInternal`, and occasionally for type `io.vertx.sqlclient.Tuple`.
Such alternating checks cause a problem with the internal type cache in hotspot, the simplest way to fix it is to always check for the same type.

Perusing the object hierarchy it seems any implementation of `Tuple` is always also a valid cast to `TupleInternal`, so my suggestion would be for all internal code to consistently cast to `TupleInternal`, and let the type get widened as needed (which doesn't need a check).

cc/ @tsegismont @vietj @franz1981

Sanne:
Proposed Fixes #1251

### DIFF
--- a/vertx-db2-client/src/main/java/io/vertx/db2client/impl/codec/ExtendedBatchQueryCommandCodec.java
+++ b/vertx-db2-client/src/main/java/io/vertx/db2client/impl/codec/ExtendedBatchQueryCommandCodec.java
@@ -27,6 +27,7 @@ import io.vertx.db2client.impl.codec.DB2PreparedStatement.QueryInstance;
 import io.vertx.db2client.impl.drda.DRDAQueryRequest;
 import io.vertx.db2client.impl.drda.DRDAQueryResponse;
 import io.vertx.sqlclient.Tuple;
+import io.vertx.sqlclient.impl.TupleInternal;
 import io.vertx.sqlclient.impl.command.CommandResponse;
 import io.vertx.sqlclient.impl.command.ExtendedQueryCommand;
 
@@ -34,7 +35,7 @@ class ExtendedBatchQueryCommandCodec<R> extends ExtendedQueryCommandBaseCodec<R,
 
   private static final Logger LOG = LoggerFactory.getLogger(ExtendedBatchQueryCommandCodec.class);
 
-  private final List<Tuple> params;
+  private final List<TupleInternal> params;
   private final List<QueryInstance> queryInstances;
   private final String baseCursorId;
 
--- a/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedBatchQueryCommandCodec.java
+++ b/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/ExtendedBatchQueryCommandCodec.java
@@ -12,7 +12,6 @@
 package io.vertx.mssqlclient.impl.codec;
 
 import io.netty.buffer.ByteBuf;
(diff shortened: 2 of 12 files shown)


vietj:
could we instead make TupleInternal an abstract class ?

Sanne:
I guess that could work too. But I should say I like this patch, as it's showing also that some casts become unnecessary - and it's all internal code.

BTW I've been re-running the benchmark in the meantime and verified that this patch does indeed resolve the problem, so I'd prefer to apply it and move on as there's several other cases to address :)

vietj:
the reason I prefer TupleInternal as abstract class is that it saves future work where casts would be done in another place of the codebase in further contributions

vietj:
we can have both TBD btw

Sanne:
converted the TupleInternal to become an abstract class

Sanne:
BTW I'm having some issues with running the integration tests locally - I'm trusting CI to tell me about mistakes, I hope that's ok.

vietj:
the PR should provide results I think

Sanne:
> the PR should provide results I think

I've run the agent and verified that it fixes the scalability issue. However I'm unable to produce performance results at this point specifically for this single patch.

We've verified that fixing such scalability issues does indeed translate to significant performance gains (with a different reproducer), but to verify these specific patches it's far more convenient to have released libraries so that the performance team doesn't have to rebuild dozens of custom projects in the lab.  I hope basing our tests on the agent for now is a good compromise?

vietj:
thanks @Sanne for the contribution

PR #1256, opened 2022-11-13
enable Reproducible Builds

hboutemy:
Motivation:

see https://maven.apache.org/guides/mini/guide-reproducible-builds.html

Conformance:


vietj:
shouldn't this be in vertx parent pom instead ?

PR #1257, opened 2022-11-15
Do not send MySQL batches at once if pipelining is disabled

### LINKED ISSUE #1254: MySQL batch executes in pipelining mode even when pipelining is disabled
In #861 , MySQL batch execution implementation has been modified to use pipelining in all circumstances.

We should be able to execute queries one after the other when pipelining is disabled.

tsegismont:
Fixes #1254

For example, if you send queries through ProxySQL, the batched operation will fail.

Also:
- make sure errors are reported correctly
- created a ProxySQL rule to be able to easily create more tests for it


vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/ExtendedBatchQueryCommandCodec.java
```diff
@@ -22,79 +22,106 @@
...
 import java.util.List;
 
-import static io.vertx.mysqlclient.impl.protocol.Packets.EnumCursorType.CURSOR_TYPE_NO_CURSOR;
+import static io.vertx.mysqlclient.impl.protocol.Packets.EnumCursorType.*;
 
 class ExtendedBatchQueryCommandCodec<R> extends ExtendedQueryCommandBaseCodec<R, ExtendedQueryCommand<R>> {
 
   private final List<TupleInternal> params;
-  private int batchIdx = 0;
+  private final boolean[] bindingFailures;
```
vietj, line 30: you can use java.util.BitSet instead, it provides a more compact representation of the same values

vietj (approved):

tsegismont:
@vietj thanks for the review. Indeed, using a `BitSet` is a good idea. I've added a commit and will merge after CI jobs finish.

PR #1260, opened 2022-11-28
Add typecheck for Tuple.of

gazcn007:
## Motivation:

There are many queries in SQL where you need to send in an array of parameters, i.e. the famous Any(...) statement`SELECT * FROM table WHERE id = 10 AND other_id = ANY(1,2,3,4)`

The Tuple.of uses `Object... elts` for multi-varaibles support, however it ignores the case where the second parameter is an array:
e574ecb7-374b-4282-869e-8a4c2b2e6bab

The query crashes in this case below:
```java
int[] otherIds = [1,2,3]
int id = 1

// Query
PostgresUtils
          .getClient()
          .preparedQuery("SELECT * FROM table WHERE id = $1 AND other_ids =  ANY($2)")
          .execute(Tuple.of(userId, otherIds.toArray(Number[]::new)))
```
Since `Tuple.of(id, otherIds)` gets you `{1, 1, 2, 3}` not `{1, [1,2,3]}` ❌
   - because the `otherIds` Array is destructured into single parameters

For all other cases, this method runs well:
- Tuple.of(otherIds, id) gets you {[1,2,3], 1} which is valid ✅
- Tuple.of(id, otherIds, otherIds) gets you {1, [1,2,3], [1,2,3]} which is valid ✅

I am adding a type check to see if `elts` is an array, which avoids this edge case 🎉.

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java
@@ -231,8 +231,12 @@ public interface Tuple {
   static Tuple of(Object elt1, Object... elts) {
     ArrayTuple tuple = new ArrayTuple(1 + elts.length);
     tuple.addValue(elt1);
-    for (Object elt: elts) {
-      tuple.addValue(elt);
+    if(elts.getClass().isArray()){
+      tuple.addValue(elts);
+    } else {
+      for (Object elt: elts) {
+        tuple.addValue(elt);
+      }
     }
     return tuple;
   }


gazcn007:
Coming to think of it, this is actually a pitfall within Java, Java treats `Object...` the same as `Object[]`, it handles the varargs as an array:
```
public static void foo(Object... args) { ... }
```

is the same as 
```
public static void foo(Object[] args) { ... }
```

Hence, simply adding a `getClass().isArray()` won't fix this issue

```java
Object[] array = {
   1, 2, 3, 4, 5, 6, 7
};
Tuple tuple = Tuple.of(1, array);
assertEquals(2, tuple.size()); // => ✅

Tuple tuple2 = Tuple.of(1, 2, 3, 4, 5, 6, 7);
assertEquals(7, tuple2.size()); // => ❌，tuple2.size() is also 2
``` 
It looks like the only way to tell if the input parameter is varargs or an array is through the [isVarArgs()](https://www.geeksforgeeks.org/method-class-isvarargs-method-in-java/) method

I will close this PR for now.

As an workaround, if the second parameter is an array, we can cast it to a primitive array instead of Object array, an primitive array is an Object:

```java
List<integer> intList = new ArrayList<>();
intList.add(1);
intList.add(2);

// this calls "static Tuple of(Object elt1, Object elt2)"
Tuple.of(1, intList.stream().mapToInt(Integer::intValue).toArray());
// this calls "static Tuple of(Object elt1, Object... elts)"
Tuple.of(1, intList.toArray(Number[]::new)); 

```

gazcn007:
As an workaround, if the second parameter is an array, we can cast it to a primitive array instead of Object array, an primitive array is an Object:

```java
List<integer> intList = new ArrayList<>();
intList.add(1);
intList.add(2);

// this calls "static Tuple of(Object elt1, Object elt2)"
Tuple.of(1, intList.stream().mapToInt(Integer::intValue).toArray());
// this calls "static Tuple of(Object elt1, Object... elts)"
Tuple.of(1, intList.toArray(Number[]::new)); 
```

*However, if it's String[], it would be troublesome... Ideas appreciated*
it's fairly common to compose a query like:
```SQL
SELECT * FROM table WHERE department = "Tech" AND user_type = Any("Engineer", "Manager", "Designer")
```

The following code crashes:
```java

PostgresUtils
   .getClient()
   .preparedQuery("SELECT * FROM table WHERE department = $1 AND user_type = Any($2)")
   .execute(Tuple.of("Tech", types.toArray(String[]::new)))
```

Open to suggestions 🤔

PR #1267, opened 2023-01-05
MSSQL fix converting DB decimal type to java BigDecimal

dusik66:
See jdbc driver from Microsoft: https://github.com/microsoft/mssql-jdbc/blob/main/src/main/java/com/microsoft/sqlserver/jdbc/Util.java#L190

Motivation:

Some decimal numbers are wrongly converted into BigDecimal type. For example number 2147483648 comes  from DB as following byte stream: [5, 1, 0, 0, 0, -128]. First byte 5 is length, second 1 is sign (positive) and array [0, 0, 0, -128] is a value it self (reverse order). But constructor new BigInteger(bytes) expects two's-complement binary representation. But this is not our case. Two's-complement for our number is  [0, 0, 0, -128, 0]. That is the reason why there has to be used another constructor with the sign and magnitude. Same way as is used in the jdbc driver from microsoft.

Conformance:


dusik66:
ECA signed, test added.

FYI: @tsegismont

tsegismont (approved):
LGTM, thank you very much @dusik66 !

PR #1275, opened 2023-02-06
Exceptions improvements

### LINKED ISSUE #1269: NoStackTraceThrowable Exception is thrown when Postgres DB connectivity is lost
### Questions

Quarkus App - Rest Operation to fetch data from Postgres DB

So when the db connection is lost while the service is running and the service receives a request , the app is throwing NoStackTraceThrowable exception. This is a generic exception and is preventing from custom handling of db connectivity issue. Say if u want add DB connect exception to circuit breaker

### Version

Which version(s) did you encounter this bug ?

Quarkus version : 2.9.0.Final
Client Lib : quarkus-reactive-pg-client

### Context

When db connection is lost temporarily and the app receives a REST request , the app throws NoStackTraceThrowable exception

### Extra

* java 11 
* container image quarkus : registry.access.redhat.com/ubi-minimal:8.6

### LINKED ISSUE #1273: OracleException does not expose sql state code
### Questions

Why does the Oracle client remove the underlying SQLException? As part of PR #1081 I see the client is stripping down the SQLException and only using its message for purposes of simplification. But that removes helpful information. Often for SQL exception handling you need the sql state code. SQLException exposes that as a field. But because we strip down SQLException, we need to do a technology dependent string pattern matching to get the sql state code. PGException exposes sql state. So it would be nice if either OracleException would also, or at least retain the underlying SQLException.

### Version

4.2.2 - 4.3.7

### LINKED ISSUE #1274: Base exception class for all SQL Clients
Currently, each client comes with its own exception class for database access problems.

Projects that are built upon the core SQL Client API, like Hibernate Reactive, don't have access to details.

A base exception class can be added for all SQL Clients, which would give (when known):

- error code
- sqlState (XOPEN or SQL:2003 conventions)

IMPORTANT: this does not mean clients will now only report exceptions extending the base class. They could still report raw `VertxException` or `NoStackTraceThrowable`, when the problem does not come from a DB interaction.

tsegismont:
Closes #1274, fixes #1273, closes #1269

Commits can be reviewed individually

### DIFF
--- a/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLException.java
+++ b/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLException.java
@@ -11,13 +11,15 @@
 
 package io.vertx.mssqlclient;
 
+import io.vertx.sqlclient.DatabaseException;
+
 import java.util.ArrayList;
 import java.util.List;
 
 /**
- * A {@link RuntimeException} signals that an error occurred.
+ * The {@link DatabaseException} for MS SQL Server.
  */
-public class MSSQLException extends RuntimeException {
+public class MSSQLException extends DatabaseException {
 
   private final int number;
   private final byte state;
@@ -29,7 +31,7 @@ public class MSSQLException extends RuntimeException {
   private List<MSSQLException> additional;
 
   public MSSQLException(int number, byte state, byte severity, String message, String serverName, String procedureName, int lineNumber) {
-    super(null, null, false, false);
+    super(null, number, null);
     this.number = number;
     this.state = state;
     this.severity = severity;
--- a/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLException.java
+++ b/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLException.java
@@ -21,21 +21,19 @@ import java.util.List;
  */
 public class MSSQLException extends DatabaseException {
 
-  private final int number;
   private final byte state;
   private final byte severity;
-  private final String message;
+  private final String errorMessage;
   private final String serverName;
   private final String procedureName;
   private final int lineNumber;
   private List<MSSQLException> additional;
 
-  public MSSQLException(int number, byte state, byte severity, String message, String serverName, String procedureName, int lineNumber) {
-    super(null, number, null);
-    this.number = number;
+  public MSSQLException(int number, byte state, byte severity, String errorMessage, String serverName, String procedureName, int lineNumber) {
+    super(formatMessage(number, state, severity, errorMessage, serverName, procedureName, lineNumber), number, null);
     this.state = state;
     this.severity = severity;
-    this.message = message;
+    this.errorMessage = errorMessage;
     this.serverName = serverName;
     this.procedureName = procedureName;
     this.lineNumber = lineNumber;
@@ -48,8 +46,12 @@ public class MSSQLException extends DatabaseException {
     additional.add(e);
   }
 
+  /**
+   * @deprecated use {@link #getErrorCode()} instead
+   */
+  @Deprecated
   public int number() {
-    return number;
+    return getErrorCode();
   }
 
   public byte state() {
@@ -61,7 +63,7 @@ public class MSSQLException extends DatabaseException {
   }
 
   public String errorMessage() {
-    return message;
+    return errorMessage;
   }
 
   public String serverName() {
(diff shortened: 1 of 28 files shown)


tsegismont:
cc @DavideD @gavinking

vertx-mysql-client/src/main/java/io/vertx/mysqlclient/MySQLException.java
```diff
@@ -11,45 +11,14 @@
...
-
-  /**
-   * Get the error message in the error message sent from MySQL server.
-   *
-   * @return the error message
-   */
-  @Override
-  public String getMessage() {
-    return super.getMessage();
+    super(message, errorCode, sqlState);
```
vietj, line 22: we should format an error message like we do for `PgException ` and `DB2Exception`

vertx-oracle-client/src/main/java/io/vertx/oracleclient/OracleException.java
```diff
@@ -11,34 +11,21 @@
...
-
-  public OracleException(String message, boolean noStackTrace) {
-    super(message, noStackTrace);
-  }
+public class OracleException extends DatabaseException {
 
-  public OracleException(String message, Throwable cause, boolean noStackTrace) {
-    super(message, cause, noStackTrace);
+  public OracleException(String message, int errorCode, String sqlState) {
+    super(message, errorCode, sqlState);
```
vietj, line 24: we should format an error message like we do for `PgException ` and `DB2Exception`

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLException.java
```diff
@@ -29,7 +31,7 @@ public class MSSQLException extends RuntimeException {
   private List<MSSQLException> additional;
 
   public MSSQLException(int number, byte state, byte severity, String message, String serverName, String procedureName, int lineNumber) {
-    super(null, null, false, false);
+    super(null, number, null);
```
vietj, line 34: we should format an error message like we do for `PgException ` and `DB2Exception`

vertx-pg-client/src/main/java/io/vertx/pgclient/PgException.java
```diff
@@ -66,7 +67,7 @@ private static String formatMessage(String errorMessage, String severity, String
   private final String constraint;
 
   public PgException(String errorMessage, String severity, String code, String detail) {
-    super(formatMessage(errorMessage, severity, code));
+    super(formatMessage(errorMessage, severity, code), 0, code);
     this.errorMessage = errorMessage;
```
vietj, line 71: - `getCode()` should be deprecated in favour of `getErrorCode`()
- the `code` field should be removed and `getCode()` should return `getErrorCode()`

vietj (changes requested):
A couple of changes.

vertx-pg-client/src/main/java/io/vertx/pgclient/PgException.java
tsegismont: Done but, in fact, in Pg client `code` is the equivalent of `sqlState`, not `errorCode`.

vertx-oracle-client/src/main/java/io/vertx/oracleclient/OracleException.java
tsegismont: I've update other exception classes but not this one because the error messages coming from the Oracle driver already contain a lot of information, including error code and states

vertx-sql-client/src/main/java/io/vertx/sqlclient/ClosedChannelException.java
```diff
@@ -0,0 +1,25 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.sqlclient;
+
+import io.vertx.core.VertxException;
+
+/**
+ * When a client operation fails with this exception, the underlying connection may have been lost unexpectedly.
+ */
+public class ClosedChannelException extends VertxException {
```
vietj, line 18: Name it `ClosedConnectionException` like we do use for HTTP.

Channel is overloaded and can mean a socket but also something else.

vertx-sql-client/src/main/java/io/vertx/sqlclient/DatabaseException.java
```diff
@@ -0,0 +1,46 @@
...
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.sqlclient;
+
+import io.vertx.core.VertxException;
+
+/**
+ * Base class for SQL Client database access failures.
```
vietj, line 16: The doc is ambiguous and let think that it happens when we cannot access the database, instead here we want to state that this reports an exception reported by the database (hence the SQL state, etc...). _database failures_ would be fine I think.

vietj (dismissed):
Couple of minor changes.

PR #1296, opened 2023-03-30
Complete dynamic/rotated credentials support

kdubb:
Motivation:

Quarkus `CredentialsProvider` provides dynamic credentials from source like Vault. These credentials are leased and rotated regularly. This PR implements proper complete support for these type of credentials by implementing credential generation and regularly scheduled connection rotation to ensure fresh credentials are always used.

### `SqlCredentialsProvider`
A simple interface that allows the connection factories to request credentials (possibly dynamically generated) during the connection process. A `credentialsProvider` property has been added to `SqlConnectOptions` along with the standard `username` & `password`. The credentials provider supersedes any provided username & password.

All the current connection factories now only use a `SqlCredentialsProvider` to simplify the code. If `credentialsPrpvoder` is not set, one is created from the `username` & `password` options.

### Connection Maximum Lifetime

A `maxLifetime` option has been added to `PoolOptions` that determines the maximum amount of time a connection will stay connected; after this amount of time it will be closed.

`maxLifetime` is passed to `SqlConnectionPool` and each `PooledConnection` has an added `creationTimestamp`. Using these new properties the expiration check now also checks for lifetime maximum and evicts the connection for either of these connections.

### 

Conformance:


vietj:
can you make this two pull requests instead because we have two features ?

for the credential part, we should instead have the ability to create a pool with a Supplier<SqlConnectOptions> that would support this easily and more.

kdubb:
I will break it apart. I'm working on a `minSize` change to the pool as well I'll add as another PR.

> for the credential part, we should instead have the ability to create a pool with a Supplier<SqlConnectOptions> that would support this easily and more.

Is this not what it is doing currently? We need to pass a `Context` to the provider ensure any blocking implementations can be run on it.

PR #1297, opened 2023-03-30
Add `SqlCredentialsProvider` to source dynamic credentials when connecting

kdubb:
* Adds `SqlCredentialsProvider` interface that allows connection factories to request credentials during the connection process.
* Adds `credentialsProvider` options to `SqlConnectOptions`
* Static `username` and `password` options are translated to `SqlCredentialsProvider.Static` implementation that returns the same credentials consistently.

Motivation:

A simple interface that allows the connection factories to request credentials (possibly dynamically generated) during the connection process. A credentialsProvider property has been added to SqlConnectOptions along with the standard username & password. The credentials provider supersedes any provided username & password.

All the current connection factories now only use a SqlCredentialsProvider to simplify the code. If credentialsProvider is not set, one is created from the username & password options.

Conformance:

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlCredentialsProvider.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlCredentialsProvider.java
@@ -1,36 +1,12 @@
 package io.vertx.sqlclient;
 
+import io.vertx.codegen.annotations.VertxGen;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
 
+@VertxGen
 public interface SqlCredentialsProvider {
 
-  class Static implements SqlCredentialsProvider {
-
-    private final Credentials credentials;
-
-    public Static(String username, String password) {
-      this.credentials = new Credentials(username, password);
-    }
-
-    @Override
-    public Future<Credentials> getCredentials(Context context) {
-      return Future.succeededFuture(credentials);
-    }
-  }
-
-  class Credentials {
-
-    public String username;
-    public String password;
-
-    public Credentials(String username, String password) {
-      this.username = username;
-      this.password = password;
-    }
-
-  }
-
-  Future<Credentials> getCredentials(Context context);
+  Future<SqlCredentials> getCredentials(Context context);
 
 }
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlCredentialsProvider.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlCredentialsProvider.java
@@ -1,0 +1,36 @@
+package io.vertx.sqlclient;
+
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+
+public interface SqlCredentialsProvider {
+
+  class Static implements SqlCredentialsProvider {
+
+    private final Credentials credentials;
+
+    public Static(String username, String password) {
+      this.credentials = new Credentials(username, password);
+    }
+
+    @Override
+    public Future<Credentials> getCredentials(Context context) {
+      return Future.succeededFuture(credentials);
+    }
+  }
+
+  class Credentials {
+
+    public String username;
+    public String password;
+
+    public Credentials(String username, String password) {
+      this.username = username;
+      this.password = password;
+    }
+
+  }
+
+  Future<Credentials> getCredentials(Context context);
+
+}
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/ConnectionFactoryBase.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/ConnectionFactoryBase.java
@@ -22,6 +22,7 @@ import io.vertx.core.net.NetClientOptions;
 import io.vertx.core.net.SocketAddress;
 import io.vertx.core.net.impl.NetClientBuilder;
 import io.vertx.sqlclient.SqlConnectOptions;
+import io.vertx.sqlclient.SqlCredentialsProvider;
 import io.vertx.sqlclient.spi.ConnectionFactory;
 
 import java.util.Collections;
@@ -40,8 +41,7 @@ public abstract class ConnectionFactoryBase implements ConnectionFactory {
   protected final Map<String, String> properties;
   protected final SqlConnectOptions options;
   protected final SocketAddress server;
-  protected final String user;
-  protected final String password;
+  protected final SqlCredentialsProvider credentialsProvider;
   protected final String database;
 
   // cache
@@ -67,8 +67,11 @@ public abstract class ConnectionFactoryBase implements ConnectionFactory {
     this.properties = options.getProperties() == null ? null : Collections.unmodifiableMap(options.getProperties());
     this.server = options.getSocketAddress();
     this.options = options;
-    this.user = options.getUser();
-    this.password = options.getPassword();
+    if (options.getCredentialsProvider() != null) {
+      this.credentialsProvider = options.getCredentialsProvider();
+    } else {
+      this.credentialsProvider = new SqlCredentialsProvider.Static(options.getUser(), options.getPassword());
+    }
     this.database = options.getDatabase();
 
     this.cachePreparedStatements = options.getCachePreparedStatements();
@@ -96,7 +99,7 @@ public abstract class ConnectionFactoryBase implements ConnectionFactory {
 
   public Future<Connection> connect(EventLoopContext context) {
     PromiseInternal<Connection> promise = context.promise();
-    context.emit(promise, p -> doConnectWithRetry(server, user, password, database, p, reconnectAttempts));
+    context.emit(promise, p -> doConnectWithRetry(server, credentialsProvider, database, p, reconnectAttempts));
     return promise.future();
   }
 
@@ -105,15 +108,15 @@ public abstract class ConnectionFactoryBase implements ConnectionFactory {
     clientCloseFuture.close(promise);
   }
 
-  private void doConnectWithRetry(SocketAddress server, String username, String password, String database, PromiseInternal<Connection> promise, int remainingAttempts) {
+  private void doConnectWithRetry(SocketAddress server, SqlCredentialsProvider credentialsProvider, String database, PromiseInternal<Connection> promise, int remainingAttempts) {
     EventLoopContext ctx = (EventLoopContext) promise.context();
-    doConnectInternal(server, username, password, database, ctx).onComplete(ar -> {
+    doConnectInternal(server, credentialsProvider, database, ctx).onComplete(ar -> {
       if (ar.succeeded()) {
         promise.complete(ar.result());
       } else {
         if (remainingAttempts > 0) {
           ctx.owner().setTimer(reconnectInterval, id -> {
-            doConnectWithRetry(server, username, password, database, promise, remainingAttempts - 1);
+            doConnectWithRetry(server, credentialsProvider, database, promise, remainingAttempts - 1);
           });
         } else {
           promise.fail(ar.cause());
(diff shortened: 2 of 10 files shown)


vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlCredentialsProvider.java
```diff
@@ -0,0 +1,36 @@
+package io.vertx.sqlclient;
+
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+
+public interface SqlCredentialsProvider {
```
tsegismont, line 6: From my perspective, it would be better to have `SqlCredentialsProvider` annotated with `@VertxGen`, so that Mutiny / Rx / Kotlin users have something that integrates well with the rest of their code base.

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlCredentialsProvider.java
```diff
@@ -0,0 +1,36 @@
...
+
+  class Static implements SqlCredentialsProvider {
```
tsegismont, line 8: This is an implementation detail that shall be moved to the `impl` package.

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlCredentialsProvider.java
```diff
@@ -0,0 +1,36 @@
...
+      this.credentials = new Credentials(username, password);
+    }
+
+    @Override
+    public Future<Credentials> getCredentials(Context context) {
+      return Future.succeededFuture(credentials);
+    }
+  }
+
+  class Credentials {
```
tsegismont, line 22: For a working integration with codegen, this should be annotated with `@DataObject` and moved to its own file.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/ConnectionFactoryBase.java
```diff
@@ -40,8 +41,7 @@ public abstract class ConnectionFactoryBase implements ConnectionFactory {
   protected final Map<String, String> properties;
   protected final SqlConnectOptions options;
   protected final SocketAddress server;
-  protected final String user;
-  protected final String password;
+  protected final SqlCredentialsProvider credentialsProvider;
```
tsegismont, line 44: It don't believe we should have this in the options class.

If you create a connection manually, the credentials provider is redundant: you could get the latest credentials before creating the connect options.

So the crendentials provider really makes sense for pools. Then it should be set the same way as the connection provider, i.e:

```java
  @Fluent
  Pool connectionProvider(Function<Context, Future<SqlConnection>> provider);
```

This way when RxJava / Mutiny users customize their pool , will get a method that works well with their codebase.

tsegismont (commented):
Thank you @kdubb , this is an interesting proposal.

Thoughts @vietj ?

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlCredentialsProvider.java
kdubb: Will do.

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlCredentialsProvider.java
kdubb: Ok.

vertx-sql-client/src/main/java/io/vertx/sqlclient/SqlCredentialsProvider.java
kdubb: Ok.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/ConnectionFactoryBase.java
kdubb: It's directly affecting the SqlConnectionOptions and is an alternative to the username/password options. I believe this is the best place for it.

> If you create a connection manually, the credentials provider is redundant: you could get the latest credentials before creating the connect options.

Yes you could. It's also pretty convenient for people using external credentials systems, that might also open a direct connection (e.g. for migration or notifications), to share a single configured `SqlConnectOptions`. Additionally, this colocates all credentials related parameters into a single options object.

> This way when RxJava / Mutiny users customize their pool , will get a method that works well with their codebase.
This already works quite well with those codebases. I've had this wired all the way through in Quarkus since I opened the PR and it's tidy and convenient. The only downside is, as you pointed out, there is currently no codegen support; but that will also requires a bump to the related smallrye mutiny library.

vietj:
I think instead we should have a way to create a pool with a `Supplier<ConnectOptions>` that would solve this use case and make other use case possible.

tsegismont:
> I think instead we should have a way to create a pool with a `Supplier<ConnectOptions>` that would solve this use case and make other use case possible.

We already have an API method to provide a list of connection options (which is used in the project to provided a round-robin load balancer).
We also have a connection provider which can create a connection dynamically.

What kind of use case a pool with a `Supplier<ConnectOptions>` would solve that's not solved already by either of the above possibilities?

vietj:
@tsegismont 

- `PgPool.pool(Arrays.asList(server1, server2, server3), options)` does not solve the same issue
- connectionProvider is really the ultimate way to operate on a pool
- it avoid to introduce a new interface that is more or less like the `Supplier<ConnectOptions>`

so it would work like :

PgPool pool = PgPool.pool(() -> new PgConnectOptions());

kdubb:
@vietj The problem with `Supplier<ConnectOptions>` is that it doesn't provide a `Context`. In practice, specifically the Quarkus implementation, requires the ability to execute them on context due to blocking issues. While you could inject a context into the `Supplier` when it's created, it seems much more flexible and proper to provide the context to the `getCredentials` method like `SqlCredentialsProvider` allows.

kdubb:
Yes. In Quarkus the only current credentials provider is for Vault (which I also maintain). During the credentials request it will occasionally, not always, need to reach out to the Vault instance to refresh credentials. Currently this is a blocking call and in my test implementation it's using `executeBlocking`.

vietj:
then maybe we should also call the `Supplier<ConnectOptions>` on a worker thread if that's configured for this

kdubb:
IIRC, everything that `ConnectionFactoryBase` pulls in. It's a shared factory so you cannot simply update the options and expect it to apply to a single connection.

kdubb:
One of the changes this PR includes is to remove the username/password from `ConnectionFactoryBase` and replace it with a `SqlCredentialsProvider` specifically for this reason of it being shared.

FYI, in the case that no `SqlCredentialsProvider` is set in the options, a `StaticSqlCredentialsProvider` is created with the username and password from the options and that is provided to `ConnectionFactoryBase`.

vietj:
I see we would need indeed some decoupling of the connection factory
from the connect options, so the connection factory manages the
NetClient so with a single connection factory we can create a
connection for a given connect options without much hassle.

vietj:
I think we should assume that the first time the supplier is called
then the NetClientOptions will be used and then only the properties
defines in SqlConnectOptions will be used on subsequent calls.

Maybe in Vert.x 5 we should likely

1/ remove the inheritance of NetClientOptions by SqlConnectOptions
2/ have a TCP ConnectOptions, so that we can have SqlConnectOptions
extends the TCP ConnectOptions to have more flexiblity

kdubb:
I'd vote for the removal of the inheritance. As I referenced before, passing a bunch of options to a method that ignores most of them is not a great DX experience and will rely on it being well documented.

That being said, I still think having this PR is a nice enhancement; `SqlCredentialsProvider` is simple and easy to understand. Adding the ability to customize the connect options at a later time and keeping `SqlCredentialsProvider` doesn't really have a downside and just provides more choice for customization to consumers.

kdubb:
FYI, the PR is updated with @tsegismont's feedback and the tests should be fixed.

vietj:
the point is that I don't want to introduce a new API
(SqlCredentialsProvider) that would be then removed and be a breaking
change because using SqlConnectOptions is the way to go.

kdubb:
I understand. I'm suggesting having both is not a downside and just offers more customization points. A consumer could then choose to implement `SqlCredentialsProvider` or the pool based `Supplier<SqlConnectOptions>`.

kdubb:
As I said previously, implementing a `SqlCredentialsProvider` can be advantageous even for non pooled connections. For example, in Quarkus, it would produce a single CDI based `SqlConnectOptions` bean with `SqlCredentialsProvider` set against it. This options bean can then be used for non-pooled and pooled connections alike. We do this in our project because before the pool is started a one-off connection is requested to configure the database.

vietj:
I believe that SqlCredentialsProvider API then could be part of
Quarkus API and it would be translated to SqlConnectOptions instead
the way its intended to work with CDI

kdubb:
I understand you want to support dynamically providing `SqlConnectOptions` but what's the downside to having both?

If we only support providing dynamic `SqlConnectOptions` then we only have the two choices provided and the preferable one of those choice won't be available until 5.0. In either of those cases we'll need to make some serious changes. Adding an `SqlCredentialsProvider` now and back porting, would allow a clean upgrade path from 4 -> 5; and as I've suggested supporting both on into the future isn't a problem (as I see it).

Given that Quarkus is on 4.3.7 (for 2.16.x) and 4.4 (for upcoming 3.0.x) and the fact that my motivation here is to fix some longstanding issues in Quarkus a clean path forward would be great.

kdubb:
Just to be clean I'm advocation for adding `SqlCredentialsProvider` now (aka this PR), and in the future adding your suggested `Supplier<SqlConnectOptions>` and then keeping/maintaining both going forward. It's a clear separation of concerns and consumers can choose which they want to implement.

vietj:
unfortunately we don't agree

kdubb:
It seems we are disagreeing on the idea of supporting a possible future use case (supporting dynamic SqlConnectOptions) verse a concrete need that exists today (dynamic credentials).

If the solution was an easy change I'd agree but I looked into it today and it's going to be a big breaking change change to get dynamically supplied `SqlConnectOptions` to work. I think the amount of change required for this feature requires a concrete use case. The biggest use case I could come up with, connecting to multiple databases, is already supported via passing multiple `SqlConnectOptions` to the pool constructor which are then wrapped in the the round robin selector.

Is there a specific use case you are thinking of that your suggested solution solves over this PR?

vietj:
actually we are disagreeing on several points

About the API, I don't see the point of adding new types to the
existing API (basically adding a Function<Context, String[]>) that we
will need to stick with, versus having on the a more universal way to
implement this use case with my proposal which does not introduce new
types. If we can do more then we can do less.

In addition I would like to revisit the load balancing feature that
currently exist to use this as well, e.g instead of
Arrays.asList(Server1, ... ServerN), specify (N, Supplier<Options>)
which would be more dynamic.

Finally, the current implementation of the pool requires to open N TCP
clients which makes it more complicated than it should for resource
management, so I would like to have a single TCP client used for the
many servers, so when a pool is closed, we only need to close this
single client and we don't need to keep track of all the clients which
have been created. One reason we used a fixed List of server and not
something dynamic is that we would have had to keep track of every TCP
client created (easy) but also check when this client would not be
used anymore by one of those servers which would have been quite
complicated.

Implementing this of course is more complex than your PR, yet it
remains a refactoring of the existing code.

kdubb:
Ok. So would a compromise be to replace `SqlCredentialsFactory` and `Function<Context, String[]>` in this PR?

We would only be adding the `SqlConnectOptions.credentialsProvider` method to the API at that point. It would not make it any more/less complicated to revisit your other concerns and wouldn't introduce any new types.

kdubb:
Obviously that would also mean the `SqlCredentials` and `StaticSqlCredentialsProvider` would go away as well.

kdubb:
Actually, the replacement would need to be `Function<Context, Future<String[]>>`.

vietj:
I created this (draft) PR to support the supplier based approach in 4.x : https://github.com/eclipse-vertx/vertx-sql-client/pull/1305

PR #1298, opened 2023-03-30
Support connection maximum lifetime to recycle connections regularly

kdubb:
Adds `maxLifetime` to `PoolOptions` and alters `PooledConnection` to honor the maximum lifetime as well as the idle timeout.

Motivation:

A `maxLifetime` option has been added to PoolOptions that determines the maximum amount of time a connection will stay connected; after this amount of time it will be closed at the next eviction check.

The "expiration" naming was changed to "eviction" to represent the more general check that is performed. The idle timeout refreshing was also change to start immediately after connection and update accordingly to make the idle timeout more predictable and match the max life time check.

Conformance:

### DIFF
--- a/vertx-pg-client/src/test/java/io/vertx/pgclient/PgPoolTest.java
+++ b/vertx-pg-client/src/test/java/io/vertx/pgclient/PgPoolTest.java
@@ -384,7 +384,7 @@ public class PgPoolTest extends PgPoolTestBase {
   public void testPoolIdleTimeout(TestContext ctx) {
     ProxyServer proxy = ProxyServer.create(vertx, options.getPort(), options.getHost());
     AtomicReference<ProxyServer.Connection> proxyConn = new AtomicReference<>();
-    int pooleCleanerPeriod = 100;
+    int poolCleanerPeriod = 100;
     int idleTimeout = 3000;
     Async latch = ctx.async();
     proxy.proxyHandler(conn -> {
@@ -393,8 +393,8 @@ public class PgPoolTest extends PgPoolTestBase {
       conn.clientCloseHandler(v -> {
         long lifetime = System.currentTimeMillis() - now;
         int delta = 500;
-        int lowerBound = idleTimeout - pooleCleanerPeriod - delta;
-        int upperBound = idleTimeout + pooleCleanerPeriod + delta;
+        int lowerBound = idleTimeout - poolCleanerPeriod - delta;
+        int upperBound = idleTimeout + poolCleanerPeriod + delta;
         ctx.assertTrue(lifetime >= lowerBound, "Was expecting connection to be closed in more than " + lowerBound + ": " + lifetime);
         ctx.assertTrue(lifetime <= upperBound, "Was expecting connection to be closed in less than " + upperBound + ": "+ lifetime);
         latch.complete();
@@ -408,7 +408,7 @@ public class PgPoolTest extends PgPoolTestBase {
     listenLatch.awaitSuccess(20_000);
 
     poolOptions
-      .setPoolCleanerPeriod(pooleCleanerPeriod)
+      .setPoolCleanerPeriod(poolCleanerPeriod)
       .setMaxLifetime(0)
       .setIdleTimeout(idleTimeout)
       .setIdleTimeoutUnit(TimeUnit.MILLISECONDS);
@@ -427,7 +427,7 @@ public class PgPoolTest extends PgPoolTestBase {
   public void testPoolMaxLifetime(TestContext ctx) {
     ProxyServer proxy = ProxyServer.create(vertx, options.getPort(), options.getHost());
     AtomicReference<ProxyServer.Connection> proxyConn = new AtomicReference<>();
-    int pooleCleanerPeriod = 100;
+    int poolCleanerPeriod = 100;
     int maxLifetime = 3000;
     Async latch = ctx.async();
     proxy.proxyHandler(conn -> {
@@ -436,8 +436,8 @@ public class PgPoolTest extends PgPoolTestBase {
       conn.clientCloseHandler(v -> {
         long lifetime = System.currentTimeMillis() - now;
         int delta = 500;
-        int lowerBound = maxLifetime - pooleCleanerPeriod - delta;
-        int upperBound = maxLifetime + pooleCleanerPeriod + delta;
+        int lowerBound = maxLifetime - poolCleanerPeriod - delta;
+        int upperBound = maxLifetime + poolCleanerPeriod + delta;
         ctx.assertTrue(lifetime >= lowerBound, "Was expecting connection to be closed in more than " + lowerBound + ": " + lifetime);
         ctx.assertTrue(lifetime <= upperBound, "Was expecting connection to be closed in less than " + upperBound + ": "+ lifetime);
         latch.complete();
@@ -451,7 +451,7 @@ public class PgPoolTest extends PgPoolTestBase {
     listenLatch.awaitSuccess(20_000);
 
     poolOptions
-      .setPoolCleanerPeriod(pooleCleanerPeriod)
+      .setPoolCleanerPeriod(poolCleanerPeriod)
       .setIdleTimeout(0)
       .setMaxLifetime(maxLifetime)
       .setMaxLifetimeUnit(TimeUnit.MILLISECONDS);
@@ -507,9 +507,9 @@ public class PgPoolTest extends PgPoolTestBase {
   public void testNoConnectionLeaks(TestContext ctx) {
     Async killConnections = ctx.async();
     PgConnection.connect(vertx, options).onComplete(ctx.asyncAssertSuccess(conn -> {
-      Collector<Row, ?, List<Integer>> collector = mapping(row -> row.getInteger(0), toList());
+      Collector<Row, ?, List<Boolean>> collector = mapping(row -> row.getBoolean(0), toList());
       String sql = "SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE pid <> pg_backend_pid() AND datname = $1";
-      PreparedQuery<SqlResult<List<Integer>>> preparedQuery = conn.preparedQuery(sql).collecting(collector);
+      PreparedQuery<SqlResult<List<Boolean>>> preparedQuery = conn.preparedQuery(sql).collecting(collector);
       Tuple params = Tuple.of(options.getDatabase());
       preparedQuery.execute(params).compose(cf -> conn.close()).onComplete(ctx.asyncAssertSuccess(v -> killConnections.complete()));
     }));


kdubb:
There is a follow on PR that adds a `min-size` attribute to ensure a minimum # of connections are always available. Will create after this is merged.

kdubb:
@vietj Fixed `testNoConnectionLeaks` test. It was expecting `pg_terminate_backend` to return an integer but it returns a boolean.

I also added a test for `maxLifetime` similar to the test for `idleTimeout` which I also altered to explicitly ensure `maxLifetime` is not participating.

kdubb:
@vietj Can we get this PR moving forward? I'm assuming this is easier than #1297 and is more important to fixing the Quarkus extensions.

vertx-sql-client/src/main/java/io/vertx/sqlclient/PoolOptions.java
```diff
@@ -179,14 +190,50 @@ public int getIdleTimeout() {
...
+
+  /**
+   * @return pooled connection max lifetime
+   */
+  public int getMaxLifetime() {
+    return maxLifetime;
+  }
+
+  /**
+   * Establish a max lifetime for pooled connections.
```
vietj, line 227: Add javadoc for 0 => no max lifetime

vertx-sql-client/src/main/java/io/vertx/sqlclient/PoolOptions.java
```diff
@@ -179,14 +190,50 @@ public int getIdleTimeout() {
...
+   *
+   * @param maxLifetime the pool connection max lifetime
+   * @return a reference to this, so the API can be used fluently
+   */
+  public PoolOptions setMaxLifetime(int maxLifetime) {
+    this.maxLifetime = maxLifetime;
```
vietj, line 233: Fail if maxLifetime < 0

vietj (approved):
A couple of very minor changes, well done.

Please provide a PR to backport to 4.x branch

vertx-sql-client/src/main/java/io/vertx/sqlclient/PoolOptions.java
kdubb: Done. Did `idleTimeout` as well.

kdubb:
Requested changes are completed. Should be good to go.

vertx-sql-client/src/main/java/io/vertx/sqlclient/PoolOptions.java
```diff
@@ -48,11 +47,21 @@ public class PoolOptions {
    */
   public static final int DEFAULT_IDLE_TIMEOUT = 0;
 
+  /**
+   * Default maximum connection lifetime in the pool = 0 (no lifetime)
```
tsegismont, line 51:
```suggestion
   * Default maximum pooled connection lifetime = 0 (no maximum)
```

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/SqlConnectionPool.java
```diff
@@ -394,5 +404,18 @@ public int getSecretKey() {
...
+      return idleEvictionTimestamp < now;
+    }
+
+    public boolean hasLifetimeExpired(long now) {
+      return lifetimeEvictionTimestamp < now;
+    }
+
+    public boolean shouldEvict(long now) {
+      return hasIdleExpired(now) || hasLifetimeExpired(now);
+    }
```
tsegismont, line 418:
```suggestion
    private boolean hasIdleExpired(long now) {
      return idleEvictionTimestamp < now;
    }

    private boolean hasLifetimeExpired(long now) {
      return lifetimeEvictionTimestamp < now;
    }

    private boolean shouldEvict(long now) {
      return hasIdleExpired(now) || hasLifetimeExpired(now);
    }
```

Making these private will make the JIT's job easier (good candidates for inlining).

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/SqlConnectionPool.java
```diff
@@ -305,6 +311,10 @@ private void close(Promise<Void> promise) {
       conn.close(this, promise);
     }
 
+    public void refresh() {
```
tsegismont, line 314:
```suggestion
    private void refresh() {
```

vertx-pg-client/src/test/java/io/vertx/pgclient/PgPoolTest.java
```diff
@@ -422,6 +423,49 @@ public void testPoolIdleTimeout(TestContext ctx) {
       .onComplete(ctx.asyncAssertSuccess());
   }
 
+  @Test
+  public void testPoolMaxLifetime(TestContext ctx) {
+    ProxyServer proxy = ProxyServer.create(vertx, options.getPort(), options.getHost());
+    AtomicReference<ProxyServer.Connection> proxyConn = new AtomicReference<>();
+    int pooleCleanerPeriod = 100;
```
tsegismont, line 430: How about taking the opportunity to rename this to `poolCleanerPeriod` in this test and others?

tsegismont (approved):
Thank you @kdubb, LGTM

A few minor comments and proposed changes.

vertx-sql-client/src/main/java/io/vertx/sqlclient/PoolOptions.java
kdubb: done

vertx-sql-client/src/main/java/io/vertx/sqlclient/PoolOptions.java
```diff
@@ -48,11 +47,21 @@ public class PoolOptions {
...
+   */
+  public static final int DEFAULT_MAXIMUM_LIFETIME = 0;
+
   /**
    * Default connection idle time unit in the pool = seconds
    */
   public static final TimeUnit DEFAULT_IDLE_TIMEOUT_TIME_UNIT = TimeUnit.SECONDS;
 
+  /**
+   * Default maximum connection lifetime time unit in the pool = seconds
```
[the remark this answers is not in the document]
kdubb: done

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/SqlConnectionPool.java
kdubb: done

vertx-pg-client/src/test/java/io/vertx/pgclient/PgPoolTest.java
kdubb: done

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/SqlConnectionPool.java
kdubb: done

kdubb:
Updated with requested changes and squashed into two distinct commits. The second one is about fixing tests unrelated to the feature, so kept it as separate commit.

tsegismont:
@kdubb can you please rebase?

tsegismont:
@kdubb have you found any issue while rebasing or haven't found time to do it?

kdubb:
Apologies I was out for the last week. Thanks for merging!

tsegismont:
No worries

chandramouli-r:
(FYI both my ideas are borrowed from [here](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L168))

kdubb:
@chandramouli-r Hikari is a great connection pool for JDBC but not everything translates to the reactive pool (which I think can be much simpler).

WRT "jitter", I really don't see the need for it given that each connection is only recycled after it's done being used. From observing this PR actually in use, even a set of connections that starts off in synchronized recycling (because they were started by a `min-size` parameter) becomes out of sync fairly quickly because of different query times used on each connection.

Flagging "in use" is likewise not entirely needed. I looked at this quite a bit during implementation. First, much of the "standard" usage acquires and releases a connection from the pool without explicitly obtaining a connection and releasing it (by client code). Second, even in the case of client code that does explicitly acquire connections they are for short lived queries and (if properly written) will close the connection soon after.

This boils down to, connections are constantly getting a chance to be closed within a reasonable amount of time and the "jitter" is basically built in due to the difference in queries. I would let this sit and see if there is anything more that is actually needed before investing time time in adding features that my not be explicitly needed.

kdubb:
As far as `min-size` I'll be opening that PR today.

kdubb:
Yes. My point was though that after about a couple hours of running the application the evictions are happening seconds apart and they only get further as time goes on.

kdubb:
That was with a 30 minute eviction time.

chandramouli-r:
@kdubb Thanks for working on the `min-size` PR.
I'm fine with using the 2 changes: `maxLifeTime + min-size` and checking (at least for my workload) if there are connections that last well beyond their lifetime. 

I agree that jitter may be naturally added since some connections will be idle during the eviction check, and some might be in use, thus delaying their eviction.

The reason I proposed the second idea (adding an "evicted" flag) is to handle the case where a busy pool is always using some connections in steady state, and the cleaner may be racing with the connection usage to evict it. I can do some bookkeeping in my application to check how long connections last beyond their lifetime, and based on that, maybe it could be worth adding an "evicted" flag.

kdubb:
The max lifetime is a hard check after every return to the pool. The only time a connection will last well beyond its lifetime is if it is acquired and held for a period of time well beyond `maxLifetime`.

I've used PostgreSQL extensively and one use case like this that exists for acquiring extremely long lived connections is notifications. Applications that use notifications tend to acquire a single connection and keep it open as long as possible (i.e. forever). An "in use" flag would do nothing in this case as it would just be marked "in use" and without a forceful closure it would stay marked.

I really don't know of another reason an application would not be regularly returning connections back to the pool after query or transaction; neither of which should come any where close to a `maxLifetime`.

chandramouli-r:
I think this is where my confusion lies:

> The max lifetime is a hard check after every return to the pool
I didn't see a check for `hasLifetimeExpired` on the "return to the pool path" in the PR (it is only invoked during `evict()` which is called by the Cleaner). 

(Maybe I'm just looking at the wrong place).

kdubb:
Actually, this might be part of the `minSize` changes. I'll be looking tonight.

tsegismont:
I agree with @chandramouli-r .

We could have a predicate in the Recycle implementation. For SqlConnection pool, it would consist in verifying maxLifetimeTimestamp is in the future. If it's not, the connection should be removed from the pool instead of being recycled.

> Actually, this might be part of the `minSize` changes. I'll be looking tonight.

@kdubb please create separate PRs if possible.

chandramouli-r:
Hi @kdubb - just wanted to check if you got time to look at the `minSize` changes (I can help out if you have a working branch). Thanks.

kdubb:
@chandramouli-r `minSize` for 5.x is in PR #1319.

The "hard check" I referred to had to be removed. Support is needed in the `ConnectionPool` and `Lease` in vertx-core

kdubb:
@chandramouli-r I've opened [#4680](https://github.com/eclipse-vertx/vert.x/issues/4680) in the core repo to request the needed change with this (or something similar) it's an easy `hasLifetimeExpired` check in the `PooledConnection.cleanup` method to ensure `maxLifetime` is honored.

PR #1300, opened 2023-04-01
Add support for level 7 proxies in Postgres

vietj:
Level 7 proxies like PgBouncer can load balance queries over several database connections. When doing so prepared statement cannot be cached by the client and special care must be taken to use extended queries. For now we support PgBouncer proxy.


vietj:
Need to check if we can test it with Yandex Odyssey

PR #1306, opened 2023-04-09
Connect options supplier

vietj:
A connection pool should be configurable with a `Supplier<Future<SqlConnectOptions>>` that is called whenever the pool needs to create a connection.

- It enables dynamic pool configuration instead of using the connection provider
- Round robin fixed options list can be reimplemented with the feature

### DIFF
--- a/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgConnectionImpl.java
+++ b/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgConnectionImpl.java
@@ -133,7 +133,8 @@ public class PgConnectionImpl extends SqlConnectionBase<PgConnectionImpl> implem
   public PgConnection cancelRequest(Handler<AsyncResult<Void>> handler) {
     Context current = Vertx.currentContext();
     if (current == context) {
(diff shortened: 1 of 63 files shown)


vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgConnectionImpl.java
```diff
@@ -34,30 +34,27 @@
...
-    } else {
-      PgConnectionFactory client;
-      try {
-        client = new PgConnectionFactory(context.owner(), options);
-      } catch (Exception e) {
-        return context.failedFuture(e);
-      }
-      context.addCloseHook(client);
-      return (Future) client.connect(context);
+  public static Future<PgConnection> connect(ContextInternal context, Supplier<PgConnectOptions> options) {
```
tsegismont, line 42: Is there a reason for changing the type of the `options` parameter to `Supplier`?

This method is only used for one-shot connection, and the only caller creates a synthetic supplier

vietj: it might be a left over of previous architecture, if you can simplify it go ahead

PR #1309, opened 2023-04-13
Close one-shot factories when connection is closed

### LINKED ISSUE #1302: Direct usage of PgConnection results in memory leak
### Version

Which version(s) did you encounter this bug ?
quarkus 2.16.5 which uses vertx 4.3.7 and vertx-pg-client 2.30.1.

### Context

Using PgConnection directly, there seems to be a memory leak. PgConnectionFactory stays in memory after close.

### Do you have a reproducer?

https://github.com/robp94/vertx-pg-memory
Relevant code is in `org.acme.GreetingResource`.
See README for Instructions how to run. Use scripts to run some requests and make a memory dump.

Look for 
image

### Steps to reproduce

1. Open connection
2. Use it
3. Close it
4. GC cannot clean up factory and some other objects. Connections on the other hand get cleaned up.
```
PgConnectOptions connectOptions = new PgConnectOptions().setPort(port).setHost(host).setDatabase(databaseName).setUser(username).setPassword(password);
PgConnection.connect(vertx, connectOptions);
```
Workaround:
```
ContextInternal context = ContextInternal.current();
context.removeCloseHook(((PgConnectionImpl) c.getDelegate()).factory());
Promise<Void> promise = Promise.promise();
((PgConnectionImpl) c.getDelegate()).factory().close(promise);
```

### Extra

See https://groups.google.com/g/vertx/c/ddQek1aGXmE

* ubuntu 22.04, kubernetes
* JDK 17

tsegismont:
Fixes #1302

A close future is created for connection factories and another one for connections. In one-shot connection factories, the factory close future is added as a hook to the connection close future. And then the factory close future is added as context close hook, instead of the factory itself.

The combination of these changes guarantees that:

- the factory is always closed properly (and thus the underlying net client)
- factory objects aren't accumulated as close hooks in context

Motivation:

### DIFF
--- a/vertx-db2-client/src/main/java/io/vertx/db2client/impl/DB2ConnectionFactory.java
+++ b/vertx-db2-client/src/main/java/io/vertx/db2client/impl/DB2ConnectionFactory.java
@@ -18,6 +18,7 @@ package io.vertx.db2client.impl;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
+import io.vertx.core.impl.CloseFuture;
 import io.vertx.core.impl.ContextInternal;
 import io.vertx.core.impl.EventLoopContext;
 import io.vertx.core.impl.VertxInternal;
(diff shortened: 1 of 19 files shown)


vietj:
can you first port this to 4.4 and then we reconsider the patch ? I'm currently reworking the java finalization in master and I think it will be better first to do this.

vietj:
FTR https://github.com/eclipse-vertx/vert.x/wiki/RFC:-Replace-finalization-by-cleaners

tsegismont:
Closing this one as the solution based on close futures is not valid (connection closed when closefuture is gc-ed)

PR #1313, opened 2023-04-18
Suspend SqlConnectionPool during CRaC snapshotting

rvansa:
This change intends to support an application using Vertx GRPC server to perform the Checkpoint and Restore on JVM implementing this, specifically using [OpenJDK CRaC](https://github.com/openjdk/crac/tree/crac) or future versions of OpenJDK. Package org.crac is a facade that either forwards the invocation to actual implementation or provides a no-op implementation.


rvansa:
This is a draft PR until https://github.com/eclipse-vertx/vert.x/pull/4673 gets integrated.

PR #1314, opened 2023-04-18
Partial support of SSVARIANTTYPE

### LINKED ISSUE #1312: [MSSQL] Unable to decode typeInfo for SSVARIANT
A simple select on the information schema:
```
 select * from INFORMATION_SCHEMA.SEQUENCES
```

Causes the following exception:
```
Caused by: java.lang.UnsupportedOperationException: Unable to decode typeInfo for SSVARIANT
	at io.vertx.mssqlclient.impl.codec.DataType.decodeTypeInfo(DataType.java:1042) ~[vertx-mssql-client-4.4.1.jar:4.4.1]
	at io.vertx.mssqlclient.impl.codec.MSSQLCommandCodec.handleColumnMetadata(MSSQLCommandCodec.java:120) ~[vertx-mssql-client-4.4.1.jar:4.4.1]
	at io.vertx.mssqlclient.impl.codec.MSSQLCommandCodec.decode(MSSQLCommandCodec.java:51) ~[vertx-mssql-client-4.4.1.jar:4.4.1]
	at io.vertx.mssqlclient.impl.codec.TdsMessageDecoder.decodeMessage(TdsMessageDecoder.java:65) ~[vertx-mssql-client-4.4.1.jar:4.4.1]
	at io.vertx.mssqlclient.impl.codec.TdsMessageDecoder.channelRead(TdsMessageDecoder.java:43) ~[vertx-mssql-client-4.4.1.jar:4.4.1]
	at io.netty.channel.CombinedChannelDuplexHandler.channelRead(CombinedChannelDuplexHandler.java:251) ~[netty-transport-4.1.90.Final.jar:4.1.90.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:442) ~[netty-transport-4.1.90.Final.jar:4.1.90.Final]
```

This is the only element we create in the database (although, it doesn't seem to matter):
```
create table BASIC_TYPES_TABLE (unmapped_column int)
```

Tested with MSSQL 2022-latest and Vert.x 4.3.8 and 4.4.1

tsegismont:
Fixes #1312

Partial implementation of SSVARIANTTYPE decoding, just enough to support querying metadata about integer MSSQL sequences.


DavideD:
Sure, I will check in a bit

DavideD:
Works like a charm! Thanks a lot!
By the way, I've noticed that `MSSQLRowImpl#getString(String)` is key sensitive. It fails for us because we pass the parameter lowercased instead of uppercased.

Is this by design?

tsegismont:
> Works like a charm! Thanks a lot! 

Awesome, thanks

>By the way, I've noticed that `MSSQLRowImpl#getString(String)` is key sensitive. It fails for us because we pass the parameter lowercased instead of uppercased.
> 
> Is this by design?

It's not a regression. The SQL Clients work with the column name as reported by database.

PR #1338, opened 2023-07-11
Honor pool connection timeout when executing queries directly in the pool

### LINKED ISSUE #1232: PgPool queries seems to disregard setConnectionTimeout setting
### Version
4.3.1

### Context
While submitting queries using connections manually retrieved from a pool, PoolOptions'a setConnectionTimeout() seems to take an effect as expected. For example, using a pool with 1 connection a flow that does
`pool.getConnection()` -> doesn't close it and attempts to do another `pool.getConnection()` will end up with a timeout exception.

This doesn't seem to be the case when `pool.query("...").execute()` is used. Once the first connection is taken and the pool is exhausted the second attempt is simply stuck forever. 

### Do you have a reproducer?
Gist:
https://gist.github.com/ashertarno/2bf88dccc13667a93dfcc560e0974a9a

### Steps to reproduce
The gist above provides two meethods:  `connectionTimeOut()` that mimics the working flow with two on demand retrieved connections and `poolTimeOut()` with the first connection being retrieved manually to exhaust the pool and an attempt to call 
pool.query().execute(). You can call each of these methods by editing main().

pablosaavedra-rappi:
Should fix #1232, as it now uses the timeout when acquiring the connection

Motivation:

We've been bugged by #1232 for some time and the fix keeps getting pushed.

### DIFF
--- a/vertx-sql-client/src/main/generated/io/vertx/sqlclient/PoolOptionsConverter.java
+++ b/vertx-sql-client/src/main/generated/io/vertx/sqlclient/PoolOptionsConverter.java
@@ -20,6 +20,11 @@ public class PoolOptionsConverter {
    static void fromJson(Iterable<java.util.Map.Entry<String, Object>> json, PoolOptions obj) {
     for (java.util.Map.Entry<String, Object> member : json) {
       switch (member.getKey()) {
+        case "alwaysUseTimeout":
+          if (member.getValue() instanceof Boolean) {
+            obj.setAlwaysUseTimeout((Boolean)member.getValue());
+          }
+          break;
         case "connectionTimeout":
           if (member.getValue() instanceof Number) {
             obj.setConnectionTimeout(((Number)member.getValue()).intValue());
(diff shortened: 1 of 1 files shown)


pablosaavedra-rappi:
Didn't find any test referencing PoolImpl that I could use to replicate the behavior

vietj:
I think the change is good but will impact pooled clients for which the promise is to schedule commands without waiting for a connection to be free, so we should have this behaviour configurable and disabled when producing a pooled client. See `PgBuilder#client()` vs `PgBuilder#pool()`

vietj (changes requested):
I think the change is good but will impact pooled clients for which the promise is to schedule commands without waiting for a connection to be free, so we should have this behaviour configurable and disabled when producing a pooled client. See `PgBuilder#client()` vs `PgBuilder#pool()`

pablosaavedra-rappi:
Thanks for the comment, will look into it and update the PR.

pablosaavedra-rappi:
@vietj finally got around to working on this one. I rebased with the 4.x branch and added the changes you requested. Let me know if you have any other comment.

Thanks.

vietj:
@pablosaavedra-rappi yes you can push generated source code

vietj:
Looking at this again, I am not a fan of the option as is, I think instead command should have their own timeout to explicitely named the feature, and we should call it `scheduleTimeout` which is the timeout to schedule a command, if a connection has not been made available within that timeout then the command cannot be scheduled and is failed

vietj:
@pablosaavedra-rappi

pablosaavedra-rappi:
So @vietj you are suggesting we overload the method with an additional `scheduleTimeout` parameter and the behavior there? I can check how that'd look and update the PR.

vietj:
instead of having a boolean that says that the timeout applies to
everything, we have a timeout that applies to command execution scheduling.

PR #1339, opened 2023-07-22
MySQL client data type mapping rework - fixes #1239

### LINKED ISSUE #1239: MySQL client should provide columnDescriptors with LONGVARCHAR JDBCType for TEXT types
[See this issue in Quarkus](https://github.com/quarkusio/quarkus/issues/22254#issuecomment-1259563555).

Is there any particular reason for this or it's a bug?

@fedinskiy:
```
I took a look into code for Postgress[1] and MySQL[2] and it looks like the text case is not handled properly in the latter.

[1] https://github.com/eclipse-vertx/vertx-sql-client/blob/master/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataType.java
[2] https://github.com/eclipse-vertx/vertx-sql-client/blob/master/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/datatype/DataType.java
```

BillyYccc:
Motivation:

This PR reworks the MySQL client data type mapping part, with fixes to #1239 

Modifications:
1. move checking unsigned & binary flags from data type codec to column definition parsing phase which can avoid checking those flags at row value decoding phase for each prepared statement executing and improve some performance
2. add dedicated unsigned 
numeric & binary datatypes to the `DataType` enum so that the data type codec can use them to decode row values
3. make some **changes** to the MySQL JDBCType mapping
FLOAT: JDBCType.REAL-> JDBCType.FLOAT
CHAR,ENUM: JDBCType.VARCHAR -> JDBCType.CHAR
BINARY: JDBCType.VARCHAR -> JDBCType.BINARY
VARBINARY: JDBCType.VARCHAR -> JDBCType.VARBINARY
TINY_TEXT: JDBCType.BLOB -> JDBCType.CLOB
TEXT: JDBCType.BLOB -> JDBCType.CLOB
MEDIUM_TEXT: JDBCType.BLOB -> JDBCType.CLOB
LONG_TEXT: JDBCType.BLOB -> JDBCType.CLOB

### DIFF
--- a/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/CommandCodec.java
+++ b/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/codec/CommandCodec.java
@@ -176,7 +176,7 @@ abstract class CommandCodec<R, C extends CommandBase<R>> {
     int characterSet = payload.getUnsignedShortLE(start + bytesToSkip);
     bytesToSkip += 6; // characterSet + columnLength
 
-    DataType type = DataType.valueOf(payload.getUnsignedByte(start + bytesToSkip));
+    short type = payload.getUnsignedByte(start + bytesToSkip);
     bytesToSkip++;
 
     int flags = payload.getUnsignedShortLE(start + bytesToSkip);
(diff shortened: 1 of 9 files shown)


vietj:
thanks @BillyYccc !

PR #1342, opened 2023-07-31
Fail with exception on timeout while waiting for connection pool

dennisschroer:
Motivation:

When a timeout occurs while waiting for a new connection, the future fails with a simple "Timeout"  message without a stack trace. This makes it very difficult to debug where in the application the error occurs. We have a longer running process with multiple requests to the database, and having a stacktrace would help us.

Conformance:


vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/SqlConnectionPool.java
```diff
@@ -217,12 +218,10 @@ public void onEnqueue(PoolWaiter<PooledConnection> waiter) {
         if (timeout > 0L && timerID == -1L) {
           timerID = context.setTimer(timeout, id -> {
             pool.cancel(waiter, ar -> {
-              if (ar.succeeded()) {
-                if (ar.result()) {
-                  handler.handle(Future.failedFuture("Timeout"));
-                }
+              if (ar.succeeded() && ar.result()) {
+                handler.handle(Future.failedFuture(new ConnectionPoolTooBusyException("Timeout while waiting for connection")));
```
vietj, line 222: rather use java.util.concurrent.TimeoutException

PR #1357, opened 2023-09-23
Client builder API replacing Pool sub types

vietj:
- introduce a new client builder API that replaces the pool static methods
- deprecates various Pool subtypes (e.g `PgPool)` as they are not necessary other than carrying static methods
- deprecate `Pool#connectionHandler` and `Pool#connectionProvider`

### DIFF
--- a/vertx-db2-client/src/main/asciidoc/index.adoc
+++ b/vertx-db2-client/src/main/asciidoc/index.adoc
@@ -99,7 +99,7 @@ Once you are done with the connection you must close it to release it to the poo
 
 == Pool versus pooled client
 
-The {@link io.vertx.db2client.DB2Pool} allows you to create a pool or a pooled client
+The {@link io.vertx.db2client.DB2Builder} allows you to create a pool or a pooled client
 
 [source,$lang]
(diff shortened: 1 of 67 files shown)


ghubstan:
@vietj , I have been long at work on a project depending on `vertx-pg-client`, and upgrading `Quarkus` (today)  also upgraded my pg-client lib to 4.5.4, and I see `PgPool` is deprecated.

I assume you would like to see others switch to the new [client builder API](https://github.com/eclipse-vertx/vertx-sql-client/pull/1357)  from PgPool ASAP (help find bugs).  Are you already encouraging ppl to migrate?

Can you roughly estimate how long I (we) have before `PgPool` is removed from the project, or is no longer usable, whichever comes first?   I am willing to migrate soon -- and report issues I might find --  but would like your feedback first, in case you think the new API is still too "alpha".

Great work, by the way...

vietj:
@ghubstan this deprecation will hold for Vert.x 4 and is removed effectively in Vert.x 5. People should migrate but don't have to unless they want to use Vert.x 5.

PR #1384, opened 2023-11-29
Considering compatibility reasons, should MySQL's BIGINT UNSIGNED be mapped to BigInteger?

jinwik:
Motivation:

https://github.com/mysql/mysql-connector-j/blob/release/8.x/src/main/user-impl/java/com/mysql/cj/jdbc/result/ResultSetImpl.java#L1153

Conformance:


tsegismont:
From a quick look, it's fine with me, wdyt @BillyYccc ?

We'd need a test case for encoding as well

PR #1386, opened 2023-11-29
Set SQL state when there is an exception #1385

### LINKED ISSUE #1385: DatabaseException#getSqlState is null for MS SQL Server
When a new `MSSQLException` is created, the value of DatabaseException#sqlState is always set to null.

See  https://github.com/eclipse-vertx/vertx-sql-client/blob/094817189aefce280a6bc7d391f0447fa20066d3/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLException.java#L33

The MSSQL JdbcDriver actually has a method to convert the vendor code to the sql state code: https://github.com/microsoft/mssql-jdbc/blob/ebf4d78eb985c7759cc74d4eb3c0de8cd010387d/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerException.java#L333

DavideD:
Fix #1385

### DIFF
--- a/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLException.java
+++ b/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/MSSQLException.java
@@ -30,13 +30,48 @@ public class MSSQLException extends DatabaseException {
   private List<MSSQLException> additional;
 
   public MSSQLException(int number, byte state, byte severity, String errorMessage, String serverName, String procedureName, int lineNumber) {
-    super(formatMessage(number, state, severity, errorMessage, serverName, procedureName, lineNumber), number, null);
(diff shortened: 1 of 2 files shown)


DavideD:
I think I did that in the past

DavideD:
I've checked the website for the ECA and it seems to me that everything is fine

tsegismont:
> I've checked the website for the ECA and it seems to me that everything is fine

Did you use the same email for your commit? The [report](https://github.com/eclipse-vertx/vertx-sql-client/pull/1386#issuecomment-1836439537) says Eclipse is not aware of your hibernate org address

DavideD:
When I go on the website of the link, it says it is and that's valid until 2024

DavideD:
The commit that I've sent had the verified label, though

PR #1387, opened 2023-12-01
MSSQL row decoder flow alignment

vietj:
The MSSQL row decoder bypasses the decoder flow since handling a row might produce no row. We would like to have the same flow to be used.

The row decoder handleRow method now checks if the returned row is not null before passing it to the collector. In addition we need to use a specific mode to let the mssql decoder which decoding it should use dynamically.


tsegismont (approved):
LGTM :shipit:

Issue #1411, opened 2024-02-22
PGClient not correctly handling SSL MODE ALWAYS


pendula95:
### Version
4.5.3

When settings SSL mode to ALLOW if pg client should do following
try without SSL -> if fail try with SSL
This is tried here:
https://github.com/eclipse-vertx/vertx-sql-client/blob/a2d2f9002a5fce562c8236c3310faad98038bb0d/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgConnectionFactory.java#L106-L107

Unfortunately recover is never called because doConnect never fails as it does not try to send Startup message.
This is done later in code on and it can not recover:
https://github.com/eclipse-vertx/vertx-sql-client/blob/a2d2f9002a5fce562c8236c3310faad98038bb0d/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgConnectionFactory.java#L81

I have created a PR with tests to cover this.
I tried to fix this by myself by always sending Startup message before returning from
https://github.com/eclipse-vertx/vertx-sql-client/blob/a2d2f9002a5fce562c8236c3310faad98038bb0d/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgConnectionFactory.java#L97 
and this resulted in recover to trigger but code fails in trying to upgrade to SSL. My best guess is that the state of socket is wrong at that point. I willing to submit a PR if I get a little more guidance on how to do it correctly.

vietj:
thanks for reporting and providing a reproducer @pendula95

it has been backported to 4.x branch

PR #1412, opened 2024-02-22
Fix SSL falback for SSL MODE ALWAYS

pendula95:
Related to #1411


vietj:
@pendula95 https://github.com/eclipse-vertx/vertx-sql-client/pull/1413

I reworked the code to move the send initial message earlier as you hinted, the test still fails with a password issue it seems related to configuration

PR #1413, opened 2024-02-22
Correct handling of SSL mode always for PostgreSQL

vietj:


pendula95:
I am not able to evaluate but now this changes the behavior of https://github.com/eclipse-vertx/vertx-sql-client/blob/a2d2f9002a5fce562c8236c3310faad98038bb0d/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgConnectionFactory.java#L86-L95

Previously canceled requests did not send Startup. For this I suggest pass a flag `cancelRequest`

vertx-pg-client/src/test/java/io/vertx/pgclient/TLSTest.java
```diff
@@ -104,18 +110,30 @@ public void testSslModeDisable(TestContext ctx) {
...
   }
 
+  @Test
+  public void testSslModeAllowFallback(TestContext ctx) {
+    Async async = ctx.async();
+    PgConnectOptions options = ruleForceSsl.options()
+      .setSslMode(SslMode.ALLOW)
+      .setSslOptions(new ClientSSLOptions().setTrustAll(true));
+    PgConnection.connect(vertx, new PgConnectOptions(options)).onComplete(ctx.asyncAssertSuccess(conn -> {
+      ctx.assertFalse(conn.isSSL());
```
vietj, line 128: that should be `assertTrue` as far as I can tell

vertx-pg-client/src/test/java/io/vertx/pgclient/TLSTest.java
```diff
@@ -124,9 +142,21 @@ public void testSslModePrefer(TestContext ctx) {
...
   }
 
+  @Test
+  public void testSslModePreferFallback(TestContext ctx) {
+    Async async = ctx.async();
+    PgConnectOptions options = ruleSllOff.options()
+      .setSslMode(SslMode.PREFER)
+      .setSslOptions(new ClientSSLOptions().setTrustAll(true));
+    PgConnection.connect(vertx, options).onComplete(ctx.asyncAssertSuccess(conn -> {
+      ctx.assertTrue(conn.isSSL());
```
vietj, line 152: that should be `assertFalse` as far as I can tell

vertx-pg-client/src/test/java/io/vertx/pgclient/junit/ContainerPgRule.java
```diff
@@ -36,18 +36,25 @@ public class ContainerPgRule extends ExternalResource {
 
   private static final String connectionUri = System.getProperty("connection.uri");
   private static final String tlsConnectionUri = System.getProperty("tls.connection.uri");
+  private static final String tlsForceConnectionUri = System.getProperty("tls.force.connection.uri");
```
vietj, line 39: README.adoc needs to be updated accordingly @pendula95

vietj:
@pendula95 a couple of concerns raised

vertx-pg-client/src/test/java/io/vertx/pgclient/TLSTest.java
pendula95: This should be SSL connection. This is force SSL configuration  (only allows SSL) so the allow mode will first try with no SSL it will fail and it will recover by trying SSL. So if the connection is successful only viable option is SSL true. https://github.com/eclipse-vertx/vertx-sql-client/blob/1bb9d04d42ad050b528925e05e85a269eb328617/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgConnectionFactory.java#L97-L98

pendula95:
You are quite right, I have corrected everything here: https://github.com/eclipse-vertx/vertx-sql-client/pull/1421

Sorry for creating additional PRs but as I dont have permissions to directly commit here I dont see any other way

vietj:
@pendula95 don't be sorry, that's how it is supposed to be done

pendula95:
https://github.com/eclipse-vertx/vertx-sql-client/pull/1422

PR #1418, opened 2024-02-28
CIDR Support in vertx-pg-client

AyanKoche:
Motivation: CIDR Data type support in vertx-pg-client

There are several data types supported by vertx-pg-client but CIDR was missing implementation. As the project I am working on requires that support, contributing it is the better way. 
Changes:
- Have added a CIDR class
- Have added 3 methods : binaryEncodeCidr, binaryDecodeCidr, textDecodeCidr
- Have added tests for CIDR class
- Have made relevant changes to the docs as well

Conformance:


vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
```diff
@@ -1713,4 +1720,92 @@ private static <T> void textEncodeArray(T[] values, DataType type, ByteBuf buff)
...
+    int size = buff.getByte(index+3);
+    byte[] data = new byte[size];
+    buff.getBytes(index+4,data);
+    InetAddress address;
+
+    switch (family){
+      case 2:
+        // IPV4
+        try {
+          address = Inet4Address.getByAddress(data);
```
vietj, line 1737: Actually `InetAddress.getByAddress`

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
```diff
@@ -1713,4 +1720,92 @@ private static <T> void textEncodeArray(T[] values, DataType type, ByteBuf buff)
...
+        }catch (UnknownHostException e){
+          throw new DecoderException(e);
+        }
+        break;
+      case 3:
+        // IPV6
+        try {
+          address = Inet6Address.getByAddress(data);
```
vietj, line 1745: Actually `InetAddress.getByAddress`

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
```diff
@@ -1713,4 +1720,92 @@ private static <T> void textEncodeArray(T[] values, DataType type, ByteBuf buff)
...
+    byte netmask = buff.getByte(index+1);
+    Integer val;
+    int size = buff.getByte(index+3);
+    byte[] data = new byte[size];
+    buff.getBytes(index+4,data);
+    InetAddress address;
+
+    switch (family){
+      case 2:
+        // IPV4
```
vietj, line 1735: IPV4 and IPV6 seems to be the same code and could be

if (family == 2 || family == 3) instead

vietj (changes requested):
a few concerns to address

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
AyanKoche: Addressed.

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
AyanKoche: Addressed.

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/codec/DataTypeCodec.java
AyanKoche: Addressed.

AyanKoche:
Addressed the previous comments. Please re-review it. @vietj

vietj (changes requested):

vietj:
can you squash your commits in a single one ?

AyanKoche:
Sure ,doing it

AyanKoche:
@vietj Hi sorry for delayed response,  I have made the changes and squashed the commits into single commit. Kindly have a look at it

AyanKoche:
Hi @vietj , any review comments on the PR?

vietj (changes requested):
I cannot see any tests with a real database and the CIDR type, we need to have at least a test that read / write CIDR objects to a test container postgresql instance. Look at how other specific types are tested.

vietj:
great feature thanks @AyanKoche

PR #1421, opened 2024-03-06
Pendula95 fix recover for ssl mode always patch 1

pendula95:


vietj:
can you instead PR https://github.com/eclipse-vertx/vertx-sql-client/tree/pendula95-fix-recover-for-ssl-mode-always ?

PR #1432, opened 2024-04-08
Branch free parse digits

franz1981:
It contains 3 improvements (unfinished):
- [O(1) collation lookup table](https://github.com/eclipse-vertx/vertx-sql-client/commit/e94a0288c9b20c4373e7f3b5bd4b3af3c4945360)
- [branchless/batchy parse digits](https://github.com/eclipse-vertx/vertx-sql-client/commit/cf176f47c270cffb04df4e47639da440450d4a0e)
- NOT YET STARTED: https://github.com/eclipse-vertx/vertx-sql-client/blob/9073da6171b7688a7491398522dcb6c454bd5960/vertx-mysql-client/src/main/java/io/vertx/mysqlclient/impl/datatype/DataTypeCodec.java#L61 exceed `MinJumpTableSize` (see https://stackoverflow.com/questions/44988841/force-tableswitch-instead-of-lookupswitch for more info): we should find a way to reduce the number of branches here or find a different way to perform it, because the branch predictor AND the JIT doesn't enjoy it


franz1981:
Some of the proposed improvements are coming from https://github.com/cescoffier/fortune-benchmark/tree/main/fortune-reactive-mariadb/src/main/java/me/escoffier/fortune/reactive/mariadb so IMO they are not actually so valid (eg the cost of decoding in digits) , but given that I was having fun reading a nice article on it I said "why not trying implementing it?"

Issue #1450, opened 2024-07-26
Oracle Connection Health Check not implemented


weberk:
https://github.com/eclipse-vertx/vertx-sql-client/blob/3417c5ab1e346f852a7b9ae99e9a7d71637dbac2/vertx-oracle-client/src/main/java/io/vertx/oracleclient/impl/OracleJdbcConnection.java#L106
io.vertx.oracleclient.impl.OracleJdbcConnection.isValid() returns allways true.
```java
  @Override
  public boolean isValid() {
    return true;
  }
```

Please implement by delegating to the oracle.jdbc.OracleConnection. So calling connection.isValid() you get the JDBC4 compliant result the Oracle JDBC driver delivers, which is much. more efficient than any alternative i.e. calling some table called `dual`!
```java
  @Override
  public boolean isValid() {
    return connection.isValid(oracle.jdbc.OracleConnection.ConnectionValidation.SOCKET, 0);
  }
```

In order to speedup health check using isValid() it might be good to put the following property to the defaults:
```properties
quarkus.datasource.reactive.additional-properties.connectionProperties=oracle.jdbc.defaultConnectionValidation=SOCKET
```

tsegismont:
We can't call JDBC `connection.isValid()` here because it's blocking call. Besides, the expectation from the caller is only to check if the TCP connection is still open.

We have talked in the past about connection validation mechanism when borrowing from the pool, but I can't find any issue.

Does that ring a bell @vietj ?

weberk:
Calling oracle.jdbc.OracleConnection.isValid() is not blocking. For Oracle at least with `ConnectionValidation=SOCKET` it is a quick check of the socket status with the Linux kernel. No remote (blocking) network activity is triggered. Using [lightweight connection validation](https://docs.oracle.com/en/database/oracle/oracle-database/23/jjdbc/JDBC-getting-started.html#GUID-6AFB71F0-EFA9-43A0-AF96-03E6FB2F3521:~:text=Note%3A-,Lightweight%20connection%20validation,-checks%20only%20the) overcomes a limitation of the Java Runtime not consuming Linux signals raised on TCP-FIN events happening on the network. So the Linux kernel knows that the connection is closed, but the Java code has no clue. So the connection pool would hand out a dead connection even though it could have been replaced and some (weak) business logic error handlers are not executed. It improves resilience if isValid() is called.

tsegismont:
[...]

PR #1451, opened 2024-07-26
Oracle: Lightweight Connection Validation

tsegismont:
See #1450

### DIFF
--- a/vertx-oracle-client/src/main/java/io/vertx/oracleclient/impl/OracleJdbcConnection.java
+++ b/vertx-oracle-client/src/main/java/io/vertx/oracleclient/impl/OracleJdbcConnection.java
@@ -108,7 +108,7 @@ public class OracleJdbcConnection implements Connection {
   @Override
   public boolean isValid() {
     try {
-      return connection.isValid(OracleConnection.ConnectionValidation.SOCKET, 0);
+      return connection.isValid(OracleConnection.ConnectionValidation.NONE, 0);
     } catch (SQLException e) {
       log.trace("Failed to validate connection", e);
       return false;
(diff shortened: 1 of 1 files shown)


tsegismont:
Hi @loiclefevre , I would appreciate your advice on this. Is the `SOCKET` connection validation level appropriate (non-blocking)? Or should we consider something lower, like `INBAND_DOWN` or `LOCAL`?

cc @DavideD

gvenzl:
Hey @tsegismont,

@loiclefevre is on his well-deserved holiday, hence I'm adding @jeandelavarene to this thread to provide guidance on your question about `OracleConnection.ConnectionValidation.` and which mode to use.

jeandelavarene:
Here is a description of the different validation modes in decreasing order of validation strength:

- "SERVER" (or "COMPLETE") makes the driver execute a basic SQL query "SELECT 'x' FROM DUAL",
- "NETWORK" makes the driver issue an OPING TTC function. This is the most lightweight roundtrip to the DB. It's the default validation mode.
- "SOCKET" makes the driver write a zero-length NS data packet on the socket that is ignored by the server. There is no guarantee that this call will always be non-blocking. It should be non-blocking in most of the cases.
- "INBAND_DOWN" makes the driver do a non-blocking socket read call on the socket
- "LOCAL" same as above
- "NONE" just checks the connection's lifecycle (variable check)

Note that, for example, a "SOCKET" validation may return a false positive (it may not always detect all socket defects). On the other hand, a "SOCKET" validation is much faster than "NETWORK". In most cases, we recommend using "SOCKET" validation at connection borrow (in UCP for example). For applications that can't tolerate an error on the connection, turning on "SERVER" or "NETWORK" would be more appropriate.

tsegismont:
Thank you @gvenzl and @jeandelavarene 

In this part of our code we can't tolerate the driver to block (event not every time). And for the other clients (Pg, MySQL, ...etc) where Vert.x manages the connection, we only check the connection's lifecycle (variable check):

https://github.com/eclipse-vertx/vertx-sql-client/blob/ccc784a3a2b1e3f8790fb2781f14404fa83886ea/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SocketConnectionBase.java#L160

So I think the appropriate setting is `NONE` in our case. This will provide the same experience as with other clients.

PR #1460, opened 2024-08-28
NPE in PgDecoder if a notice is raised while no query is being executed

tsegismont:
See #1442

This may happen if the connection is used by a PgSubscriber.


vietj:
no test ?

tsegismont:
@vietj I searched the web for cases when the server would raise a notice/error not related to a client request, but I couldn't find any. Have you encountered this case yourself before?

vietj:
I think it can be reproduced using a proxy, we have similar tests like that

vietj:
you can look at testDisconnectAbruptlyDuringStartup

vietj:
yes, Im saying to use the proxy server to create a fake message to be
sent to the client using it

tsegismont:
@vietj PTAL

Thanks for the tip

vietj (approved):

PR #1462, opened 2024-09-06
Pg Client: improve Interval support

### LINKED ISSUE #1281: Reasonable support for `java.time.Duration` -> Interval conversion
#### Describe the feature

Working with the standard `java.time.Duration` class in postgres is unnecessarily painful.

#### Use cases

I ran into this issue with a simple query that invokes a stored procedure:

```sql
select incident_count,risk_level from safety_score_v2($1::interval, $2::latitude, $3::longitude)
```

where the code supplying the `interval` is handed a `java.time.Duration`.  Having done some testing in `psql`, Postgres supports ISO-8601 duration strings perfectly well in a query - i.e. `P1Y::interval` gets you an interval of a year.  Great!  So, I try to pass `Tuple.of(duration.toString(), ...)`.

vertx-pg-client helpfully (and **silently!**) replaces the duration with `NULL`.  Very surprising when my queries start returning tens of thousands of result.

Tried a few incantations of `::varchar` and similar trying to head-fake vertx-pg-client into doing the right thing.  No luck.

Solutions:

 * First, `Tuple` should really just accept a `java.time.Duration` and convert it to an `Interval` (or whatever) under the hood
 * Second, a well-formed ISO-8601 string for an interval should be accepted
 * Third, there should be some sane way of converting `Duration` to `Interval` in a constructor, so if none of the above is acceptable, there is at least a low-complexity way of doing that (could be as simple as creating an interval with the number of seconds and microseconds in the duration - though it would be more polite to anyone reading queries from the database log to extract those elements that don't have varying lengths - i.e. days, hours, minutes, seconds, microseconds)

#### Contribution

I might consider it.

Quick'n'dirty, unpretty code to do the job *with some necessary inaccuracy around years and months* (but good enough for my current purposes):

```java
    static Interval durationToInterval(java.time.Duration duration) {
        // Vertx's duration class really ought to do this itself.
        double seconds = duration.toSeconds();
        long years = (long) (seconds / 31540000.0010112);
        long yearsRemainder = (long) (seconds - (years * 31540000.0010112));
        long months = (long) (yearsRemainder / 2628336.2137828967534);
        long monthsRemainder = yearsRemainder - (long) (months * 2628336.2137828967534);
        long days = (long) (monthsRemainder / 86410.958906879997812);
        return new Interval(
                (int) years,
                (int) months,
                (int) days,
                duration.toHoursPart(),
                duration.toMinutesPart(),
                duration.toSecondsPart()
        );
    }
```

tsegismont:
Faster and garbage-free codec, compatibility with `java.time.Duration`

Closes #1281

Commits can be reviewed individually.

### DIFF
--- a/vertx-pg-client/src/main/java/io/vertx/pgclient/data/Interval.java
+++ b/vertx-pg-client/src/main/java/io/vertx/pgclient/data/Interval.java
@@ -244,7 +244,7 @@ public class Interval {
   /**
    * Convert this interval to an instance of {@link Duration}.
    * <p>
-   * The conversion algorithm assumes a year last 12 months and a month lasts 30 days, as <a href="https://github.com/postgres/postgres/blob/5bbdfa8a18dc56d3e64aa723a68e02e897cb5ec3/src/include/datatype/timestamp.h#L116">Postgres does</a> and ISO 8601 suggests.
+   * The conversion algorithm assumes a year lasts 12 months and a month lasts 30 days, as <a href="https://github.com/postgres/postgres/blob/5bbdfa8a18dc56d3e64aa723a68e02e897cb5ec3/src/include/datatype/timestamp.h#L116">Postgres does</a> and ISO 8601 suggests.
    *
    * @return an instance of {@link Duration} representing the same amount of time as this interval
    */
--- a/vertx-pg-client/src/main/java/io/vertx/pgclient/data/Interval.java
+++ b/vertx-pg-client/src/main/java/io/vertx/pgclient/data/Interval.java
@@ -86,7 +86,7 @@ public class Interval {
   /**
    * Creates an instance from the given {@link Duration}.
    * <p>
-   * The conversion algorithm assumes a year last 12 months and a month lasts 30 days, as <a href="https://github.com/postgres/postgres/blob/5bbdfa8a18dc56d3e64aa723a68e02e897cb5ec3/src/include/datatype/timestamp.h#L116">Postgres does</a> and ISO 8601 suggests.
+   * The conversion algorithm assumes a year lasts 12 months and a month lasts 30 days, as <a href="https://github.com/postgres/postgres/blob/5bbdfa8a18dc56d3e64aa723a68e02e897cb5ec3/src/include/datatype/timestamp.h#L116">Postgres does</a> and ISO 8601 suggests.
    *
    * @param duration the value to convert
    * @return a new instance of {@link Interval}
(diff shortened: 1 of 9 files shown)


tsegismont:
Not ready yet, do not merge

tsegismont:
This is ready for review, PTAL @vietj

vertx-pg-client/src/main/java/io/vertx/pgclient/data/Interval.java
```diff
@@ -84,6 +83,35 @@ public static Interval of(int years) {
     return new Interval(years);
   }
 
+  /**
+   * Creates an instance from the given {@link Duration}.
+   * <p>
+   * The conversion algorithm assumes a year last 12 months and a month lasts 30 days, as <a href="https://github.com/postgres/postgres/blob/5bbdfa8a18dc56d3e64aa723a68e02e897cb5ec3/src/include/datatype/timestamp.h#L116">Postgres does</a> and ISO 8601 suggests.
```
tsegismont, line 89:
```suggestion
   * The conversion algorithm assumes a year lasts 12 months and a month lasts 30 days, as <a href="https://github.com/postgres/postgres/blob/5bbdfa8a18dc56d3e64aa723a68e02e897cb5ec3/src/include/datatype/timestamp.h#L116">Postgres does</a> and ISO 8601 suggests.
```

vertx-pg-client/src/main/java/io/vertx/pgclient/data/Interval.java
```diff
@@ -203,7 +231,25 @@ public int hashCode() {
...
+           + hours + " hours "
+           + minutes + " minutes "
+           + seconds + " seconds "
+           + microseconds + " microseconds )";
+  }
+
+  /**
+   * Convert this interval to an instance of {@link Duration}.
+   * <p>
+   * The conversion algorithm assumes a year last 12 months and a month lasts 30 days, as <a href="https://github.com/postgres/postgres/blob/5bbdfa8a18dc56d3e64aa723a68e02e897cb5ec3/src/include/datatype/timestamp.h#L116">Postgres does</a> and ISO 8601 suggests.
```
tsegismont, line 247:
```suggestion
   * The conversion algorithm assumes a year lasts 12 months and a month lasts 30 days, as <a href="https://github.com/postgres/postgres/blob/5bbdfa8a18dc56d3e64aa723a68e02e897cb5ec3/src/include/datatype/timestamp.h#L116">Postgres does</a> and ISO 8601 suggests.
```

PR #1465, opened 2024-09-10
Use org.testcontainers:testcontainers instead of per database specializations

vietj:
Motivation:

Test containers specializations for database do have a split package with the main test containers jars, preventing running tests in IDE. Since the benefit of these specialization is very thin, the tradeof is not worth and we should use basic testcontainers.

Changes:

Use org.testcontainers:testcontainers everywhere.


vietj:
done the right way by @tsegismont

PR #1548, opened 2025-09-18
Temporarily use bitnamilegacy repository

tsegismont:
This gives us some time to plan a migration. Not sure how much.


tsegismont:
See https://github.com/vert-x3/issues/issues/662

PR #1566, opened 2025-11-07
Oracle: Support POOLED server mode for DRCP

basselworkforce:
Motivation:

Add `POOLED` as an option to `ServerMode` in order to support Database Resident Connection Pooling (DRCP) on Oracle databases: https://docs.oracle.com/en/database/oracle/oracle-database/26/jjdbc/database-resident-connection-pooling.html#GUID-6C968C9D-2AAE-4D42-A6EC-3D78335F9BE3


basselworkforce:
Sorry for the late response but yes, it seems like it is more complicated than I thought before (and also doesn't seem to work over TCPS) so closing this for now

PR #1568, opened 2025-11-10
SqlConnectionPool does not invoke PoolMetrics begin/end methods

adamdickmeiss:
Fixes https://github.com/vert-x3/vertx-micrometer-metrics/issues/278


tsegismont:
Thank you @adamdickmeiss

PR #1064, opened 2021-11-08
Preferred server type connection factory

Rattenkrieg:
This PR introduces a way to specify required/preferred server type among multiple servers when several hosts provided to pool factory.
Example:
```java
PgPool.pool(Vertx.vertx(), List.of(primaryHostOptions, replica1HostOptions, replica2HostOptions), poolOptions.setServerRequirement(PREFER_REPLICA))

```

Similar functionality provided by pgjdbc via [targetServerType](https://jdbc.postgresql.org/documentation/head/connect.html#connection-failover) and libpq via [target_session_attrs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-MULTIPLE-HOSTS).

- [x] javadocs

PR #1623, opened 2026-02-23
feat: preferred connection type pg only

Rattenkrieg:
- Add `targetServerType` support for PostgreSQL multi-host connection pools, matching the semantics of pgjdbc's `targetServerType` and libpq's `target_session_attrs`;
- Revives #1064 which targeted 4.x; this implementation is redesigned for the 5.x `ConnectionFactory` architecture;
- Supports `ANY`, `PRIMARY`, `SECONDARY`, `PREFER_PRIMARY`, and `PREFER_SECONDARY` modes;
- Server type detection uses PG 14+ `in_hot_standby` GUC_REPORT (zero-cost, during handshake) with `SHOW transaction_read_only` fallback for older versions;

Differences from #1064

- Single shared factory instead of one `ConnectionFactory` per host - 5.x `ConnectionFactory.connect()` takes options per-call, so a `ServerType[]` array indexed by server position replaces per-factory state;
- Stale cache recovery - when all servers are skipped by cache, retries with cache disabled to force re-verification. #1064 had no cache invalidation, so failovers caused permanent misrouting;
- Connection leak on detection failure - if `detectServerType` fails after connect, the connection is explicitly closed before retrying. #1064 did not handle this path;
- Close-before-retry - mismatched connections are closed and the close is awaited before opening the next probe, matching pgjdbc behavior and ensuring PG connection slots are freed promptly;

Connection algorithm

The `ServerTypeAwarePgConnectionFactory` wraps `PgConnectionFactory` and manages a list of servers with round-robin selection:

- `PRIMARY` / `SECONDARY` (strict): Probes servers in round-robin order, returns the first match, closes mismatched connections. Fails with an error if no server of the desired type is found;
- `PREFER_PRIMARY` / `PREFER_SECONDARY`: Two-pass approach matching pgjdbc - strict pass first;
    (probe-and-close for mismatches), then falls back to connectToAny which accepts any reachable server regardless of type;
- `ANY`: Connects to the first reachable server with no type detection;
-

### DIFF
--- a/vertx-pg-client/src/main/java/io/vertx/pgclient/TargetServerType.java
+++ b/vertx-pg-client/src/main/java/io/vertx/pgclient/TargetServerType.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.vertx.pgclient;
+
+/**
+ * Defines the target server type when connecting to a PostgreSQL cluster with multiple hosts.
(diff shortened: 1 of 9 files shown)


Rattenkrieg:
Hey @tsegismont @vietj , are you interested moving it forward?

tsegismont:
@Rattenkrieg thanks for sharing your proposal. There are some good ideas but I don't have time at this moment to commit to such an important change.

PR #1632, opened 2026-03-11
Add support for streaming and cursor query execution in SQL templates

### LINKED ISSUE #1624: Support streaming in SqlTemplate
### Describe the feature

It would be nice if SqlTemplate exposed something like executeStream(int size, Map<String, Object> params) method that would provide ability to use row streaming (on supported clients). Otherwise user needs to result to using raw client, which is viable but if you are using SqlTemplate and for streaming you need to switch directly to use the client it can be a bit clunky.

### Contribution

I would be willing to take a look at this

zZHorizonZz:
Motivation:
- Added a support for streaming as discussed in #1624 to handle large result sets via streaming or cursor queries trough `vertx-sql-client-templates`

Changes:
- Added `forStream` and `forCursor` methods to `SqlTemplate`.
- Added `SqlTemplateStreamImpl`, `CursorSqlTemplateImpl`, and `MappingRowStream`
- Added documentation to `index.adoc` with examples

Closes: #1624

### DIFF
--- a/vertx-sql-client-templates/src/main/asciidoc/index.adoc
+++ b/vertx-sql-client-templates/src/main/asciidoc/index.adoc
@@ -65,6 +65,35 @@ When your template must be executed inside a transaction, you might create a tem
 ----
 ====
 
+== Streaming
+
+When dealing with large result sets, you can use {@link io.vertx.sqlclient.templates.SqlTemplateStream} to read rows progressively using a cursor with a configurable fetch size, instead of loading all rows in memory at once.
+
+NOTE: Streaming requires a {@link io.vertx.sqlclient.SqlConnection}. Some databases (e.g. PostgreSQL) also require an active transaction for cursors.
+
+[source,$lang]
+----
+{@link examples.TemplateExamples#streamExample}
+----
+
+You can use `mapTo` to map each row emitted by the stream to a custom type:
+
+[source,$lang]
+----
+{@link examples.TemplateExamples#streamWithMapToExample}
+----
+
+== Cursor
+
+If you need finer control over row fetching, you can use a cursor-based template with {@link io.vertx.sqlclient.templates.SqlTemplate#forCursor}. This gives you a {@link io.vertx.sqlclient.Cursor} that allows you to read rows in batches.
+
+NOTE: Cursors require a {@link io.vertx.sqlclient.SqlConnection}. Some databases (e.g. PostgreSQL) also require an active transaction for cursors.
+
+[source,$lang]
+----
+{@link examples.TemplateExamples#cursorExample}
+----
+
 == Template syntax
 
 The template syntax uses `#{XXX}` syntax where `XXX` is a valid https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.8[java identifier] string
--- a/vertx-sql-client-templates/src/main/java/examples/TemplateExamples.java
+++ b/vertx-sql-client-templates/src/main/java/examples/TemplateExamples.java
@@ -9,6 +9,7 @@ import io.vertx.docgen.Source;
 import io.vertx.sqlclient.*;
 import io.vertx.sqlclient.templates.RowMapper;
 import io.vertx.sqlclient.templates.SqlTemplate;
+import io.vertx.sqlclient.templates.SqlTemplateStream;
 import io.vertx.sqlclient.templates.TupleMapper;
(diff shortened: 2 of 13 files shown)


tsegismont (commented):
At first glance, I'm wondering why do we need a separate `SqlTemplateStream`.

In the original issue we talked about adding a couple methods to `SqlTemplate` and I assumed they would return a `Cursor` or a `RowStream`.

tsegismont (changes requested):
Thank you for the updates @zZHorizonZz 

I took some time to review the proposal and limitations, and I think we're not going in the right direction. Indeed, we have new APIs that look like the original `SqlClientTemplate`, but not all methods can be supported because it doesn't fit with the usage model of cursor/rowstream.

I think that, instead, we should make it possible for users to parse a template query and map objects to parameters. Then they can use the existing preparedQuery, cursor and rowstream APIs like they usually do.

The parser could look like this (just the basics):

```java
@VertxGen
public interface SqlTemplateParser {

  /**
   * Create a parser by extracting driver info from a SqlClient.
   * The client reference is NOT retained after this call.
   *
   * @param client the SQL client to extract driver information from
   * @param template the SQL template string with named parameters (e.g., "SELECT * FROM users WHERE id = #{id}")
   * @return the template parser
   */
  static SqlTemplateParser create(SqlClient client, String template) {
    SqlClientInternal internal = (SqlClientInternal) client;
    return SqlTemplateParserImpl.create(internal.driver(), template);
  }

  /**
   * Get the parsed SQL query with driver-specific placeholders.
   *
   * @return the SQL query string
   */
  String sql();

  /**
   * Convert parameters to Tuple using the provided TupleMapper.
   *
   * @param params the parameters object
   * @param mapper the tuple mapper
   * @param <T> the parameters type
   * @return the tuple
   */
  <T> Tuple mapToTuple(T params, TupleMapper<T> mapper);
}

  /**
   * Convert Map parameters directly to Tuple.
   *
   * @param params the parameters map
   * @return the tuple
   */
  default Tuple mapToTuple(Map<String, Object> params) {
    return mapToTuple(params, TupleMapper.mapper(Function.identity()));
  }
}
```

Then for prepared queries:

```java
class User {
  public long id;
  public String firstName;
  public String lastName;
}

// Define row mapper
RowMapper<User> ROW_USER_MAPPER = row -> {
  User user = new User();
  user.id = row.getInteger("id");
  user.firstName = row.getString("firstName");
  user.lastName = row.getString("lastName");
  return user;
};

// Parse template
SqlTemplateParser parser = SqlTemplateParser.create(client,
  "SELECT * FROM users WHERE id=#{id}");

// Execute with row mapping
Map<String, Object> params = Collections.singletonMap("id", 1);
Tuple tuple = parser.mapToTuple(params);

client.preparedQuery(parser.sql())
  .mapping(ROW_USER_MAPPER::map)
  .execute(tuple)
  .onSuccess(users -> {
    users.forEach(user -> {
      System.out.println(user.firstName + " " + user.lastName);
    });
  });
```

And since this is generic, you could leverage the `SqlTemplateParser` for cursor/rowstreams

zZHorizonZz:
Thank you for the time @tsegismont 

This is an interesting approach and actually seems cleaner, but on the other hand I'm not sure how this would work in the case of streaming, because one of the things `SqlTemplateStreamImpl` does is simplify that interaction. Right now there's nothing like `client.stream("select ....").mapping`, instead you have to do something like:

```java
connection.prepare(sqlTemplate.getSql()).map(ps -> {
    RowStream<Row> stream = ps.createStream(fetchSize, tuple);
    stream.handler(row -> handler.handle(mapper.map(row)));
    //etc
}
```

I.e. there's no easier way to map rows at the moment (or at least I wasn't able to find one, happy to be corrected if I'm missing something).

TLDR - I like the idea of making SQL more or less agnostic to the operation, but I think we'd then have a separate issue of making streams easier to use.

tsegismont:
Right, we would still get a standard `RowStream` which emits `Row` items. The user would either manually convert the `Row` with the row mapper, or use RxJava/Mutiny bindings to build complex pipelines.

zZHorizonZz:
I think I will need to think about Stream some more but that seems like separate issue, I'm just not the fan of how you can transform them. Anyways, I will rework this to the SqlTemplateParser you suggested.

tsegismont:
We did not implement stream transformation and manipulation API on purpose (there are great libraries out there that do the job well, like Mutiny or RxJava).

PR #1646, opened 2026-03-31
Pipelining depth grows unbounded when prepared statement is retried with explicit parameter types

### LINKED ISSUE #1644: Missing inflight++ in the "prepare with explicit params" branch
### Version

5.0.8

### Context

Hi. I was discovering sources of sql client and possibly found bug.

Inside **SocketConnectionBase.prepareCommand** method in this branch 
```
 else {
        if (queryCmd.autoCommit() && isIndeterminatePreparedStatementError(cause) && !sendParameterTypes) {
          ChannelHandlerContext ctx = socket.channelHandlerContext();
          // We cannot cache this prepared statement because it might be executed with another type
          ctx.write(prepareCommand(queryCmd, false, true), ctx.voidPromise());
          ctx.flush();
        } else {
...
        }
      }
```

we actually schedule new prepare command without incrementing **inflight** so considering **prepare slot** already decremented in 
```
 if (msg instanceof CommandResponse) {
      inflight--;
      CommandResponse resp =(CommandResponse) msg;
      resp.fire();
    }
```
we actually execute this prepare on the extended command slot, so when this prepare finish we already will get inflight==0, so after the bind/exec part finished this inflight will be decremented to -1. I checked in debugger, it can decrement this way to -2, -3 and etc on each execution of let's say 

```
connection.preparedQuery("SELECT CONCAT('HELLO ', $1)")
          .execute(Tuple.of("WORLD"))
```

and as this lag stays forever. After 100 query you will have additional 100 pipelining depth because of 

```
 while (!paused && inflight < pipeliningLimit && (cmd = pending.poll()) != null) {

```

Besides that, the minor thing, is that if prepared statement cache is enabled, such "inititial prepare fail because of missing type" will not be cached but still will spend statements names. Like you can approach statement name
```
 000442F5 
```
while no actual statement cached. Not sure if it is big problem, but still it spends hex name per each query while it have no meaning as no cache and prepared statement is actually created.

### Steps to reproduce

Enough to run any Postgres instances.

### Do you have a reproducer?

https://gist.github.com/doxlik/acbc02e388a29953288a5a735ba2730a

doxlik:
Fix https://github.com/eclipse-vertx/vertx-sql-client/issues/1644


tsegismont (changes requested):
Thank you @doxlik 

Can you please add a test (that does not pass without the fix)

doxlik:
@tsegismont done!

tsegismont (commented):
Thanks for the update @doxlik 

I think there are a couple of issues with this test. First, it extends `PreparedStatementTestBase`, so it executes all tests defined in this class, which is not necessary. Then I'd rather have a simpler test here: we want to make sure inflight is never negative, so perhaps we could make the field visible for testing or add an assertion.

doxlik:
@tsegismont good points. Done!

vertx-sql-client-codec/src/main/java/io/vertx/sqlclient/codec/SocketConnectionBase.java
```diff
@@ -115,6 +115,10 @@ public int pipeliningLimit() {
     return pipeliningLimit;
   }
 
+  public int inflight() {
```
tsegismont, line 119:
```suggestion
  // Visible for testing
  public int inflight() {
```

tsegismont (approved):
Thank you @doxlik , this looks good to me.

tsegismont:
@doxlik when this is merged, I'll take care of backporting

PR #1666, opened 2026-05-11
Savepoints

### LINKED ISSUE #1369: Support rollback to savepoint after failure
### Questions

I have encountered some strange transaction behaviour. If I execute a query after the transaction starts, create a checkpoint, then this query terminates with an exception (in my case an index constraint error). Then I rollback the checkpoint, and after that when commit transaction the system hangs. At the same time rollback works. In Debug I saw that there is a flag failed in the transaction object. Is this how it should be or am I doing something wrong?

### Version

4.4.6

#### Language

Kotlin

#### Packages

* io.vertx:vertx-core
* io.vertx:vertx-lang-kotlin-coroutines
* io.vertx:vertx-lang-kotlin
* io.vertx:vertx-pg-client

#### Location

Vertx Verticle

#### Code

```kotlin
@OptIn(DelicateCoroutinesApi::class)
fun <T> launchCoroutineOnEventLoopAsVertxFuture(block: suspend () -> T) : Future<T> {
  val vertx = Vertx.currentContext()

  val promise = Promise.promise<T>()

  GlobalScope.launch(vertx.dispatcher()) {
    try {
      promise.complete(block())
    } catch (ex: Exception) {
      promise.fail(ex)
    }
  }

  return promise.future()
}
```

```kotlin
suspend fun <T, Error> withSavepoint(sqlConnection: SqlConnection, block: suspend () -> OkOrError<T, Error>): OkOrError<T, Error> {
    sqlConnection
      .query("SAVEPOINT AsyncSqlConnectionPgSavepointWithErrorHandling")
      .execute()
      .await()

    try {
      val result = block()

      if(result.isError) {
        sqlConnection
          .query("ROLLBACK TO SAVEPOINT AsyncSqlConnectionPgSavepointWithErrorHandling")
          .execute()
          .await()
      }

      return result
    }
    catch (e: PgException) {
      sqlConnection
          .query("ROLLBACK TO SAVEPOINT AsyncSqlConnectionPgSavepointWithErrorHandling")
          .execute()
          .await()

      throw e
    }
    finally {
      sqlConnection
          .query("RELEASE SAVEPOINT AsyncSqlConnectionPgSavepointWithErrorHandling")
          .execute()
          .await()
    }
  }
```

```kotlin
pgPool.withConnection { sqlConnection ->
  launchCoroutineOnEventLoopAsVertxFuture {
    val transaction = sqlConnection.begin().await()

      try {
        val result = 
           try {
              withSavepoint(sqlConnection) {
                 functionWithDatabaseException()
              }
           } catch(e: PgException) {
              ....
             Ok(...)
           }

        if(result.isOk) {
          transaction.commit().await() // here problem
        } else {
          transaction.rollback().await()
        }

        return result
      } catch (e: Exception) {
        transaction.rollback().await()
        throw e
      }
  }
}.await()
```

After stuck on the line `transaction.commit().await()`. With a query in PostgreSQL, I can see that the last query from the application to the database is `ROLLBACK`.

```sql
SELECT *
FROM pg_stat_activity
WHERE application_name = 'vertx-pg-client'
```

PS Everything works fine if I manually make BEGIN/COMMIT requests with SqlConnection instead `sqlConnection.begin()`/`transaction.commit()`.

doxlik:
This PR adds API for Savepoints and implementation of it for Postgres client.

Closes https://github.com/eclipse-vertx/vertx-sql-client/issues/1369

### DIFF
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
@@ -26,6 +26,16 @@ import io.vertx.core.Future;
 public interface Transaction {
 
   /**
+   * Create a savepoint in this transaction.
+   *
+   * <p>Fails with {@link UnsupportedOperationException} when the driver does not
+   * support savepoints.
+   *
+   * @return a future notified with the created savepoint
+   */
+  Future<Savepoint> createSavepoint();
+
+  /**
    * Commit the current transaction.
    */
   Future<Void> commit();
--- a/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SavepointImpl.java
+++ b/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SavepointImpl.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2011-2025 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.sqlclient.impl;
+
+import io.vertx.core.Future;
+import io.vertx.sqlclient.Savepoint;
+
+public class SavepointImpl implements Savepoint {
+
+  private enum State {
+    ACTIVE,
+    PENDING,
+    RELEASED
+  }
+
+  private final TransactionImpl transaction;
+  private final String name;
+  private State state;
+
+  public SavepointImpl(TransactionImpl transaction, String name) {
+    this.transaction = transaction;
+    this.name = name;
+    this.state = State.ACTIVE;
+  }
+
+  @Override
+  public Future<Void> rollback() {
+    return execute(false, () -> transaction.rollbackToSavepoint(name));
+  }
+
+  @Override
+  public Future<Void> release() {
+    return execute(true, () -> transaction.releaseSavepoint(name));
+  }
+
+  private Future<Void> execute(boolean release, Action action) {
+    synchronized (this) {
+      if (state == State.RELEASED) {
+        return transaction.failedFuture("Savepoint already released");
+      }
+      if (state == State.PENDING) {
+        return transaction.failedFuture("Savepoint command already in progress");
+      }
+      state = State.PENDING;
+    }
+    return action.execute().andThen(ar -> {
+      synchronized (SavepointImpl.this) {
+        if (ar.succeeded()) {
+          state = release ? State.RELEASED : State.ACTIVE;
+        } else {
+          state = State.ACTIVE;
+        }
+      }
+    });
+  }
+
+  @FunctionalInterface
+  private interface Action {
+    Future<Void> execute();
+  }
+}
--- a/pom.xml
+++ b/pom.xml
@@ -189,6 +189,7 @@
     <module>vertx-db2-client</module>
     <module>vertx-sql-client-templates</module>
     <module>vertx-oracle-client</module>
+    <module>vertx-pg-savepoints-example</module>
   </modules>
 
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -189,7 +189,6 @@
     <module>vertx-db2-client</module>
     <module>vertx-sql-client-templates</module>
     <module>vertx-oracle-client</module>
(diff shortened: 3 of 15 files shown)


tsegismont:
Can you please rebase the PR? The unrelated CI failures should be fixed now

doxlik:
@tsegismont rebasing done!

CI still have some failure but seems like it couldn't build in action with Postgres14, not any test problem.
Not sure what is reason for that, seems like it is some maven issue in that particular action.

tsegismont:
The CI has been stabilized hopefully, would you mind rebasing again? Thank you

doxlik:
@tsegismont Hi, there were no new commits since the last rebase, so I amended my latest commit to trigger a push. It seems CI is passing now. Looking forward to your review!

doxlik:
@vietj @tsegismont Hi. I added support for other DBs. Please have a look once you have time.

vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SavepointImpl.java
```diff
@@ -0,0 +1,69 @@
...
+          state = release ? State.RELEASED : State.ACTIVE;
+        } else {
+          state = State.ACTIVE;
+        }
+      }
+    });
+  }
+
+  @FunctionalInterface
+  private interface Action {
```
tsegismont: You could get rid of this type and use `BiFunction<TransactionImpl, String, Future<Void>>`

Then the two methods become:

```java
  @Override
  public Future<Void> rollback() {
    return execute(false, TransactionImpl::rollbackToSavepoint);
  }

  @Override
  public Future<Void> release() {
    return execute(true, TransactionImpl::releaseSavepoint);
  }
```

vertx-sql-client/src/main/java/io/vertx/sqlclient/Transaction.java
```diff
@@ -25,6 +25,16 @@
...
 
+  /**
+   * Create a savepoint in this transaction.
+   *
+   * <p>Fails with {@link UnsupportedOperationException} when the driver does not
+   * support savepoints.
+   *
+   * @return a future notified with the created savepoint
+   */
+  Future<Savepoint> createSavepoint();
```
tsegismont, line 36: Have you considered letting users provide a name? As far as I understand, names are scoped to the transaction so the risk of conflict is low. And if the database includes it in an informational or error message, it will be easier for the user to reason about it.

tsegismont (changes requested):
Thank you @doxlik 

Overall, this looks good to me, but we're missing documentation for this feature.

doxlik:
@tsegismont thank you for review. I added docs, refactor to BiFunction and added API for user supplied savepoint name, can you please have a look!

tsegismont (commented):
Since we have access to the driver, perhaps could we have a new method in the driver that tells how to quote a name, and then lift restrictions on the savepoint name

doxlik:
@tsegismont done, can you please check

vertx-mssql-client/src/main/java/io/vertx/mssqlclient/spi/MSSQLDriver.java
```diff
@@ -67,6 +67,27 @@ public int appendQueryPlaceholder(StringBuilder queryBuilder, int index, int cur
     return index;
   }
 
+  /**
+   * Transact-SQL delimits identifiers with square brackets.
+   */
+  @Override
+  public StringBuilder appendQuotedIdentifier(StringBuilder sql, String identifier) {
+    return sql.append('[').append(identifier.replace("]", "]]")).append(']');
```
tsegismont: Shouldn't we also replace `[` ?

vertx-sql-client/src/main/asciidoc/transactions.adoc
```diff
@@ -40,3 +40,40 @@ After the transaction completes, the connection is returned to the pool and the
...
+empty, when it holds a double quote, which Oracle refuses inside a savepoint identifier, or when it is
+longer than 32 characters, the shortest limit across the supported databases:
+
+[source,$lang]
+----
+{@link examples.SqlClientExamples#savepoint03(io.vertx.sqlclient.Pool)}
+----
+
+Not every database supports savepoints, and some support only part of the API. When an operation is not
+supported the call fails with an `UnsupportedOperationException` rather than pretending to succeed.
```
tsegismont: I think we can remove this part since all clients add their own savepoint notes

tsegismont (commented):
Just a last minor comment

vertx-sql-client/src/main/asciidoc/transactions.adoc
doxlik: @tsegismont done!

tsegismont (approved):
Excellent, thank you @doxlik !

PR #1671, opened 2026-06-04
Support channel binding configuration in PgConnectOptions

jorsol:
Motivation:

Introduce support for configuring channel binding mechanisms during the
SCRAM-SHA-256-PLUS authentication handshake. This allows users to require,
prefer, or disable channel binding securely.

- Add channelBinding property and getters/setters to PgConnectOptions
- Update PgConnectionUriParser to parse 'channel_binding' from connection URIs
- Propagate the configuration to ScramAuthentication to negotiate -PLUS variants
- Fail the connection handshake if a mismatch or unsupported negotiation occurs

Conformance:

### DIFF
--- a/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgSocketConnection.java
+++ b/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgSocketConnection.java
@@ -25,6 +25,7 @@ import io.vertx.core.net.ClientSSLOptions;
 import io.vertx.core.spi.metrics.ClientMetrics;
 import io.vertx.core.internal.ContextInternal;
 import io.vertx.core.internal.net.NetSocketInternal;
+import io.vertx.pgclient.ChannelBinding;
 import io.vertx.pgclient.PgConnectOptions;
 import io.vertx.pgclient.PgException;
 import io.vertx.pgclient.impl.codec.ExtendedQueryPgCommandMessage;
@@ -136,6 +137,10 @@ public class PgSocketConnection extends SocketConnectionBase {
 
   public int getSecretKey() {
     return secretKey;
+  }
+
+  public ChannelBinding channelBinding() {
+    return connectOptions.getChannelBinding();
   }
 
   @Override
--- a/vertx-pg-client/src/test/java/io/vertx/tests/pgclient/TLSTest.java
+++ b/vertx-pg-client/src/test/java/io/vertx/tests/pgclient/TLSTest.java
@@ -20,10 +20,12 @@ import io.vertx.core.net.PemTrustOptions;
 import io.vertx.ext.unit.Async;
 import io.vertx.ext.unit.TestContext;
 import io.vertx.ext.unit.junit.VertxUnitRunner;
+import io.vertx.pgclient.ChannelBinding;
 import io.vertx.pgclient.PgConnectOptions;
 import io.vertx.pgclient.PgConnection;
 import io.vertx.pgclient.SslMode;
 import io.vertx.pgclient.SslNegotiation;
+import io.vertx.sqlclient.ClosedConnectionException;
 import io.vertx.sqlclient.Tuple;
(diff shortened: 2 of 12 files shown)


tsegismont:
Can you please sign the Eclipse ECA? Would you mind adding the option field to connect options while you're at it?

tsegismont:
Ideally we'd also have a test that verifies it is not possible to connect if the user sets the `REQUIRE` policy for the client and the server does not explicitly advertise a channel-bound mechanism.

jorsol:
@tsegismont Could you please review again? I'm not fully familiar with the code base, so maybe you can point me to improvements.

vertx-pg-client/src/test/java/io/vertx/tests/pgclient/TLSTest.java
```diff
@@ -261,4 +263,33 @@ public void testSslNegotiationDirectWithVerifyFull(TestContext ctx) {
...
+  public void testChannelBindingRequireWithoutSsl(TestContext ctx) {
+    Async async = ctx.async();
+    PgConnectOptions options = new PgConnectOptions(ruleOptionalSll.options())
+      .setSslMode(SslMode.DISABLE)
+      .setChannelBinding(ChannelBinding.REQUIRE)
+      .setSslOptions(new ClientSSLOptions().setTrustAll(true));
+
+    PgConnection.connect(vertx, options).onComplete(ctx.asyncAssertFailure(err -> {
+      // ctx.assertEquals("Channel bindins is required", err.getMessage());
+      ctx.assertTrue(err instanceof ClosedConnectionException); // TODO: handle ChannelBindingException
```
tsegismont, line 291: As in #1672 (this could be due to changes in the way we handle exceptions since 5.1)

tsegismont: I'll take a look after merging

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgSocketConnection.java
```diff
@@ -81,6 +82,11 @@ protected PgConnectOptions connectOptions() {
   @Override
   public void init() {
     codec = new PgCodec(useLayer7Proxy);
+
+    socket.channelHandlerContext().channel()
+        .attr(AttributeKey.valueOf("channel_binding"))
+        .set(connectOptions.getChannelBinding());
```
tsegismont: We can avoid indirect communication through a channel attribute.

You could add a method in this class like:

```java
  public ChannelBinding channelBinding() {
    return connectOptions.getChannelBinding();
  }
```

In `io.vertx.pgclient.impl.codec.InitPgCommandMessage#handleAuthenticationSasl`

```java
    PgSocketConnection pgSocketConn = (PgSocketConnection) cmd.connection().unwrap();
    ScramClientInitialMessage initialSaslMessage = scramSession.createInitialSaslMessage(in, encoder.channelHandlerContext(), pgSocketConn.channelBinding());
    encoder.writeScramClientInitialMessage(initialSaslMessage);
```

tsegismont (commented):
Thank you @jorsol ! Please find a few comments

vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgSocketConnection.java
jorsol: Hi @tsegismont, thanks for pointing to this, those changes are done now.

vertx-pg-client/src/test/java/io/vertx/tests/pgclient/TLSTest.java
jorsol: Thank you, I'm really lost on how to handle the exceptions here.

tsegismont (approved):
LGTM, thank you @jorsol

PR #1673, opened 2026-06-16
Instructions for coding agents

tsegismont:
This is an attempt to guide coding agents when contributing or reviewing code.

`AGENTS.md` and `SKILL.md` are standards that should be understood by most coding agents.

### DIFF
--- a/.agents/skills/writing-tests/SKILL.md
+++ b/.agents/skills/writing-tests/SKILL.md
@@ -1,0 +1,109 @@
+---
+name: writing-tests
+description: >
+  Testing patterns for Vert.x SQL Client: test frameworks, async testing,
+  test locations, and how to run tests.
+---
+
+# Writing Tests
+
+Vert.x SQL Client uses JUnit 4 with Vert.x-specific test utilities. Tests and documentation are mandatory for contributions.
+
+## Test Framework
+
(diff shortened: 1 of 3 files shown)


tsegismont:
@vietj can you please take a look? Consider these as experimental (and thus likely to evolve when this PR is merged).

.agents/skills/writing-tests/SKILL.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Standard Test Pattern
+
+```java
+@RunWith(VertxUnitRunner.class)
+public class MyTest {
+
+  @Test
+  public void testAsyncOperation(TestContext ctx) {
+    Async async = ctx.async();
```
vietj, line 31: I think this is not correct, because calling asyncAssertSuccess creates an async under the hood and will prevent the test from finishing.

So I think we should have three testAsyncOperation methods

1. demonstrate the usage of async with a simple callback:

vertx.runOnContext(v -> async.complete());

2. demonstrate with direct usage of asyncAssertSuccess (the currnet method).

3. demonstrate the usage of asyncAssertSuccess called from runOnContext that requires an async.

vietj (approved):

Issue #1675, opened 2026-06-16
Socket not closed on exception-driven connection failure


tsegismont:
When a connection handles a failure (e.g. decoding exception), the underlying Netty channel is not closed unless the application explicitly calls `connection.close()`.

This leaks socket file descriptors and Netty resources until the application closes the connection or the Vertx instance shuts down.

tsegismont:
Fixed by https://github.com/eclipse-vertx/vertx-sql-client/commit/261ca5555bdfe4483a8cd3f66e5e825c21e45b97

Issue #1674, opened 2026-06-16
Inflight commands fail with wrong exception after connection error


tsegismont:
When a connection handles a failure (e.g. decoding exception), `inflight` commands fail with `ClosedConnectionException` instead of the actual error that killed the connection, making it hard for the caller to diagnose what went wrong.

tsegismont:
Fixed by https://github.com/eclipse-vertx/vertx-sql-client/pull/1679

Issue #1672, opened 2026-06-11
MSSQL client hides the cause of an error


DavideD:
### Version

5.1.2

### Context

  When the MSSQL client encounters an unsupported data type (e.g. XML), [DataType.decodeTypeInfo()](https://github.com/eclipse-vertx/vertx-sql-client/blob/e4847d72a6f97360c98e8ff28da5f913a26069a4/vertx-mssql-client/src/main/java/io/vertx/mssqlclient/impl/codec/DataType.java#L1068) throws an UnsupportedOperationException:

```java
  // DataType.java — default implementation
  public TypeInfo decodeTypeInfo(ByteBuf byteBuf) {
      throw new UnsupportedOperationException("Unable to decode typeInfo for " + name());
  }
```

  In previous versions (5.0.x), this exception was propagated to the caller, allowing downstream libraries to catch it and provide a meaningful error message.

  Since 5.1.0, the exception is no longer propagated. The only exception thrown is `ClosedConnectionException`

tsegismont:
@DavideD such connection errors are now reported through `io.vertx.sqlclient.SqlConnection#exceptionHandler`

Does that work for you?

DavideD:
Mmmh... OK, I will test it and let you know

DavideD:
Yep, that works.

Thanks @tsegismont

tsegismont:
Reopened and filed https://github.com/eclipse-vertx/vertx-sql-client/issues/1672 , because this is not specific to MSSQL

tsegismont:
@DavideD see https://github.com/eclipse-vertx/vertx-sql-client/pull/1676

tsegismont:
Fixed by https://github.com/eclipse-vertx/vertx-sql-client/pull/1679

PR #1676, opened 2026-06-16
Fix wrong exception propagation on connection errors and channel binding failure

tsegismont:
See #1672, #1674, #1675

When a connection handles a failure, the underlying Netty socket should be closed, avoiding leak of file descriptors and Netty resources.

Additionally, inflight commands shouldn't fail with ClosedConnectionException but with the actual error.

Some portions of this content were created with the assistance of Claude Code.

### DIFF
--- a/vertx-mssql-client/src/test/java/io/vertx/tests/mssqlclient/MSSQLQueriesTest.java
+++ b/vertx-mssql-client/src/test/java/io/vertx/tests/mssqlclient/MSSQLQueriesTest.java
@@ -40,8 +40,7 @@ import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import static org.hamcrest.CoreMatchers.instanceOf;
-import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.*;
 import static org.hamcrest.MatcherAssert.assertThat;
 
 @RunWith(VertxUnitRunner.class)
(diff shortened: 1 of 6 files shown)


tsegismont:
@vietj can you please take a look? This is the fix for problems introduced by c684c42ea818c4f3c61e444835f39ef074cf58ba

jorsol:
Hi @tsegismont 

I tested your branch but I don't get the real message, it seems that it still returns the ClosedConnectionException instead of the actual error.

Could you please confirm what I'm doing wrong?

```java
    Vertx vertx = Vertx.vertx();
    PgConnectOptions connectOptions = new PgConnectOptions()
        .setPort(5432)
        .setHost("localhost")
        .setDatabase("postgres")
        .setUser("postgres")
        .setSslOptions(new ClientSSLOptions().setTrustAll(true))
        .setSslMode(SslMode.DISABLE)
        .setChannelBinding(io.vertx.pgclient.ChannelBinding.REQUIRE)
        .setPassword("123");

    PgConnection.connect(vertx, connectOptions)
        .onComplete(res -> {
          if (res.succeeded()) {
            System.out.println("Connected...");
          } else {
            System.out.println("Could not connect: " + res.cause().getMessage());
          }
        });
    vertx.close();
```

vietj:
my branch does not aim to fix the bug we are currently observing in this issue, it aims first to get the exception handling flow correct from the NetSocket / SocketConnectionBase perspective before doing any other fix.

vietj:
in particular : handleException should not close the socket in your patch because the socket will be closed by the code calling handleException, this is what I'd like to clarify and avoid.

tsegismont:
@vietj please take another look

I had to make a special case in `InitPgCommandMessage.java`, because if we let the exception bubble up, then:

- the client tries to close the channel gracefully
- it sends a close command
- Postgres replies with a failure like "invalid state"
- the user gets the "invalid state failure" instead of the sasl failure

Nevertheless, I don't believe we have a way to close the connection abruptly like PostgreSQL expects:

>If the frontend does not support the authentication method requested by the server, then it should immediately close the connection.

https://www.postgresql.org/docs/current/protocol-flow.html

I believe this would require changes in `VertxConnection`. Thoughts?

tsegismont:
Well in fact, `encoder.close()` does exactly that :sweat:

vietj:
your changes in `InitPgCommandMessage` makes sense to me and it looks more idiomatic to proceed this way.

vietj:
another way would be to close the ChannelHandlerContext ?

tsegismont:
> another way would be to close the ChannelHandlerContext ?

Right. It has the same effect for our case. The channel output becomes null and no close message is sent to the server.

PR #1683, opened 2026-07-17
[5.1] Bump scram-client from 3.2 to 3.3 fixing CVE-2026-53712

julianladisch:
scram-client 3.3 fixes a flaw that allows an attacker capable of performing a TLS machine-in-the-middle (MITM) attack to silently downgrade a connection from SCRAM-SHA-256-PLUS (with channel binding)
to standard SCRAM-SHA-256 (without channel binding), bypassing strict client-side enforcement policies.

See https://github.com/advisories/GHSA-p9jg-fcr6-3mhf and https://github.com/ongres/scram/releases/tag/3.3

### DIFF
--- a/vertx-pg-client/pom.xml
+++ b/vertx-pg-client/pom.xml
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>com.ongres.scram</groupId>
       <artifactId>scram-client</artifactId>
-      <version>3.2</version>
+      <version>3.3</version>
     </dependency>
 
     <!-- Testing purposes -->


julianladisch:
5.1 is not affected because is uses the default "prefer"/"allow" channel binding policy and not the affected "strict" channel binding policy:
* https://github.com/eclipse-vertx/vertx-sql-client/blob/5.1.5/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/auth/scram/ScramSessionImpl.java#L71-L77
* https://github.com/ongres/scram/security/advisories/GHSA-p9jg-fcr6-3mhf

tsegismont:
Thanks for the heads-up @julianladisch !

Indeed, Vert.x is not affected because strict mode is not supported in 5.1, and 5.2 will use the fixed version of Scram Client https://github.com/eclipse-vertx/vertx-sql-client/pull/1671

PR #1684, opened 2026-07-21
Fail acquisition with the hook failure and release the lease when aft…

andrewkav:
When the afterAcquire hook failed during SqlConnectionPool#acquire, the acquisition was failed with the enclosing (null) failure instead of the hook failure, so callers observed a null cause. The lease was also neither recycled nor closed, leaking the pool slot: with enough hook failures the pool starved with no way to recover.

Recycle the lease and fail the acquisition with the actual hook failure.

Motivation:

Conformance:


tsegismont:
Thanks, closing in favor of https://github.com/eclipse-vertx/vertx-sql-client/pull/1699

PR #1686, opened 2026-08-10
Support IBM JDBC URL separators in DB2 connection URI parser

### LINKED ISSUE #1118: DB2ConnectionUriParser doesn't support standard DB2 URL character separators
While trying to leverage the **`SqlConnectOptions.fromUri( URI uri)`** parsing feature in the SQL I ran into a problem with our DB2 url.

In Hibernate Reactive we're testing against DB2 via standard DB2 URL like: 
 * **`jdbc:db2://localhost:52353/hreact:user=hreact;password=hreact;`**
 
This url is defined based on the IBM spec/doc and which uses a **`:`** colon as the separator between Database and parameter info and a **`;`** semicolon between parameter key/value pairs.

However, the call to **`SqlConnectOptions.fromUri( URI uri)`** with this DB2 url will failed because vertx-db-client's parser doesn't recognize these separators. In stead it uses  the same separators (  `?` and `&` ) as it does with the other client parsers.

Was this intended? or is it a bug?

You can reproduce the error by adding these two tests (one succeeds, one fails) to: [DB2ConnectionUriParserTest.java](https://github.com/eclipse-vertx/vertx-sql-client/blob/master/vertx-db2-client/src/test/java/io/vertx/db2client/DB2ConnectionUriParserTest.java)

```
  @Test
  public void testAlternateURL(){
    uri = "db2://localhost:4444/hreact?user=hreact&password=hreact";
    actualParsedResult = parse(uri);

    expectedParsedResult = new JsonObject()
      .put("user", "hreact")
      .put("password", "hreact")
      .put("host", "localhost")
      .put("port", 4444)
      .put("database", "hreact");

    assertEquals(expectedParsedResult, actualParsedResult);
  }

  @Test(expected = IllegalArgumentException.class)
  public void testAlternateURLWithSemicolonFail(){
    uri = "db2://localhost:4444/hreact:user=hreact&password=hreact";
    actualParsedResult = parse(uri);
  }
```

jnbdz:
The DB2 connection URI parser only accepted query-string style attributes (?key=value&key=value). IBM's documented URL syntax uses a colon to separate the database from the connection attributes and semicolons between key/value pairs:

  db2://host:port/database:user=app;password=secret;

Accept this form in addition to the existing one. The attribute block is only allowed after a database name, and mixing it with & separators is rejected.

Fixes eclipse-vertx/vertx-sql-client#1118

### DIFF
--- a/vertx-db2-client/src/main/asciidoc/index.adoc
+++ b/vertx-db2-client/src/main/asciidoc/index.adoc
@@ -143,6 +143,12 @@ The URI format for a connection string is:
 db2://[user[:[password]]@]host[:port][/database][?<key1>=<value1>[&<key2>=<value2>]]
 ----
 
+The IBM JDBC URL syntax is supported as well, using `:` to separate the database from the parameters and `;` between parameter key/value pairs:
+
+----
+db2://[user[:[password]]@]host[:port]/database:<key1>=<value1>;[<key2>=<value2>;]
+----
+
 NOTE: Configuring parameters in connection URI will override the default properties.
 
 Currently, the client supports the following parameter keys:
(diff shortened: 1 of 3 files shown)


tsegismont (approved):
LGTM, thanks

PR #1693, opened 2026-09-03
Fix character encoding problem for non-ASCII/Chinese data (#795)

### LINKED ISSUE #795: vertx-db2-client character encoding problem
### Version

3.9.4 and 4.0.0.Beta3

### Context

when query data with vertx-db2-client, non ascii characters(for example, chinese characters) not encoded correctly

### Steps to reproduce

```kotlin
...
val rs = pool.connection.await().query("select * from V_ITEMS").execute().await()
val list = Sequence(rs::iterator).map(Row::toJson).toList()
println(Json.encode(list))
...
```

sanjaykumarmtt:
**### Motivation**
_Fixes #795_

When reading Chinese or other multi byte characters from DB2 they show up as `???` because the driver only checked OS type instead of the server CCSID.

This change fixes the issue:
- Maps DB2 CCSID numbers to Java Charset in `CCSIDConstants`.
- Uses server column CCSID (`sda.ccsid_`) if present; otherwise, uses connection level encoding (`getCcsidMbcEncoding()`).

### Conformance
- I have signed the Eclipse Contributor Agreement (ECA).
- Followed code style and tested with DB2 database.


tsegismont (commented):
Thank you @sanjaykumarmtt 

Could you please add a test?

sanjaykumarmtt:
​@tsegismont thank you for the review I have added the unit test and formatted the code accordingly

tsegismont (commented):
Thanks for adding a test, but I was thinking about actually testing a DB2 query rather than a unit test. Is it possible?

sanjaykumarmtt:
thanks for the feedback @tsegismont yes that makes sense I will add a DB2 query integration test for this and update the PR soon

sanjaykumarmtt:
hi @tsegismont ​apologies for the delay I have replaced the previous unit test with a DB2 query integration test in DB2QueriesTest as suggested please review it and let me know if any further changes are needed

sanjaykumarmtt:
​I have updated the PR with the suggested fix: casting the untyped parameter marker with CAST(? AS VARCHAR(100)) in DB2QueriesTest, and reverted the unrelated changes.
​Could you please approve and trigger the CI workflow when you have a moment?

tsegismont (commented):
Thanks. For me the test passes without the code changes. Can you check if it's actually testing what you want?

sanjaykumarmtt:
sorry for the delay ​I found why the test passed earlier the target chinese string was not present in the database table so the query returned 0 rows and did not properly test the encoding.

sanjaykumarmtt:
​I have updated the test ​added the Chinese string to init.sql ​updated the test assertion to verify that 1 row is returned and the decoded message matches the expected value.

sanjaykumarmtt:
​Could you please approve and trigger the CI workflow when you have a moment?

tsegismont:
The test still passes without the changes. I think this happens because it's all UTF-8.

Please make sure to create a test that does not pass without the changes.

sanjaykumarmtt:
hi @tsegismont the original test was passing without changes because the database environment and connection default to UTF-8 CCSID 1208 which bypasses the specific Chinese CCSID mapping logic

to trigger the mapping I tried casting the column using CAST('你好世界' AS VARCHAR(100) CCSID 1386) but DB2 threw SQLCODE=-622 (SQLSTATE=56031) because this database configuration does not support that clause

sanjaykumarmtt:
I tried writing a dedicated unit test (CCSIDConstantsTest) locally to verify the mapping and it passes with the changes and fails without them
Could you please guide me on whether I should commit this unit test or if you have any suggestions on how to test this properly?

tsegismont:
I'm not a DB2 expert unfortunately. Perhaps is it possible to create a specific database with different encoding?

sanjaykumarmtt:
​to be honest I m not a DB2 expert at all I recently graduated from college where I built a chat app using Java sockets and I really wanted to contribute to a real world open source project like vert.x. earlier I only tried casting the encoding inline at the query level to see if it would trigger the mapping code.

​after your suggestion I read up on creating a dedicated non UTF 8 database like GBK or CCSID 1386 for testing but I noticed a few practical drawbacks:
​Slower CI builds: Running CREATE DATABASE ... USING CODESET GBK in DB2 takes an extra 3 to 5 minutes on every CI run.
​CI instability: DB2 is resource-heavy, and creating an additional database during testing can easily lead to timeouts or memory crashes in GitHub Actions runners.
​Scope of the fix: Our fix is purely Java-side logic in CCSIDConstants to map CCSID numbers to standard Java Charsets.


because of this the dedicated unit test CCSIDConstantsTest feels like the cleanest and safest solution It reliably asserts that CCSIDs 1386, 5488, 1114 map to GBK and 935 maps to Cp935 It fails on master and passes cleanly with this fix
​would it be okay if I push this unit test commit to the PR?

tsegismont:
Instead of creating an entire database, how about creating a new table in the setup file:

```sql
DROP TABLE IF EXISTS encoding_test;
CREATE TABLE encoding_test (
  id      INTEGER NOT NULL,
  message VARCHAR(256) CCSID 1386,
  PRIMARY KEY (id)
);
INSERT INTO encoding_test VALUES (1, '你好世界');
```

sanjaykumarmtt:
​I checked the official IBM DB2 documentation and in DB2 LUW a UTF-8 database does not allow setting a column level CCSID 1386
then ​I also tested your suggested query in the container setup and DB2 failed to create the table with this exact error

[DB2] CREATE TABLE encoding_test ( id      INTEGER NOT NULL, message VARCHAR(256) CCSID 1386, PRIMARY KEY (id) )
[DB2] DB21034E  The command was processed as an SQL statement because it was not a 
[DB2] valid Command Line Processor command.  During SQL processing it returned:
[DB2] SQL0622N  The clause "CCSID <integer>" is invalid for this database.  
[DB2] SQLSTATE=56031
Screenshot 2026-09-14 192726

Screenshot 2026-09-14 182112
Screenshot 2026-09-14 192712

sanjaykumarmtt:
could you please let me know how you would like to proceed next?

tsegismont:
In this case, yes, we'll stick to the unit testing approach. Ideally we'd try with another database, but that's not realistic with our test setup.

Thank you for your going through the research, I know it requires patience and commitment, but sometimes it's the only option to make the right decision.

sanjaykumarmtt:
thankyou  @tsegismont
​I add the unit test in the CCSIDConstantsTest class file I update the PR with this test
​review the changes you tell me if you need any more changes

tsegismont:
Please revert files with only import or formatting changes, then squash your commits

sanjaykumarmtt:
​done I have reverted the formatting changes and squashed the commits

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/Typdef.java
```diff
@@ -1079,8 +1082,14 @@ void updateColumn(Cursor netCursor,
...
-            netCursor.charset_[columnIndex] = metadata.isZos() ? CCSIDConstants.EBCDIC : CCSIDConstants.UTF8;
-            netCursor.ccsid_[columnIndex] = sda.ccsid_;
+        	if (sda.ccsid_ != 0) {
+                netCursor.charset_[columnIndex] = CCSIDConstants.getCharsetForCCSID(sda.ccsid_);
+                netCursor.ccsid_[columnIndex] = sda.ccsid_;
+            } else {
+                //If Server 0 is sent, use connection-level typedef encoding.
+                netCursor.charset_[columnIndex] = getCcsidMbcEncoding();
+                netCursor.ccsid_[columnIndex] = getCcsidMbc();
+            }
```
tsegismont: The new code falls back to `getCcsidMbcEncoding()`, which returns `UTF8` when no CCSID was negotiated.

It breaks when we are dealing  with a z/OS EBCDIC connection.

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/CCSIDConstants.java
```diff
@@ -18,15 +18,62 @@
...
+                return ISO_8859_1;
+            case 1200:
+                return UTF16BE;
+            case 37:
+                return Charset.forName("CP037");
+            case CCSID_EBCDIC:
+            case 1047:
+                return EBCDIC;
+            case 1252:
+                return Charset.forName("windows-1252");
```
tsegismont: `CP037` and `windows-1252` are returned via `Charset.forName()` on every call, while `GBK`, `CP935`, and `EBCDIC` are cached as static constants. Please make them consistent: all new charsets should be loaded eagerly at class-load time using `Charset.isSupported()` as the guard:

```java
private static final Charset CP037 = Charset.isSupported("CP037") ? Charset.forName("CP037") : null;
private static final Charset WINDOWS_1252 = Charset.isSupported("windows-1252") ? Charset.forName("windows-1252") : null;
```

Then in `getCharsetForCCSID()`, guard the return and fall back to UTF-8 if the charset is unavailable:

```java
case 37:
    return CP037 != null ? CP037 : UTF8; // when using the default we should probably log something
case 1252:
    return WINDOWS_1252 != null ? WINDOWS_1252 : UTF8; // when using the default we should probably log something
```

`GBK` and `CP935` were also added in this PR as unconditional `Charset.forName()` calls in static initializers, but they carry the same risk and should receive the same treatment.

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/CCSIDConstants.java
```diff
@@ -18,15 +18,62 @@
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 
+import io.vertx.core.internal.logging.Logger;
+import io.vertx.core.internal.logging.LoggerFactory;
+
 public class CCSIDConstants {
 
+	// Log errors and warnings to the log file
```
tsegismont: This comment is redundant

tsegismont (commented):
Thank you @sanjaykumarmtt , another round of comments

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/Typdef.java
sanjaykumarmtt: updated the fallback logic to check isCcsidMbcSet() first and default back to getCcsidSbcEncoding() for z/OS EBCDIC connections

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/CCSIDConstants.java
sanjaykumarmtt: cached CP037, WINDOWS_1252, GBK, and CP935 statically with Charset.isSupported() guards and added UTF-8 fallbacks

vertx-db2-client/src/main/java/io/vertx/db2client/impl/drda/CCSIDConstants.java
sanjaykumarmtt: removed the redundant comment

tsegismont:
Thank you @sanjaykumarmtt

PR #1696, opened 2026-09-12
Support streaming COPY FROM STDIN and COPY TO STDOUT

### LINKED ISSUE #153: COPY command
Hi @vietj.  Is it possible to run a [COPY](https://www.postgresql.org/docs/10/static/sql-copy.html) command through reactive-pg-client?  Specifically, to load a CSV of data in what I believe is the most efficient way to load a large number of records into the system.  I believe the CSV has to either be on the DB server itself (which we'd like to avoid if possible), so can we in some way use the STDIN argument of the COPY command to provide the CSV file over the wire to it?

Thanks!

doxlik:
Expose the PostgreSQL COPY subprotocol on PgConnection as raw byte streams, so bulk data transfers no longer have to be materialized in memory.

copyOut(String) executes a COPY ... TO STDOUT statement and returns a PgCopyOut, a ReadStream of Buffer whose completion() carries the row count reported by the server. Each CopyData message is emitted as one buffer by default; PgCopyOutOptions.setAggregationThreshold(int) coalesces small messages when throughput matters more than message boundaries.

copyIn(String) executes a COPY ... FROM STDIN statement and returns a PgCopyIn, a WriteStream of Buffer honouring writeQueueFull() and drainHandler(). Small writes are batched up to PgCopyInOptions chunkSize, and abort(String) sends CopyFail so the server rolls the copy back.

The outcome of a COPY is always the one PostgreSQL reports, never synthesised by the client: completion() resolves at ReadyForQuery, which is also what leaves the connection usable for the next statement.

Resolves https://github.com/eclipse-vertx/vertx-sql-client/issues/153

### DIFF
--- a/vertx-pg-client/src/main/asciidoc/index.adoc
+++ b/vertx-pg-client/src/main/asciidoc/index.adoc
@@ -13,6 +13,7 @@ The client is reactive and non-blocking, allowing to handle many database connec
 * Publish / subscribe using PostgreSQL `NOTIFY/LISTEN`
 * Batch and cursor
 * Row streaming
+* `COPY` streaming
 * Command pipelining
 * RxJava API
 * Direct memory to object without unnecessary copies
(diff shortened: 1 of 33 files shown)


doxlik:
@vietj @tsegismont hi. Please have a look once you have time.

